Issue #76, opened 2016-05-30
Fix for metadata template section


txau:
Its all the validations inside that form.

- Explain to users when a property name is duplicated
- Explain to users when extra information is needed (ie. select dictionary on list fields)

PR #81, opened 2016-05-30
Templates form migration

txau:
Solves issue: #76


konzz:
- [ ] Prevent elements to move when the duplicated label error appears.
- [ ] Change "Use as library filter" for "Use as filter"
- [ ] Change "Required field" for "Required"
- [ ] Change "This property will be used togheter for filtering with other equal to him." for "This property will be used for filtering the library results. When properties match in equal name and field type with other document types, they will be combined for filtering."
- [ ] Change "Save Template" for "Save"

PR #84, opened 2016-05-31
document form resets when closing metadata panel

daneryl:
solves #70 

When clicking outside, this panel does not close, the only way to close it is clicking on the close button, this is when the form resets.


daneryl:
Add confirm dialog before reseting form:

"All changes will be lost, are you sure you want to proceed ?"

daneryl:
confirm dialog only will show when the form is dirty.

Issue #68, opened 2016-05-30
Tweak filters


konzz:
- [x] Change "FILTER BY DOCUMENT TYPE:" to "DOCUMENT TYPE"
- [x] Remove bottom dotted border, right after document types.

danicatalan:
![design](https://cloud.githubusercontent.com/assets/6364153/15650895/2844e91a-267c-11e6-9591-3f3fd3fd9233.png)

PR #102, opened 2016-06-02
Tweak filters

danicatalan:
Solves https://github.com/huridocs/uwazi/issues/68


danicatalan:
Closed because target is master, not development.

Issue #99, opened 2016-06-01
Uploads: change drag zone message


txau:
To "Drag & drop your files / or click here for browsing your local files".

danicatalan:
This issue is fixed with the upper PR as a collateral change.

Issue #80, opened 2016-05-30
Provide feedback when trying to upload document types other than PDF


txau:
- Remove re-upload button.

danicatalan:
I uploaded the video here, same functionality than **upload failed** or **conversion failed**: https://github.com/huridocs/uwazi-assets/tree/master/80_invalid_file_type

@daneryl @konzz we have a doubt. Where we check the file, before of after uploading the document? If made the screen supposing we will check after that, but if we check first, just ignore the upload progress bar.

Thanks!

daneryl:
@danicatalan @txau @konzz actually we do not have a file type validation, only a file extension validation, we need to do a proper mime type check on the server because not all browsers supports the FileReader and Blob api to do it, so i think its better to do it after the upload.

konzz:
@danicatalan we still need a way to tell the user that only PDF is accepted, or the user will try to drag and drop other document type and nothing will happen, and thats confusing, or will click in the area to select a document and only PDFs will appear on the list.

danicatalan:
Alright I understand the issue. It would be great if we can do three things:
- we can add a sub paragraph inside the drag-drop area for letting know the user that he just can upload a PDF. In the future, we can expand this note with more file types.
- if the user drags another document, can we change the color of that paragraph as a feedback?
- if the user clicks to open the file browser, can we show only PDF files?

konzz:
@txau @danicatalan 
The copy for the paragraph? "We currently support PDF", "Supported documents: PDF"

> if the user clicks to open the file browser, can we show only PDF files?

That's the default behaviour of the browsers.

> if the user drags another document, can we change the color of that paragraph as a feedback?

I will check, not sure

danicatalan:
@konzz Supported documents: PDF.

I will improve this with this other story https://github.com/huridocs/uwazi/issues/99

danicatalan:
This will solve this and #99, @txau @konzz what do you think about it?

![drag-drop-redesign](https://cloud.githubusercontent.com/assets/6364153/15819811/b46b4efc-2be5-11e6-941e-969b96b70de1.png)

txau:
Looks good to me. I'd change "Supported documents" for "Supported formats".

txau:
@konz @danicatalan lets keep this simple. If detecting the document format becomes too complicated, lets live with just announcing the supported formats.

What would happen after the upload, conversion fail? That acceptable.

danicatalan:
[...]

PR #129, opened 2016-06-07
Add "supported formats" note below drag&drop box inside Uploads

danicatalan:
Solves https://github.com/huridocs/uwazi/issues/80 and https://github.com/huridocs/uwazi/issues/99


daneryl:
LGTM

Issue #44, opened 2016-05-30
You shouldn't be able to save templates, dictionaries and relationship types with an existing name (duplicates)


txau:

konzz:
Solved with: #156

PR #158, opened 2016-06-18
Prevent save duplicated templates

konzz:
Now backend returns an error when trying to save a template with a name that already exists, and client side validation

Solves #44


daneryl:
LGTM

PR #164, opened 2016-06-22
Returning empty object when there is no settings on the DB

konzz:
When there is no settings in the DB, the model returns an empty object to prevent errors


danicatalan:
LGTM.

Issue #184, opened 2016-07-05
can not type on the searchbar, order buttons produce js errors.


daneryl:

konzz:
Fixed with #185

PR #185, opened 2016-07-05
default state of documents list changed to new document structure, changed version of react-redux-form

daneryl:
solves #184


konzz:
LGTM

Issue #140, opened 2016-06-13
Display inbound connections in the document viewer


daneryl:

RafaPolit:
If you delete an inbound connection, there connection is not deleted from the side pane.

RafaPolit:
Connections 'Count' on top of the right connections pane is not accounting for inbound references.

PR #249, opened 2016-08-04
Render inbound connections

daneryl:
Fixes issue #140 

PR check list:

- [x] Client test
- [x] Server test
- [x] End-to-end test
- [ ] Database views added ?

QA check list:

- [x] Smoke test potential collaterals
- [ ] UI responsiveness
- [x] Tested in a browser other than chrome
- [ ] Code review ?


daneryl:
When a document references itself, the panel shows 2 references, 1 inbound and 1 outbound. discuss what we want to show, if we want this behaviour, we need to make sure both cards work independently, now the click on the card its always for the source referenced text.

daneryl:
- When deleting an inbound connection should disappear from the list and from the document.
- Outbound connections should show target text snippet, inbound should show source text snippet.

Issue #323, opened 2016-08-22
Be able to use entities as dictionaries


txau:

daneryl:
refactors:

[ ] -> change thesauri naming to dictionary
[ ] -> selectOptions endPoint are actual thesauri

txau:
Closed in #357

PR #357, opened 2016-08-26
323 entities as thesauris

konzz:
Fixes issue #323 

PR check list:

- [x] Client test
- [x] Server test
- [x] End-to-end test
- [x] Database views added ?

QA check list:

- [x] Smoke test potential collaterals
- [x] UI responsiveness
- [x] Tested in a browser other than chrome
- [ ] Code review ?


txau:
- [] You shouldn't be able to use an entity as the dictionary for a select field in that same entity
- [] There is a bug when switching templates on entity creation (more or less)
- [] An entity was created with blank info

Issue #1053, opened 2017-06-01
Allow admins to change input[type="date"] format


danicatalan:
Implements #979 + #1024. HTML ready on `1053-date-format` branch.

- [ ] Allow admin to choose 3 different styles to write dates on inputs:
`YYYY-MM-DD`
`DD-MM-YYYY`
`MM-DD-YYYY`

- [ ] Allow admin to choose 2 different styles of separators:
`/`
`-`

![screen shot 2017-06-01 at 16 45 55](https://cloud.githubusercontent.com/assets/6364153/26685254/d807889c-46e9-11e7-919f-57ec5dfe90c4.png)

RafaPolit:
Perhaps label it: Date Input Format to avoid confusion as to what exactly this format will encompass?

danicatalan:
Agree. Also like "Date input".

Issue #1054, opened 2017-06-01
Add placeholder on input[type="date"]


danicatalan:
Implements #872.

![placeholder](https://cloud.githubusercontent.com/assets/6364153/26346497/b04a6f44-3fa6-11e7-8e3b-0697f5105134.png)

This way users will know the date format they need to use in the collection - both order and separator.

PR #1069, opened 2017-06-11
date format configuration in settings

konzz:
Fixes issue #1053 #1054 

PR check list:

- [x] Client test
- [x] Server test
- [x] End-to-end test
- [ ] Update READ.me ?

QA check list:

- [ ] UI responsiveness
- [ ] Code review ?


RafaPolit:
@konzz I have added support to keep the existing format with different separator when you change only the separator.

Also, pairing with @daneryl, we removed the "state" implementation on RadioButtons and we are only using the Props right now.

If you see any potential issues with that change, please let me know.

Issue #1032, opened 2017-05-29
Implement date range field type


RafaPolit:
Japiquay has, as Cejil, requested multi date range fields.

It would be interesting if we could extend those type of properties that are universally useful to all instances.

@tuxpiper

txau:
Lets discuss in our next team call.

tuxpiper:
@RafaPolit @txau  A key question for us here regarding this:

Once the extension has been applied to our instance, would that keep compatibility with any present documents & metadata? Or would we be looking at starting from scratch with the data?

kjantin:
I think including a date range would be really useful for groups who want to use Uwazi to investigate human rights abuses, so that users can search and filter for cases that are within a date range.

txau:
@kjantin Actually date range search is already implemented in Uwazi.

What we are talking about here is to have fields that support multiple dates (ie. date start - date end). I'm not so sure about wanting that functionality for everyone. Every new field type we commit to create, is a new field type we have to maintain.

kjantin:
Sorry if I wasn't clear. I mean that having a date range field (not just the ability to search for dates within a certain range) with a start and end date would be helpful to many. There are many events that have a start and end date, not just one date, and I can imagine that capture these ranges in one field would be really helpful, and would allow someone to filter/search on "tell me all the events that were active on June 6, 2015" and it could find any event whose date range includes that date. 
As far as I understand it, that feature is not available in Uwazi now.

konzz:
We have these current fields regarding to dates:
- Date
- Multi Date (only shows in cejil)
- Multi Date range (only shows in cejil)

We want to create a Date Range field, and make the Multi Date and Multi Date Range appear in all instances?

@kjantin I would say that we close issues that are for discussing, and create issues with clear objectives referencing the discussion issue.

kjantin:
I know that a Date Range (single) field has been requested (by Japiqay) so yes let's have that appear in all instances. But I don't know about the Multi Date and Multi Date Range use cases or requests - since @RafaPolit created this issue, maybe he can provide more information/clarification?

Yes sounds good that we can close issues that are for discussing and create issues with clear objectives. I think this particular issue turned into a discussion because the description of it wasn't clear so we discussed what we thought it meant.

PR #1089, opened 2017-06-20
Daterange field

konzz:
Fixes issue #1032 

PR check list:

- [x] Client test
- [x] Server test
- [x] End-to-end test
- [ ] Update READ.me ?

QA check list:

- [ ] UI responsiveness
- [ ] Code review ?

### DIFF
--- a/app/api/search/search.js
+++ b/app/api/search/search.js
@@ -20,7 +20,7 @@ function processFiltes(filters, properties) {
     if (property.type === 'nested') {
       type = 'nested';
     }
-    if (property.type === 'multidaterange') {
+    if (property.type === 'multidaterange' || property.type === 'daterange') {
       type = 'nestedrange';
     }
     result[property.name] = {value: filters[property.name], type};
--- a/app/react/Layout/Item.js
+++ b/app/react/Layout/Item.js
@@ -31,7 +31,7 @@ export class Item extends Component {
       if (property.value && property.value !== '' || property.markdown) {
         let dlClassName = '';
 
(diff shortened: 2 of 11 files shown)


RafaPolit:
@konzz 
Bugs:
- Even when marked as 'use as filter', the option is not appended to the "Sorting buttons" to be able to sort by that property.  What will be the criteria for sorting? The from or the to? Perhaps if when asc the from and when desc the to? Not sure if that is even possible at this point.
- When the field is required there is no visual feedback as to what was wrong, it only won't allow me to save. (but see question bellow to further expand this issue)

Questions:
- What should be the expected behavior when filling only one of the dates? It reads: 16 jun, 2017 - invalid date.  I am thinking of a property, for instance, of: active period in court.  What happens if it started on Jan 1, but is it not yet finished.  I would leave the end date 'empty', but that is not an 'invalid' date.

- Thinking about the previous point, what should be the 'required' characteristics of a range? Do I need at least one date or do I need both?  In the same case as above, if that field is required, you would not be able to save it.  Perhaps one of the two is enough to fill the 'required' condition?

UX observations (@danicatalan)
- All the icons for date properties are the same.  It makes sense, but when seeing them on the templates already assigned (not on the list where their names are very clear)d, one has no idea which type they are.  Perhaps we could find a solution for that?

konzz:
@RafaPolit 
I fixed the problem that the required error didn't show, and also the "Invalid date" label, now when you don't fill one of the fields, nothing appears instead of the "Invalid date" message.

I think a date range property should not be in the sortable options, the same for the multi date field or the multi date range field, even if we decide how to sort by a range of dates, it wont be clear for the users.

@danicatalan any coment regarding the icons?

Issue #1119, opened 2017-07-11
Filters support to AND/OR logics


danicatalan:
## Update 1

We decided to implement a toggle to switch between AND/OR in each multiselect filter. The styles are on [1119-and-or](https://github.com/huridocs/uwazi/commit/13cdece0cf62c9adcdd95ee863df09133e1dc624) branch.

![img](https://user-images.githubusercontent.com/6364153/28423979-c7f73bec-6d6c-11e7-8521-7f15446f252d.png)

- - - - -

**Current behaviour**

All our filters work with OR logic.

**Expected behaviour**

Some properties doesn't make sense with OR logic. A specific example is **Descriptors** on CEJIL. They want to search documents with both properties, but know this is impossible because selecting two properties will show ALL documents with one of those properties.

**Next steps**

We should discuss how to implement AND/OR logics in our filters. Some questions to start this discussion:

- Should the admin set AND/OR by default while creating the property?
- Should the users be able to change between AND/OR depending on their needs?

danicatalan:
- Should the admin set AND/OR by default while creating the property?

We are not going to allow admins to set AND/OR by default.

- Should the users be able to change between AND/OR depending on their needs?

Yes, users should toggle between AND/OR for each property on filters.

danicatalan:
@konzz @daneryl @RafaPolit how do you want to implement this toggle?

I think something like Rightdocs would work right now:

![image](https://user-images.githubusercontent.com/6364153/28421542-c902035c-6d65-11e7-867a-469b44aeaf01.png)

txau:
AND - OR - OFF? We already have a toggle in the filters. Although I don't think the OFF is needed since normally you jut reset all filters and research o unclick/blank the filters.

danicatalan:
We thought about using the same toggle. We should remove the current one because the utility of that is not clear, what do you think?

Also, this toggle should work as a switcher, so no OFF.

![untitled-1](https://user-images.githubusercontent.com/6364153/28422962-e74f01b2-6d69-11e7-9bdf-83a0fd54a31f.png)

txau:
The only problem is that is unclear whether the AN/OR is acting internally within the options in the filter or externally between the different filters. Maybe this is not a issue and users will figure it out quickly?

danicatalan:
As they are all inside their own property, once you click on the switcher the multiselect below should update with the new options (show all if OR, show the matches if AND).

I think that will clear any doubt.

txau:
[...]

Issue #1109, opened 2017-07-10
Artículos violados filter is broken


txau:
- [ ] The internal condition should be an AND instead of an OR.
- [x] In the case of "strict mode" it should apply the strict mode condition (relation of violated articles).

danicatalan:
Add examples of expected results here.

@danicatalan

danicatalan:
These are some examples of expected results using the current collection of cases on CEJIL.

## I want to find entities that violated one specific article

- Convención Americana sobre Derechos Humanos.
  - 4.2

```
RESULTS

Hilaire, Constantine, Benjamin y otros
Boyce y otros
Raxcacó Reyes
Dacosta Cadogan
```

These entities violated the 4.2 at some point of the sentence. Easy peasy.

## I want to find entities that violated two specific articles

- Convención Americana sobre Derechos Humanos.
  - 4.2
  - 2

```
RESULTS

Hilaire, Constantine, Benjamin y otros
Raxcacó Reyes
Dacosta Cadogan
```

These entities violated the 4.2 and the 2 at some point the sentence, but not necessary together in the same line.

## I want to find entities that violated two specific articles together

- Convención Americana sobre Derechos Humanos.
  - 4.2
  - 2
- Strict mode activated.

```
RESULTS

Raxcacó Reyes
Dacosta Cadogan
```

These two entities violated the 4.2 and the 2 together, in the same line.

Yeah, _Hilaire y otros_ violated 4.2 and 2 at some point on the sentence, but with **strict mode** we search for cases where 4.2 and 2 are in the same line.

danicatalan:
We think it's better to implement #1119 first. We will review this issue again once that issue is solved.

txau:
Please keep the sprint label or we'll loose visibility on this task.

kjantin:
@txau do you want us to apply the AND/OR toggle feature (as described in #1119) to this custom filter? Or just stick with the original request of changing the internal condition to AND instead of OR?

txau:
This is custom filter that has a very particular logic, we need to keep that logic which is not exactly the same as the one described in #1119

PR #1202, opened 2017-08-02
1119 and or

konzz:
Fixes issue #1119 #1109 

PR check list:

- [x] Client test
- [x] Server test
- [x] End-to-end test
- [ ] Update READ.me ?

QA check list:

- [ ] UI responsiveness
- [ ] Code review ?

### DIFF
--- a/app/api/i18n/systemKeys.js
+++ b/app/api/i18n/systemKeys.js
@@ -65,6 +65,8 @@ export default [
 {key: 'Entity types', label: 'Entity types'},
 {key: 'Filename label', label: 'filename'},
 {key: 'Filters', label: 'Filters'},
+{key: 'Filters AND operator', label: 'AND'},
+{key: 'Filters OR operator', label: 'OR'},
 {key: 'Flag', label: 'Flag'},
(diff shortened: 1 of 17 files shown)


daneryl:
@konz found a weird case that causes all filter options to disappear, and the only way to return back is to reset the whole filter.

issue

daneryl:
@konzz found another issue that did not test before.
the and/or switch gets inverted when selecting "and" and then reload the page.

issue

konzz:
The issue @daneryl reported was related to the Control that was wraping the input, the Control is an element from react-redux-form that passes the values to the input and receives the changes, in this case was wrongly configured and was not getting the intial value when reloading.

The issue @RafaPolit reported is a problem with the "active" status of the fields, the multiselect filter is an object with key "values" (array) and key "and" (boolean), when you unselect all the options of a filters, the values of the filter gets resset to prevent sending empy arrays of options that will result in 0 matches, but in this case the "and" key did not get properly resset. Now it gets reset to OR whenever you unselect all options, it's not the ideal solution, but keeping the AND/OR status while reseting the empity array probably will require an extensive refactor of the filters form / actions / reducer.

Issue #1174, opened 2017-07-24
Elasticsearch uses the first occurrence of a property to determine its type, numbers are having issues


RafaPolit:
Elasticsearch uses the first occurrence of a property's value to determine its type.

When creating numeric values, if your first document value for that property is an integer, it maps the type to "long", if it is a decimal number, it maps the type to "double".

The problem is that once a type is set, it uses that for all the future values.  So after being registered as type "long", adding values with decimal numbers treats them as "long" for purposes of filtering and ordering.

Also, re-indexing the elastic index has the same effect: it takes the first occurrence.  So, best case scenario: it keeps all doubles and longs consistent (meaning it is still broken but at least it acts the same), and worst case scenario: it changes some doubles back to longs in which case things are even more broken than before.

We need to decide if we are going to 'manually assign' type Double to all numeric input fields, if we create a custom mapping for those fields (this will involve some name spacing or something similar) or if we should take a different route, investigating what downside it could have to index all numeric values as Doubles.

txau:
"Investigating what downside it could have to index all numeric values as Doubles" probably ony performance, if that is the case, I guess we could go for this solution on the short term.

PR #1212, opened 2017-08-08
Replaced LONG occurrences with DOUBLE, creationDate is LONG

RafaPolit:
Fixes issue #1174 

PR check list:

- [x] Rebase current DEV branch
- [x] Client test
- [x] Server test
- [x] End-to-end test
- [X] Pass code linter to client and server
- [ ] Update READ.me ?

QA check list:

- [ ] UI responsiveness
- [ ] Code review ?

### DIFF
--- a/database/elastic_mapping.js
+++ b/database/elastic_mapping.js
@@ -113,11 +113,11 @@ let config = {
           match: '*',
           match_mapping_type: 'long',
           mapping: {
-            type: 'long',
+            type: 'double',
             doc_values: true,
             fields: {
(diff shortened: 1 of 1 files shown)


RafaPolit:
@konzz 
I have forced all the numbers (by tapping into the "long" dynamic mapping, I cannot get any other mappings to occur: 0 matches as long as well, for instance, not byte as per our mapping configuration).

The important addition is that I have manually set the creationDate to LONG to avoid huge memory usage since it is the one numeric property that every document shares.

Please confirm if you think this is the way to go.

Issue #1402, opened 2017-12-04
Restrict viewing information to only logged users


AlWalid:
Actually, users can see documents even if they are not logged.
It is important to have a feature that enable hiding information for non-logged users especially when the documents are not meant to be shared publicly

kjantin:
Proposed solution: add option to "Settings" to check whether or not this instance is private. (added to all instances)
Question to @AlWalid - is it ok if all the users are able to edit all the documents (for now)?

AlWalid:
Ideally users shouldn't be able to view / edit other users documents. But it is related to #1395 
Having a new role that enables the consultation and editing of only the user document/entities

PR #1432, opened 2018-01-23
1402 private uwazi

daneryl:
Fixes issue #1402 

PR check list:

- [x] Rebase current DEV branch
- [x] Client test
- [x] Server test
- [x] End-to-end test
- [ ] Update READ.me ?

QA check list:

- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/api.js
+++ b/app/api/api.js
@@ -1,4 +1,3 @@
-import bodyParser from 'body-parser';
 import languageMiddleware from './utils/languageMiddleware';
 import elasticConfig from './config/elasticIndexes';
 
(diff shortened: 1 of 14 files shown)


konzz:
The only thing I can say is that the middleware could check if there is a user first, and if it is don't even request the settings since you can access anyway and we can avoid the extra request to the database. Anyway I don't think it will add much time.

daneryl:
@konzz but it is needed to check if the instance is configured as private, if it is not private then even if there is no user it should call next and render normally.

Issue #1608, opened 2018-04-27
Deleting a template property does not sanitize entities data.


RafaPolit:
If a template property is deleted, the DB metadata for the entities is not removed.

### To Reproduce:
- Create a regular text field
- Put data for that field on a couple of entities / documents
- Delete the property from the template
- Check that the DB still has the value for that property in the metadata

The complex scenario here is that, if someone creates a new field OF A DIFFERENT TYPE within the same template, and names it like this old field, the data types can mismatch throwing all types of errors.

### To Further Reproduce
- Create a new field WITH THE SAME NAME as the one deleted, but of a different type (for example date)
- Go into the entity, and you will see that the text has been transformed into 01/01/1970

This could lead to all sorts of errors.  I think that sanitizing the data is important here, not only to keep the DB clean, but also to prevent data missmatch in the future.

PR #1641, opened 2018-05-24
Entities metadata is now properly updated after modifying the template

konzz:
Fixes issue #1608 

PR checklist:
- [x] End-to-end test
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/entities/entities.js
+++ b/app/api/entities/entities.js
@@ -267,49 +267,46 @@ export default {
     .then(() => this.indexEntities({ sharedId: { $in: entitiesToReindex } }));
   },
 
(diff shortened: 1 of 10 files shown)


txau:
Looks like it still has some leftovers in the database. After removing the property "sex" of judges from the dev fixtures, this is how that property looks like in the different languages:

Language ES: "sexo" : null,
Language EN: property was removed
Language PT: "sexo" : null,

konzz:
@txau We need a call about this

txau:
LGTM!

Issue #1442, opened 2018-01-31
"New item"  button in thesauri


txau:
It wasn't obvious how to add items to a newly created thesauri. Add an additional button/hint in at the next blank space?

![screen shot 2018-01-30 at 19 52 51](https://user-images.githubusercontent.com/489831/35601216-751234d0-05f7-11e8-89c1-44333aa71af1.png)

Issue #1436, opened 2018-01-26
Organising thesauri items alphabetically - humphrey.uwazi


HSips:
thesauri items in settings are arranged as when added. Humphrey instance has about 600 thesauri items under themes, which makes it tedious. Can these be arranged alphabetically?

kjantin:
This has also been requested by CLEEN and others.

danicatalan:
This needs to be reviewed, because some users don't want this feature (they created thesauri with other criteria). I think it is better to allow users rearrange the list.

txau:
@danicatalan anyway dictionaries are later on ordered by number of occurrences in filters. Maybe we could add an "order alphabetically" button at the top of the list, so users can manually force that kind of ordering?

Issue #720, opened 2017-01-05
Nested thesauri


txau:
Some users have request the possibility of having nested dictionaries.

danicatalan:
We need this for Region / Cities in zorla, right? @daneryl

kjantin:
Estimate: 2 weeks (user design, development, testing, feedback)
Decision: We'll address this feature request when it is required from a partner.

danicatalan:
Minority Forum also needs this feature for their thesauri.

txau:
There are a series of thesauri related issues that we could address during this development:

https://github.com/huridocs/uwazi/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+dictionary

Issue #307, opened 2016-08-18
Automatically focus on input on new item in thesauris


txau:
It would greatly improve UX to have the focus set automatically on the input box when creating a dictionary.

![screen shot 2016-08-18 at 13 55 03](https://cloud.githubusercontent.com/assets/489831/17772886/9f2c867e-654b-11e6-9c43-6881d0ecae18.png)

kjantin:
Currently, when I click on "add item" I see the "Item name" field:
![image](https://user-images.githubusercontent.com/354426/28718320-5db38fc6-7373-11e7-965c-51bf2a6a2fd3.png)

@txau are you suggesting that the cursor be placed in that new/empty "Item name" field?

txau:
Yes.

txau:
Related: https://github.com/huridocs/uwazi/issues/1442

PR #1647, opened 2018-06-03
Dictionary groups

konzz:
Fixes issue #1442 #720 #307 #1436 

PR checklist:
- [x] End-to-end test
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


txau:
- Can't add items to a newly created dictionary.
- Trying to add a group when there are not options tosses a js error:

screenshot from 2018-06-03 09-16-31
- Tabbing from one input to another lands in the delete button instead of the next input.
- When hitting the "Add group" link, automatically focus on the input for the group title, otherwise  user don't even know where the group was added.
- Respect margins for groups. Group title should be aligned with the rest of the items, using only indentation for group's children. Also the vertical spacing should be the same as every other item.

screenshot from 2018-06-03 09-24-48

- Drag and drop for adding/moving items between groups, otherwise users have to delete and recreate the item, loosing the assigned values in the entities.
- Be able to fold/unfold the groups in settings/thesauri. This will greatly help in organizing via dragndrop, otherwise the lists are too long.

txau:
When filtering, we should not display items with value "0", as is the behavior of the rest of the filters.

txau:
This could be for a second phase, but maybe we should display groups in the options rendered in cards and/or side panel:

screenshot from 2018-06-03 10-04-53

txau:
Apparently it explodes when creating empty groups:

screenshot from 2018-06-11 09-34-07

txau:
There is a couple of CSS glitches:

screenshot from 2018-06-11 09-35-47

screenshot from 2018-06-11 09-34-55

txau:
Can't select and move items into groups =/

PR #1685, opened 2018-06-13
"hidden" config page to customize css

daneryl:
Fixes issue #1680 
to check the page go to /settings/customisation write some css update and reload.
did not include it in the menu to maintain this hidden from users, should we add it ?

PR checklist:
- [x] End-to-end test
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


danicatalan:
We can keep it hidden @daneryl until we saw its something useful for users.

txau:
@danicatalan @daneryl I think we should make it available to users right away. Give them freedom of action =)

danicatalan:
@txau @daneryl If we want to spend some time to properly adding this for all users, I think instead of using the /customization route this can go under /collection, Advanced settings with a smaller textarea (same as mailer configuration).

Issue #1812, opened 2018-08-03
Typo on ranged filters


danicatalan:
Fiilters should say **From/To** instead of **Form/To.**

![image](https://user-images.githubusercontent.com/6364153/43644811-501affe8-9730-11e8-973c-be214f59a837.png)

txau:
Fixed

OzrenMuic:
@txau this bug is still there:

![image](https://user-images.githubusercontent.com/8134850/69539874-ae721300-0f85-11ea-9d18-b12729584cea.png)

konzz:
Is not, is a matter of local traduction, that instance needs to change that label

OzrenMuic:
It is not. I've tried in few instances by adding metadata in English to the template and default label is 'Form' and not 'From'. 

Please check before closing issue.

OzrenMuic:
It is not. I've tried in few instances by adding metadata in English to the template and default label is 'Form' and not 'From'. 

Please check before closing issue.

RafaPolit:
This should be fixed by translating in the Settings -> Translations -> System section.

PR #1816, opened 2018-08-08
fix typo on Date filter label

elric-wamugu:
Fixes issue #1812 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


txau:
@konzz Taken a look to the code, I'm wondering why these strings are not available for translation in the translation interface?

konzz:
@txau I see them in my local instance in Settings -> Translations -> System.

txau:
@elric-olalekan actually this is more complicated than it looks like. There is bug with the translations system, these are the steps to reproduce:

1. Go to settings > translations > system
2. Look for the keys 
- Label date "From"
- Label date "to"
3. As you can see, the values are empty. If you try to feel in the values and save, it creates new entries in the translations such as:
- "Label date "From""
- "Label date "to""

So as you can see, its adding extra quotes, that keep stacking up.

So I guess we need to review how translation is handling these quotes and fix that bug so the typo can be fixed. 

With your current PR new instances will get that key right. But we need to fix this so current instances can be fixed.

Issue #1826, opened 2018-08-16
List "slider" is not working as expected if items are less than default count


RafaPolit:
When items inside the slider are less than the default count, all sorts of weird things are happening.

This has been reported here: https://git.huridata.org/huridocs-projects/albinism/issues/42

Already fixed, just creating the issue in the Uwazi repo.

PR #1827, opened 2018-08-16
Added logic for when children count is less than sliding value.

RafaPolit:
Fixes issue #1826 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


daneryl:
@RafaPolit fix its working fine, seems good, but i think we should test this components, there is no test at all ?

Issue #786, opened 2017-02-06
Be able to open the side panel with the metadata when a card is displayed in a page


txau:

danicatalan:
Added an image with the proposed functionality:

![screen shot 2017-02-08 at 14 54 54](https://cloud.githubusercontent.com/assets/6364153/22740083/b49717d0-ee0e-11e6-9a2e-9032aaf22f46.png)

I think this is one of the most important features in our app: the possibility to view document's metadata without enter on the document.

@txau the CSS code is already used on library, so we can start developing the functionality if we don't see any other UX implications.

PR #1837, opened 2018-08-24
ViewMetadataPanel is now loaded in pages 

konzz:
Fixes issue #786 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


daneryl:
@konzz 
found a couple of issues,

1.- sidepanel shows a tab to search and says "Type something in the search box to get some results.", this for where you are in the library, i think best think here will be to remove that tab ?

2.- delete or edit an entity does not change the values on the list

txau:
@daneryl @konzz if possible it would be great to have it work for the map markers as well.

Issue #1750, opened 2018-07-04
Library column switcher


danicatalan:
Implement column switcher on library:

![column switcher](https://user-images.githubusercontent.com/6364153/42289670-42ce7918-7fc1-11e8-9d25-e3865d99482b.gif)

danicatalan:
Idea: if one column, switch to table view.

PR #1845, opened 2018-08-28
1750 column switcher

RafaPolit:
Fixes issue #1750 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


txau:
@danicatalan @RafaPolit the feature looks great, but it collides with the map when geolocation is available. Ideally, these are the controls for the map zoom when users are in the map. Can we implement this before merging?

Issue #1909, opened 2018-09-19
Templates creation and editing routes are failing due to validation


RafaPolit:
Newly implemented route validation is failing with templates creation and editing with:
```
error: "ValidationError: child "properties" fails because ["properties" at position 0 fails because ["defaultfilter" is not allowed]]
```

RafaPolit:
This is being merged as part of #1903 develpment.

PR #1918, opened 2018-09-21
1903 pdf thumbnails

RafaPolit:
Fixes issue #1903, #1909, #1879 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/upload/PDF.js
```diff
@@ -28,6 +28,21 @@ export default class PDF extends EventEmitter {
...
+  async createThumbnail(documentId) {
+    let response;
+    try {
+      response = await spawn(
+        'pdftoppm',
+        ['-f', '1', '-singlefile', '-scale-to', '320', '-jpeg', this.filepath, path.join(path.dirname(this.filepath), documentId)],
+        { capture: ['stdout', 'stderr'] }
+      );
+    } catch (err) {
+      debugLog.debug(`Thumbnail creation error for: ${this.filepath}`);
```
txau: This probably needs to go into the errorLog instead?

daneryl: i think this should throw the error and let the route deal with it, it is catched there and there is a process already implemented for errors on PDF conversion [here](https://github.com/huridocs/uwazi/blob/1903-PDF-Thumbnails/app/api/upload/routes.js#L80) ?  or do we want to handle this in a diferent way because its only the thumbnail  ?

daneryl:
@RafaPolit reupload does not work for me, it changes locally after the upload but if i reload it is showing the previous file.

also we have this "error": "ValidationError: "originalname" is not allowed" when i try to download the file.

txau:
@RafaPolit I think there is also the migration missing? I think we should do a more intensive migration and activate thumbnails for all instances, otherwise this feature is going to pass unnoticed by most users.

Maybe we should make this a default for when a new document template is created as well (adding the preview field and activating it).

txau:
@RafaPolit I think we need to review the styles for some display sizes:

screenshot from 2018-09-24 12-53-35
.

txau:
@RafaPolit we need also to review the behaviour of the thumbnails with the search result snippets.

PR #1919, opened 2018-09-24
fixed typo on date range fields translation

konzz:
Fixes issue #1812 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


txau:
@konzz correct me if I'm wrong, but I think in this PR removing the old key is not covered? that means users are going to find two keys in the same place and won't know which to change.

More over, the problem described in the other PR (the one deleted today), explained how users could reintroduce this bug by adding template or field names with double quotes.

Issue #1995, opened 2018-10-11
Improve user feedback when attempting to download a document that doesn't exist


txau:
- Broken link (file doesn't exist)
- Wrong id
![screenshot from 2018-10-11 11-48-00](https://user-images.githubusercontent.com/489831/46795946-b05e2000-cd4b-11e8-9f14-05cb6ccf667b.png)

daneryl:
We should Throw a 404 error on any attachment that for any reason does not exists.

txau:
@daneryl @habbes unless we actually have the file registered in the database but the file does not exist in the file system. In this case a 400 will log the problem.

habbes:
@txau when the attachment exists in the database but does not exist in the file system, it currently returns a 500 error with the following message:
```
{
    "error": "Error: ENOENT: no such file or directory, stat '/path/to/uwazi/uploaded_documents/1540935218679ovspsvkgouf.jpg'"
}
```
It also logs the error

txau:
@habbes no, in this case its ok to leave the 500.

habbes:
Here's a PR #2060

PR #2060, opened 2018-10-31
Throw 404 error when downloading an attachment file that doesn't exist

habbes:
Fixes issue #1995 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


daneryl:
@habbes works well for non existent attachments in the db, in my case i tested fixtures with existent files on db but non existent file on disk and it throws another error, @txau should we also 404 this case ?

txau:
@daneryl that is correct, we want a 500 if the file is in the DB but it doesn't exist in disk.

PR #2079, opened 2018-11-14
Naive implementation of opportunistic OCR

yuvadm:
This PR is a work in progress that attemps to provide OCR processing for incoming PDF files, under the assumption that `tesseract` and `imagemagick` are installed system-wide.

Some initial thoughts / questions:
- Should OCR happen upon file upload? This might be a time consuming process.
- Should intermediate TIFF file be placed in `/tmp` (which is usually a `tmpfs` in-memory filesystem? Note that it can grow quite large in case of large input files.
- Where would be the best place to do sanity tests for the existance of tesseract and imagemagick? (Preferably one time during startup or during first OCR invocation)
- Are we comfortable with directly spawning the processes as implemented?
- Are there any UI considerations?
- Should OCR always take precedence over `pdftotext`?
- ImageMagick might not be the best tool, given that it is considered unsecure for processing PDFs (https://bugs.archlinux.org/task/60580)

Would love to hear thoughts on how to proceed.

<hr>

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/upload/PDF.js
+++ b/app/api/upload/PDF.js
@@ -3,6 +3,7 @@ import fs from 'fs';
 import path from 'path';
 import { spawn } from 'child-process-promise';
 import errorLog from 'api/log/errorLog';
+import debugLog from 'api/log/debugLog';
 
 export default class PDF extends EventEmitter {
   constructor(filepath) {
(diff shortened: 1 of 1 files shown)


txau:
Hi @yuvadm!

So I've been testing it locally and already got a few errors. I had to run the commands manually since the output is redirected to /dev/null:

> convert-im6.q16: not authorized `./FRN v HOME BUILDER NIG LTD & ANOR.pdf' @ error/constitute.c/ReadImage/412.
convert-im6.q16: no images defined `/tmp/tmp.tiff' @ error/convert.c/ConvertImageCommand/3258.

To be honest, I believe if instead of trying to use tesseract directly we used a wrapper that already handles OCR for PDF we may avoid many of the boilerplate issues. Notably, I think this solution has most of it figured out: https://github.com/jbarlow83/OCRmyPDF

That library is already in system packages via apt and brew for Ubuntu, Fedora, Debian and Mac. So its pretty easy installed.

> Should OCR happen upon file upload? This might be a time consuming process.

I think it should be a user triggered event that can happen anytime. Some Uwazi's are already hosting not OCR'ed documents so it would be great if the administrators can opt to manually run the ocr. Also, this process is pretty time and CPU intensive, and most of the PDFs being uploaded are already OCRed, so I don't think its advisable to force OCR on every PDF upload.

So ideally we add a button that will trigger the OCR. We can take care of that part if you don't want to get your hands dirty with the interface =D

> Should intermediate TIFF file be placed in /tmp (which is usually a tmpfs in-memory filesystem? Note that it can grow quite large in case of large input files.

If we use https://github.com/jbarlow83/OCRmyPDF that is handled internally, normally deleting the temp files when the process is done.

> Where would be the best place to do sanity tests for the existance of tesseract and imagemagick? (Preferably one time during startup or during first OCR invocation)

I guess the best option would be to manually activate the OCR option via ENV vars (https://github.com/huridocs/uwazi/wiki/Install-Uwazi-on-your-server), so this condition can be used also for toggling user interface options as well.

> Are we comfortable with directly spawning the processes as implemented? 

I guess with the previous conditions, manually triggered.

> Are there any UI considerations?

As explained above, a button in the side panel for triggering the OCR manually.

> Should OCR always take precedence over pdftotext?

Yes, in fact I'm wondering whether we should actually swap files and replce the file the user uploaded with the one containing the text layer on the top of it. So if users download that file they also get the text functionalities.

So ideally, without altering the current workflow, we first manipulate the PDF and add the text layer and then we pass it to Uwazi as if it was the original file for text extraction etc.

> ImageMagick might not be the best tool, given that it is considered unsecure for processing PDFs (https://bugs.archlinux.org/task/60580)

Aparently https://github.com/jbarlow83/OCRmyPDF uses Leptonica instead of ImageMagick. I wonder if it has security implications...

Issue #2085, opened 2018-11-15
Blank State Improvements


konzz:
- [ ] Create 2 default templates for entities and documents with no properties
- [ ] Show help text in the library when there are no entities to go to uploads and create new ones.

Issue #2086, opened 2018-11-15
Unify entities and documents


konzz:
Templates should be the same for entities and documents.

PR #2097, opened 2018-11-26
new blank state

konzz:
Fixes issue #2085 #2086 #2087 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


konzz:
Deleting a document after a language has been added throws an error because it tries to delete the thumbnail and it has not been created when the language was added.

RafaPolit:
Some questions regarding this:
- Uploading a PDF still reads 'No type selected', which requires to edit and save.  If we have a default, shouldn't it just be the default template assigned to an uploaded document?

Issue #2112, opened 2018-12-17
Can't save filters


txau:
![screenshot from 2018-12-17 14-02-05](https://user-images.githubusercontent.com/489831/50088822-64749f00-0204-11e9-9a71-3cb7671c310a.png)

PR #2114, opened 2019-01-07
Fixed server validation error when saving filters

konzz:
Fixes issue #2112 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


daneryl:
@konzz error still showing up for me, "index" validation should be put on the settings route, but i do not understand exactly what is the "index" property for ? maybe removing it locally before saving.

PR #2186, opened 2019-02-15
2184 attachments delete

daneryl:
Fixes issue #2184 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


RafaPolit:
There are two bugs, one certainly related to this PR, the other not necessarily or perhaps a side effect:
- When I delete an attachment, the attachment is removed from the DB, but the library still shows that attachment from Elastic Search
- I cannot download an attachment, there is a route validation error pointing out that 'file' is not allowed.

daneryl:
@RafaPolit fixed the issues, please review

Issue #2217, opened 2019-03-07
Error when publishing an entity after I changed the coordinates


kjantin:
I'm trying to create an "act" entity with a geo location on hr-reports.uwazi.io, but the coordinates I entered were not the right ones. When I tried to delete those coordinates, I was unable to publish the entity. I received this error:

![image](https://user-images.githubusercontent.com/354426/53960759-18275580-40b5-11e9-9159-96cfedd8ddd6.png)

Also, when I tried to type numbers into the first coordinates field, a white screen appeared and I couldn't do anything.

habbes:
@kjantin I've not been able to reproduce the errors.
How did you attempt to delete the coordinates? Did you try to place a marker at a new location, or enter new values directly in the latitude/longitude fields or remove the values that were there before?

habbes:
@kjantin I've been able to reproduce the errors by deleting the values in the latitude and longitude, leaving the fields empty, and then try to save. This results in an error because it doesn't delete the geolocation field, but tries to save it with empty lat and lon values, which is not allowed.

So I think the solution is to provide a way to remove the field.

txau:
@habbes how does the field handle empty values on the first place? maybe we should convert empty strings to null or similar?

PR #2228, opened 2019-03-12
Allow user to delete geolocation values from entity

habbes:
Fixes issue #2217 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Forms/components/Geolocation.js
+++ b/app/react/Forms/components/Geolocation.js
@@ -2,6 +2,10 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import Map from 'app/Map/Map';
 
+function isCoordValid(coord) {
+  return typeof coord === 'number' && !isNaN(coord);
+}
+
 export default class Geolocation extends Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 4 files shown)


habbes:
Here are the updates:
- When the user removes both latitude and longitude values from a geolocation field, the field is removed from the entity. And the entity can be saved without validation errors.
- I've added a small text button that clears the coordinates so that the user does not have to delete the field values manually

Screenshot 2019-03-12 at 16 18 59

habbes:
I also a fixed bug preventing the map from displaying coordinate (0, 0).

app/react/Forms/components/Geolocation.js
```diff
@@ -1,17 +1,28 @@
...
     this.lonChange = this.lonChange.bind(this);
     this.mapClick = this.mapClick.bind(this);
+    this.clearCoordinates = this.clearCoordinates.bind(this);
   }
 
   onChange(value) {
-    this.props.onChange(value);
+    if (!isCoordValid(value.lat) && !isCoordValid(value.lon)) {
+      this.props.onChange();
+    } else {
```
konzz: dont use else
```suggestion
     return;
    } 
```

RafaPolit:
Also, please look at the Codeclimate issues.  Some are not mandatory, but if you find a button missing a type, fixing that takes 3 seconds!  Don't leave those like that.

habbes:
Thanks @RafaPolit. I wasn't sure what `type` to add a to a button that's not used as a form submit or reset button. I guess I could use `reset` in this case. But I've just found out I can also use `button` as the type: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/button-has-type.md

habbes:
I've made changes.

Issue #2240, opened 2019-03-15
Items being duplicated in thesauri


txau:
Playing around with groups/items in thesauris can eventually duplicate entries. In the screenshot, Item 2 got duplicated after a combination of moving items inside/outside of the group:

![Screenshot from 2019-03-15 11-09-26](https://user-images.githubusercontent.com/489831/54424275-d3359b80-4712-11e9-874d-82f62cdb33bf.png)

More over, the way the items are stored in the database is not consistent for _id, ObjectId:

![Screenshot from 2019-03-15 11-07-32](https://user-images.githubusercontent.com/489831/54424365-0710c100-4713-11e9-8914-2636e72fe160.png)

habbes:
So I've noticed those issues occur when the group is the first item in the thesauri list. I think I've found the culprit code, and it's a quick-fix.

However, I've seen other issues as well:
- When you click the check icon of an item inside a group, it's not highlighted even though the item becomes selected
- There seems to be issues when deleting items inside a group. Sometimes they don't get deleted and it's items outside the group that get deleted. I'm not sure whether this bug is related to the first

txau:
@habbes great. Please see these things fixed. Make an automated test when possible.

habbes:
I've fixed the issues and added tests in this PR #2258

txau:
I added a few comments con the PR. I guess the main bug is fixed with this PR, but there is another but when moving empty items to groups. Also I think we should switch the UX to grad and drop.

PR #2258, opened 2019-03-20
Fix bug causing duplicates in thesaurus when moving or removing items

habbes:
Fixes issue #2240 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


txau:
@habbes After moving an item into a group, the next item I select does not get highlighted.

txau:
@habbes if I select the empty item from a group (the placeholder for a new item) and move it into that same group, one existing element from that group gets deleted:
1. Open a thesauri with several items
2. Create a group
3. Move a couple of items into the group
4. Select the empty element inside the group and move it inside the group
5. An item got deleted

txau:
@habbes as you pointed out, the UX is not self-evident when initially getting to thesauri. I think we should enable drag and drop for this section. What do you think @konzz ?

habbes:
@txau I've pushed some updates implementing drag-and-drop support. Could you review it and tell me whether the feature is okay? Still have some clean up and tests to add.

app/react/Thesauris/components/ThesauriForm.js
```diff
@@ -59,9 +61,10 @@ export class ThesauriForm extends Component {
     const addedValue = values.length > previousProps.thesauri.values.length;
     const lasValueIsGroup = values.length && values[values.length - 1].values;
     const previousLasValueWasGroup = previousValues.length && previousValues[previousValues.length - 1].values;
-    if (lasValueIsGroup && (!previousLasValueWasGroup || addedValue)) {
-      this.groups[this.groups.length - 1].focus();
-    }
+    // if (lasValueIsGroup && (!previousLasValueWasGroup || addedValue)) {
```
konzz: Don't leave commented code

app/react/Thesauris/components/ThesauriForm.js
```diff
@@ -96,15 +99,25 @@ export class ThesauriForm extends Component {
       movingValues.push(value);
     }
     this.setState({ movingValues });
-    console.log(movingValues);
   }
 
-  moveToGroup(groupIndex) {
-    this.props.moveValues(this.state.movingValues, groupIndex);
+  moveToGroup(values, groupIndex) {
+    console.log('moved', values, groupIndex);
```
konzz: Delete console log and commented code

habbes:
@txau I've made updates to the branch. I've rewritten the drag-and-drop logic to use the existing `DragAndDropContainer` component. It's not ready to be merged yet, it still has some console statements, obsolete code and other stuff I need to clean out, as well as add tests for the new logic.

Could you review the UX and whether it has bugs? In particular:
- check whether moving item in and out of groups works as it should (e.g. dragging an item outside the group should append it to the root list, but right before the new item field)
- ensure that the empty new item input that gets added automatically cannot be moved to a different location (it should remain at the bottom of the list or bottom of its group). This was a source of bugs and complexity so test it extensively.

habbes:
@konz I've had to make a few modifications to `DragAndDropContainer` and `DragAndDropItem` to get it to work properly on this page:
- The components now pass the item index to the `renderItem` callback
- I'm passing a reference to the original item passed to `DragAndDropItem` and using this reference instead of `monitor.getItem()` to update the list.

txau:
Some glitches:
- You can drag-select text in firefox. This is a known issue. We decided to fix this in the past (as in menu), by enabling drag and drop only for the the icon (the three lines icon).
- Too many bounding boxes.
- Icons, text box and delete buttons are not aligned.

txau:
@konzz @habbes typing is super slow on the input boxes. I think this issue was already there, but now seems to be more evident, please review if we are rerendering too much, or if duplicate checks need a delay.

txau:
Trying to drag and drop and item from the end of the group, to the first item in the group moves it outside the group.

habbes:
@txau I've made some updates:
- tweaked the UI
- improved performance when typing

I've not fixed the issue that causes an item inside of a group to be moved out of the group when it's dragged to the top of the group. I'm working on it. But I think it happens because the item is dropped on the group component itself and not a specific item inside the group. And since the group component is in the root list, the item is moved outside the group and onto the root list. So far I'm not sure how to solve it.

I haven't fixed the text drag issue yet. Not sure if it's due to my code or the code in `DragAndDropContainer`.

habbes:
@txau we've me a number of updates to the branch, including performance improvements thanks to @daneryl. Could you review the ux and functionality?

txau:
@habbes @daneryl it is still affected by our old friend bug for Firefox were you can drag-select text.

txau:
@habbes @daneryl the interaction is a little bit clumsy in some cases, but only one is particularly annoying. When dragging inside a group, or into a group, if the target area is this one, its equivalent to drag the item outside the group. I think we should fix this.

Screenshot from 2019-04-03 13-08-58

txau:
@habbes @daneryl I tried to break the functionality several times and its performing well, so I guess if you guys fix the 2 interaction issues and conduct a code review, we are good to go.

habbes:
Alright @txau, thanks.

habbes:
I believe I've fixed the text input dragging issue. But I've not fixed the second issue though. I believe I know the issue but I'm not sure what the ideal solution is. Could @konzz or @daneryl provide some insights on this?

txau:
@habbes both issues are fixed now. Thanks!

habbes:
@txau I used a bit of a hack on the issue with items going out of the group. So I'm not sure if it was ideal. I ended up setting a negative margin in the css so as to move the group items closer to the edge and to give them a larger surface area, that makes it harder to drag them out of the group by mistake.

txau:
@habbes I saw that, I think is good enough.

app/react/Thesauris/components/ThesauriFormGroup.js
```diff
@@ -0,0 +1,82 @@
...
+export class ThesauriFormGroup extends Component {
+  constructor(props) {
+    super(props);
+    this.focus = () => { this.groupInput.focus(); };
+    this.renderItem = this.renderItem.bind(this);
+    this.onChange = this.onChange.bind(this);
+    this.removeGroup = this.removeGroup.bind(this);
+  }
+
+  shouldComponentUpdate(nextProps) {
```
daneryl: @habbes we should not use shouldComponentUpdate, if this is for performance reasons consider making the component pure, connect()(Component) we should have a pure() function to do this actually, so make component pure and remove shouldComponentUpdate, also running a JSON.stringify on a shouldComponentUpdate function its going to be not that good performance wise as it can be really really slow and this is a function that gets called always before render.

daneryl: @habbes we should not use shouldComponentUpdate, if this is for performance reasons consider making the component pure, connect()(Component) we should have a pure() function to do this actually, so make component pure and remove shouldComponentUpdate, also running a JSON.stringify on a shouldComponentUpdate function its going to be not that good performance wise as it can be really really slow and this is a function that gets called always before render.

habbes: Yeah, I think they were there for performance reasons.

daneryl:
@habbes groups disappear in when reordering across another large group, in this example when i drag a group to the first position from the bottom.
Apr-05-2019 16-29-16

habbes:
@joan found the issue occurred when dragging one group into another. The drag and drop feature triggered two actions: one to move the group inside of the target group and the other to remove the group from the root list. So I think in your case the group failed to move into the new group but it was removed from the root list, which caused it to be lost.

I've fixed the issue by:
- preventing a group from being dragged out of the main list
- preventing a group from being moved into another group

txau:
@habbes @daneryl looks good to me

PR #2266, opened 2019-03-22
Inherit relationships

konzz:
Fixes issue #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


RafaPolit:
The inherited properties, when of type geolocation, are not 'adding' themselves to the library Map points (not the counter, and, apparently, not the map itself).  So, while the property is inherited, it is not going to show on the library.

RafaPolit:
The 'inheriter' is not taking into account the 'published' value of the entity.  It inherits from unpublished entities, even if logged out.  The protocol has been to somehow 'mask out' those things that are not published.  Not sure if its too critical.

RafaPolit:
The Geolocation inheritance is not taking advantage of the labeling system of the map.  So I see two different points, but I don't see the labels of those entity names, so I can disambiguate what is which.

RafaPolit:
When the template has no properties, I can select 'inherit property' and I see only a 0 as option.

RafaPolit:
when selecting the 'value' for an inherited property, I see the title of the property... not the actual values I am going to inherit.  That makes sense for certain inheritances, but for others they may be somewhat confusing.
For instance:

- I created a template named Country
- I assigned a property: President.
- I then created a template called Case
- I created an inherited property called: Current President and I inherited if from the country.

When I go to edit the Case, I have the label: Current President, and my options are: Peru, Ecuador, Colombia, Brasil.

It works nicely, it's just somewhat confusing.  No idea what a different UI would look like?  Perhaps we could expand the label and say:
- Current President, Inherited from Country:

This seems like the simpler approach. 

We could also populate the results:
- Ecuador: Lenin Moreno

But for multifields, that becomes more difficult.  The label showing that it will inherit a subproperty makes a bit more sense to me.  Perhaps its a non-issue at all?

RafaPolit:
There is a mismatch in the display of multi-value fields.

When viewing the original document, I see this:
Screenshot from 2019-03-26 11-35-40

When viewing the 'inherited' property in the other document, the multi-valued properties get a 'bulletted' list:
Screenshot from 2019-03-26 11-35-46

RafaPolit:
When the inherited property is of type geolocation, when rendered in card, it is rendering as a map instead of the coordinates.  The formatter on the inheriter is probably failing to pass along the onlyForCards value that the metadata renderer need in order to know the difference.

txau:
Inheriting a thesauri type triggered this error:

ERROR Failed to index document 58ada351299e826748545805: {
 "type": "mapper_parsing_exception",
 "reason": "object mapping for [metadata.firmantes_de_los_documentos.inherit_object] tried to parse field [null] as object, but found a concrete value"
}

txau:
Here are some improvements that we add to the backlog:
  - Allow any entity or document when configuring the inheritance.
  - What happens with the entities that have no connection type? Right now its messy with the two sided connection labels.
  - When adding a connection in the relationships view, if the added connection will display inherited properties, its not updated in real time.

app/api/entities/entities.js
```diff
@@ -56,42 +113,24 @@ function updateEntity(entity, _template) {
...
-    return Promise.all(docs.map(d => model.save(d)));
+    return Promise.all(docs.map(async _entity => inheritMetadata(_entity)))
+    .then(_docs => Promise.all(_docs.map(model.save)));
   });
 }
 
 function createEntity(doc, languages, sharedId) {
   const docs = languages.map((lang) => {
-    const langDoc = Object.assign({}, doc);
+    const langDoc = { ...doc, metadata: { ...doc.metadata } };
```
RafaPolit: Not sure how is this different from:
const langDoc = { ...doc};
Does it assign the metadata property (even if undefined) that it can be iterated later?

app/react/Metadata/components/specs/__snapshots__/Metadata.spec.js.snap
```diff
@@ -24,7 +24,7 @@ exports[`Metadata should render a Geolocation viewer when the metadata is type g
 <React.Fragment>
   <dl
     className="metadata-type-geolocation "
-    key="Geolocation Label"
+    key="undefined_0"
```
RafaPolit: Something is fishy here: previously, the label was "Geolocation  label", but upon adding the $index to the value, it registers as "undefined_0".  Either the fixtures are wrong or there is a problem with the label namespacing.

app/react/Metadata/components/Metadata.js
```diff
@@ -48,18 +52,18 @@ const removeEmptyValues = (p) => {
...
-            {showByType(prop, compact)}
-          </dd>
-        </dl>
-      );
-    })}
+    {metadata.filter(removeEmptyValues).map((prop, index) => {
+          let type = prop.type ? prop.type : 'default';
+          type = type === 'image' || type === 'media' ? 'multimedia' : type;
+          return (
+            <dl className={`metadata-type-${type} ${prop.fullWidth ? 'full-width' : ''}`} key={`${prop.name}_${index}`}>
```
RafaPolit: The previous comment probably originates here: prop.label has been replaced with prop.name.

If this is correct, the fixtures need to reflect the additional name so they don't show as undefined.
If it is not correct, please use prop.label.

app/react/Metadata/components/specs/__snapshots__/Metadata.spec.js.snap
```diff
@@ -6,7 +6,7 @@ exports[`Metadata should not render metadata without value 1`] = `
 <React.Fragment>
   <dl
     className="metadata-type-default "
-    key="Label"
+    key="undefined_0"
```
RafaPolit: Also, every index is _0.  Am I missing somewhere where it is tested that further indexes would receive a subsequent number?

app/react/Metadata/components/specs/__snapshots__/ValueList.spec.js.snap
```diff
@@ -47,27 +45,19 @@ exports[`ValueList should render the values as a ul list 1`] = `
   className="multiline"
 >
   <li
-    key="labelfirst_value"
+    key="NaN"
```
RafaPolit: All this NaNs are also probably a indicators that fixtures require changing, or at least reviewing.  These had values on the previous version.

RafaPolit:
As noted, there are several snaps that previously held values and now are undeined_0 or NaN, or other incoherent values.

This is the danger with snaps, that we would just hit the 'update' command and not fix what is required.  If that snap is testing that the correct label or key or value is returned by the myriad of possible type fields, having all with NaN or Undefined is really hazardous, even if the code itself is correct.

Please either fix the fixtures, or ensure that the code is not requesting the wrong property.

app/api/entities/entities.js
konzz: When you do ` { ...doc};` you are cloning the object doc, but the property metadata points to the same metadata object as the original, so if you modify anything inside metadata you are also modifying the original.

PR #2298, opened 2019-04-04
Inherit geolocation map route

RafaPolit:
Fixes issue #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/search/search.js
+++ b/app/api/search/search.js
@@ -299,10 +299,10 @@ const getInheritedEntitiesData = async (toFetchByTemplate, language) => Promise.
   ))
 );
 
-const getInheritedEntities = async (baseResults, language) => {
+const getInheritedEntities = async (results, language) => {
   const templates = await templatesModel.get();
   const templatesInheritedProperties = determineInheritedProperties(templates);
-  const toFetchByTemplate = whatToFetchByTemplate(baseResults, templatesInheritedProperties);
+  const toFetchByTemplate = whatToFetchByTemplate(results, templatesInheritedProperties);
   const inheritedEntitiesData = await getInheritedEntitiesData(toFetchByTemplate, language);
 
   const inheritedEntities = inheritedEntitiesData.reduce((_memo, templateEntities) => {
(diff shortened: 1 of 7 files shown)


app/api/search/search.js
```diff
@@ -105,16 +105,30 @@ function snippetsFromSearchHit(hit) {
...
     template.properties.forEach((prop) => {
       if (prop.type === 'geolocation') {
         geolocationProperties.push(prop.name);
       }
+
+      if (prop.type === 'relationship' && prop.inherit) {
+        const contentTemplate = templates.find(t => t._id.toString() === prop.content.toString());
+        const inheritedProperty = contentTemplate.properties.find(p => p._id.toString() === prop.inheritProperty.toString());
+        if (inheritedProperty.type === 'geolocation') {
+          inheritedGeolocationProperties.push(prop.name);
```
konzz: Why not push directly to geolocationProperties? so you don't need to concat or process them in different map functions for the select properties part.

app/api/search/search.js
```diff
@@ -147,96 +161,193 @@ const processResponse = (response) => {
...
+    .then(processResponse)
+    .catch(() => {
+      throw createError('Query error', 400);
+    });
+  });
+};
 
-      if (query.geolocation) {
-        searchGeolocation(documentsQuery, filteringTypes, templates);
+const determineInheritedProperties = templates => templates.reduce((memo, template) => {
```
konzz: why not just call the argument memo inheritedProperties?
```suggestion
const determineInheritedProperties = templates => templates.reduce((inheritedProperties , template) => {

```

app/api/search/search.js
```diff
@@ -147,96 +161,193 @@ const processResponse = (response) => {
...
+  const inheritedProperties = memo;
```
konzz:
```suggestion

```

konzz: Why not call the argument memo inheritedProperties ?

app/api/search/search.js
```diff
@@ -147,96 +161,193 @@ const processResponse = (response) => {
...
+);
+
+const getInheritedEntities = async (results, language) => {
+  const templates = await templatesModel.get();
+  const templatesInheritedProperties = determineInheritedProperties(templates);
+  const toFetchByTemplate = whatToFetchByTemplate(results, templatesInheritedProperties);
+  const inheritedEntitiesData = await getInheritedEntitiesData(toFetchByTemplate, language);
+
+  const inheritedEntities = inheritedEntitiesData.reduce((_memo, templateEntities) => {
+    const memo = _memo;
```
konzz: is this just to avoid the linter error?

app/api/search/search.js
```diff
@@ -147,96 +161,193 @@ const processResponse = (response) => {
...
+    });
+  });
+
+  return toFetchByTemplate;
+};
+
+const getInheritedEntitiesData = async (toFetchByTemplate, language) => Promise.all(
+  Object.keys(toFetchByTemplate).map(t => entities.get(
+    { language, sharedId: { $in: toFetchByTemplate[t].entities } },
+    { ...toFetchByTemplate[t].properties.reduce((memo, n) => Object.assign(memo, { [`metadata.${n}`]: 1 }), {}), sharedId: 1 }
```
konzz: is sharedId hardcoded to 1 intended?

app/api/search/search.js
RafaPolit: Can we discuss this? @joan  Yeah, memo is inherited properties.  Memo in a reduce is DESIGNED to be modified.  But our eslint rules prevent that from happening.  So I do this first assignment line to prevent the warning.  What do we prefer?

app/api/search/search.js
RafaPolit: I agree.  At first, the inherited Geolocation properties had more info.  I was expecting to fit more logic into this stage, but then I decided to move that into the main function.  Here, it can be simplified.

app/api/search/search.js
RafaPolit: Yeap. :(

app/api/search/search.js
RafaPolit: This is the 'project' part of the get.  It's to just fetch the shared ID and the expected metadata and not all the other info in entities.  Its:
entities.get({ _id: { $in: [1,2,3]} }, {'metadata.prop1': 1, 'metadata.prop2': 1, sharedId: 1})

txau:
Getting an error when editing an entity with two inherited properties:

Screenshot from 2019-04-08 13-35-09

txau:
There is an odd behavior that may need some thought: when an entity has two inherited properties from the same relation/target entity, editing that entity and adding/removing connected entities will act erratically.

txau:
Inherited fields disappear when editing:

1. Go to the entity view of an entity with inherited properties.
2. From this view, edit the entity.
3. Save.

The inherited fields disappear.

txau:
When editing the options in settings > templates for an inherited property, if I have a library open in another browser tab it automatically crashes with this error:

Screenshot from 2019-04-08 13-48-27

This also happened to me once without an extra tab opened, just navigating from settings to library after tweaking the config, but I haven't been able to replicate the bug.

txau:
Inheriting a thesauri based property is not showing the values.

txau:
Multidate-range is also not being inherited.

txau:
When working with two tabs the rendering of the entity that has inherited properties gets updated automatically when changing the configuration of its template, which is great. Ie. when changing the property that the entity is inheriting, the inherited value gets updated automatically.

But when changing also the target template for inheritance, the automatic update fails and you get a blank value until refresh. This is not a critical bug, but it can mislead the users and maybe is a smell for something else going on.

PR #2309, opened 2019-04-09
Prevent direct access to entity files

### LINKED ISSUE #2224: Insufficient document validation
Uwazi does not sufficiently validate documents by their type and mimetype. In effect, Uwazi does not
restrict the type of documents that can be uploaded by a user with editor privileges. However, some
file content (such as HTML and JavaScript) is inherently dangerous because it may be rendered in the
user’s browser.

habbes:
fixes #2224

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


daneryl:
@habbes LGM, rebase it with dev pls and ill merge

daneryl:
srry i still think the approach  of not returning the filenames complicates things a lot, having to partial update entities manually etc and having to add manual select for filenames when needed all around the code, what you think @RafaPolit  @txau  @habbes  , i made the qa and it works, its just i do not feel the approach is right: https://github.com/huridocs/uwazi/pull/2309

txau:
@daneryl you make the decision. If you really think this is not the right approach, lets do the other implementation.

habbes:
@daneryl I also had a similar concern. I don't know much about whether the alternative will be better though.

daneryl:
@habbes the other proposed solution was to have separate folders for uploaded_documents and custom uploads and have the first one not accessible, i think its a lot simpler than this does not pollute the code, at this point ill prefer this one

habbes:
@daneryl I think @RafaPolit had mentioned that it could require a migration (not sure how much of an impact that is), will that be an issue?

txau:
@habbes it will require a migration and maybe some changes in the deployment process. I guess oyu are going to need some guidance to do this.

habbes:
@txau yes, sure

daneryl:
closing in favor of #2322

PR #2311, opened 2019-04-10
Added validation to templates to prevent duplicated relationship fields

### LINKED ISSUE #2306: Prevent creating a relationship property (in template) with same type to same template
There should not be an option to add a second relationship in a template that:
- has the same type AND points to the same template type
- if there is a type that has ANY, then you cannot have a relationship to a specific type configured
- if you already have a TYPE and TEMPLATE, you cannot assign a new TYPE to ANY neither.

This causes issues when editing the metadata fields.

konzz:
fixes #2306 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/templates/routes.js
+++ b/app/api/templates/routes.js
@@ -1,7 +1,6 @@
 import Joi from 'joi';
 
 import settings from 'api/settings';
-
 import { validateRequest } from '../utils';
 import needsAuthorization from '../auth/authMiddleware';
 import templates from './templates';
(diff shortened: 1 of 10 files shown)


daneryl:
@konzz i think validations should be implemented first at server side and then in the client as a ui friendly helper.

txau:
@daneryl definitely. It makes sense to also keep the API consistent.

daneryl:
@konzz 

an error appears the moment i add a relationship field, is this intended ?
also when server validation fails it returns a 500 internal error instead of a validation error.

PR #2319, opened 2019-04-16
requesting entity before editing

### LINKED ISSUE #2253: Possible issue when saving from the Library side panel which sends ELASTICSEARCH data
I think we have a conceptual error regarding saving procedures in the Library:

- The Library gets its data from Elastic Search to create the cards, but it loads all the entity's info to populate the side panel when a card is clicked as well.
- When one clicks a card, one is still seeing the Elastic Search data in the side panel.
- When one clicks on EDIT, it is the Elastic Search data that is loaded into the forms, AND NOT the Database data.
- When you save that form, you are sending Elastic Search data into the Database instead of the original database content.

If, for whatever reason, something failed in an Elastic indexation, we are corrupting the Database data with the incomplete / corrupted Elastic data.

On the other hand, if we ever want to store slightly different data in Elastic than the one in the DB, we have an issue.  We may want auto-generated fields (like inheritances), or post-processed fields, etc.  This may cause an issue.

I think it is desirable to have the side panel show the Elastic data, but when the user clicks on EDIT I strongly believe we should load the DB data to ensure there are no inconsistencies.  This reduces the probability of the user sending old data from an old query into an entity until we have some form of version control or optimistic offline lock.

What do you guys think @txau @daneryl @konzz

konzz:
fixes #2253

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Metadata/actions/actions.js
+++ b/app/react/Metadata/actions/actions.js
@@ -40,26 +40,26 @@ const resetMetadata = (metadata, template, options, previousTemplate) => {
   return resetedMetadata;
 };
 
-export function loadInReduxForm(form, onlyReadEntity, templates) {
+export function loadInReduxForm(form, _entity, templates) {
   return (dispatch) => {
-    const entity = Object.assign({}, onlyReadEntity);
+    api.get(_entity.sharedId)
+    .then(([entity]) => {
+      const sortedTemplates = advancedSort(templates, { property: 'name' });
+      const defaultTemplate = sortedTemplates.find(t => t.default);
+      if (!entity.template && defaultTemplate) {
+        entity.template = defaultTemplate._id;
+      }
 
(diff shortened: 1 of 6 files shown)


app/react/Metadata/actions/actions.js
```diff
@@ -40,26 +40,26 @@ const resetMetadata = (metadata, template, options, previousTemplate) => {
   return resetedMetadata;
 };
 
-export function loadInReduxForm(form, onlyReadEntity, templates) {
+export function loadInReduxForm(form, _entity, templates) {
```
daneryl: @konzz works ok, but i think we should fix eslint warning on the code we touch at least, pls fix / refactor/fix lints on the action

PR #2322, opened 2019-04-17
Separate documents and custom uploads and make documents folder inaccessible

### LINKED ISSUE #2224: Insufficient document validation

habbes:
fixes #2224 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,5 +17,8 @@ app/api/upload/specs/uploads/*
 !app/api/upload/specs/uploads/spn.pdf
 app/api/attachments/specs/uploads/toDelete.txt
 uploaded_documents/*
+custom_uploads/*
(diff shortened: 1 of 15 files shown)


daneryl:
@habbes delete for custom uploads throws an error, i guess its because its using the old folder, the rest LGTM

PR #2331, opened 2019-04-24
Another attempt at opportunistic OCR

yuvadm:
This is a continuation of #2079 which adds opportunistic PDF processing using the suggested `ocrmypdf` tool.

Same comments as #2079 apply, if this is the right direction I'll continue to add any missing parts.


txau:
@yuvadm we are a little but busy at the moment, but I'll review this with a developer as soon as I have a chance. Thanks so much!

PR #2340, opened 2019-05-03
language, update and file support for csv import

### LINKED ISSUE #2310: Import CSV 2.0
- [x] Add an icon in uploads for importing CSV file
- [x] Halt on first error and provides feedback
- [x] Add support for multiple languages (add id and lang columns)
- [x] Support updates when Uwazi ids are provided
- [x] Support file uploads (upload a zip)
- [ ] validate zip has only one csv
- [x] When I don't have a lang, I use the default lang to create the records
- [ ] ~~Add an icon for importing thesauri in a thesauri~~

daneryl:
partially fixes #2310 

import script basic usage:

yarn import-csv will show help on how to use the script,

by converntion there are several header reserverd names, 
**title**: this is required, will be the title of the entity
**id**: in case you want to update existing entities (sharedid)
**file**: to assign a main file to the entity, this needs to be compressed in a zip with the csv and all files needed

 every other field is treated as metadata if configured in the instance, thesauri and template to relate to need to be also created in the instance.

to translate properties the convetion is propertyName__languageCode, this only works on title, text fields and file

example csv:

```
Title__es,  Title__en,  text label,    select label,  file__en,     file__es,
title1_es,  title1_en,  text value 1,  thesauri1,     pdf1_en.pdf,  pdf1_es.pdf,
title2_es,  title2_en,  text value 2,  thesauri2,     pdf2_en.pdf,  ,
title3_es,  title3_en,  text value 3,  thesauri2,     pdf3_en.pdf,  ,



```


RafaPolit:
When dealing with multiple languages, some multiselects get assigned one value per each language.  We are looking into this.

daneryl:
@RafaPolit, @konzz just in case, relationships and thesauri, do NOT support import with multiple languages, if that is the problem you are facing ? translation of those should be done on separate processes.

konzz:
@daneryl It is exactly happening when we add title in other languages, the relationships get duplicated. Thanks for clarifying it.

daneryl:
@konzz @RafaPolit aparently it is duplicating the metadata ids but not the relationships itself ? relationship parser should take into account duplicated values on the entity and return a unique array probably ??

konzz:
@daneryl I've created a test for the relationships parsers to work when there are other properties in different languages, could you review it?

daneryl:
@konzz LGTM

PR #2341, opened 2019-05-06
Proper error logging for failed pdf conversions

### LINKED ISSUE #2226: Better error reporting on wrong file format
When uploading a malformed PDF or if the file gets wrongly truncated during upload (ie, due to a socket error), the error reported is just "Undefined". Improve this messaging.

habbes:
fixes #2226

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/upload/PDF.js
+++ b/app/api/upload/PDF.js
@@ -16,25 +16,34 @@ export default class PDF extends EventEmitter {
   }
 
   async extractText() {
-    const result = await spawn('pdftotext', [this.filepath, '-'], { capture: ['stdout', 'stderr'] });
-    const pages = result.stdout.split('\f').slice(0, -1);
-    return {
-      fullText: pages.reduce(
-        (memo, page, index) => ({
-          ...memo,
-          [index + 1]: page.replace(/(\S+)(\s?)/g, `$1[[${index + 1}]]$2`)
-        }),
-        {}
-      ),
-      fullTextWithoutPages: pages.reduce(
-        (memo, page, index) => ({
-          ...memo,
-          [index + 1]: page
-        }),
-        {}
-      ),
-      totalPages: pages.length
-    };
+    try {
+      const result = await spawn('pdftotext', [this.filepath, '-'], { capture: ['stdout', 'stderr'] });
+      const pages = result.stdout.split('\f').slice(0, -1);
+      return {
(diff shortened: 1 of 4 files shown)


habbes:
@jaume I've fixed the error log so that it includes a proper error message, taken from the `pdftotext` process's `stderror` output.

Here's what it looks like now:

```
2019-05-06T10:48:39.069Z [localhost] `pdftotext /Users/habbes/code/huridocs/uwazi/uploaded_documents/15571397181794kqyw57rlha.pdf -` failed with code 1
stderr output:
Syntax Warning: May not be a PDF file (continuing anyway)
Syntax Error: Couldn't find trailer dictionary
Syntax Error: Couldn't find trailer dictionary
Syntax Error: Couldn't read xref table
```

habbes:
This is ready for QA

habbes:
I was confused by the following block of code in the `uploadProcess` function in the `api/upload/routes.js` module:

```js
 .catch((err) => {
    errorLog.error(err.error);
    debugLog.debug(err.error);

    // ... other stuff
  });
```
Specifically, what type of object is the `err` object there? Why is `err.error` being passed to the error and debug log functions? Why not `err.message`?

What I ended up doing was intercept the error thrown by the `PDF.extractText` method in `PDF.js`, generated an error message based on the stderr output of the child process and then added it as an `error` field of the error, i.e. I assumed `err.error` to be the error message and just made it a copy of `err.message`.

daneryl:
@habbes yeah, that does not seem right, it should be only err in my opinion, so the other error/log functions have the knowledge of diferent types of error if necessary, can you change it to only err ?

habbes:
@daneryl I had tried to do `new Error('...')`, but that did not work. It still resulted in `undefined` being logged. I will try using the `createError` function and see if it helps.

habbes:
@daneryl @RafaPolit I've changed the loggers to `errorLog.error(err)` instead of `errorLog.error(err.error)`. I've also used `createError()` to create the error instead of `new Error`. Now it works. It seems the logger works accepts strings or plain objects with a `message` property but not work with `Error` instances.

daneryl:
LGTM !

PR #2343, opened 2019-05-07
Display 'no options found' message in empty Multiselect fields

### LINKED ISSUE #2270: When editing / creating an entity, relationship fields pointing to templates that have no entities gives no feedback
When editing or creating an entity or document for a template that has a relationship field pointing to a particular type of template... if that template has no valid entities assigned, the form shows the label and no options to select, nor any feedback that there are no entities of that type to select from.  Its confusing.

Provide feedback that 'no entities of type X found'.

habbes:
fixes #2270

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Forms/components/MultiSelect.js
+++ b/app/react/Forms/components/MultiSelect.js
@@ -215,8 +215,17 @@ export default class MultiSelect extends Component {
     );
   }
 
+  renderInfoForEmptyOptions() {
+    const { sourceName } = this.props;
+    const info = sourceName ? `No items of type '${sourceName}' found.` :
+      'No options found.';
+    return (
+      <span>{ info }</span>
+    );
+  }
+
   render() {
-    let { optionsValue, optionsLabel } = this.props;
+    let { optionsValue, optionsLabel, sourceName } = this.props;
     optionsValue = optionsValue || 'value';
     optionsLabel = optionsLabel || 'label';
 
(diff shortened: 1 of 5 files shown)


habbes:
I set the message displayed without using the `t` translation utility. I'm not sure how to handle messages with variables (in this case, the template or thesaurus name) with `t`.

app/react/Forms/components/MultiSelect.js
```diff
@@ -215,6 +215,15 @@ export default class MultiSelect extends Component {
     );
   }
 
+  renderInfoForEmptyOptions() {
+    const { sourceName } = this.props;
+    const info = sourceName ? `No items of type '${sourceName}' found.` :
```
konzz: This text should be translated using the <Translate> component, but also the var sourceName is the name of a thesaurus that has its own translation. 
Is `No items of type '${sourceName}' found.` a requirement of the issue? If not I'll suggest to remove it so we reduce complexity in the component and also avoid the problem with the translation.

habbes: I don't think it's a hard requirement, but it was specified in the issue. Without it, it may not be immediately obvious to the user which thesauri or template the multiple options are coming from until they go check in the settings.
But I can change it back to `No options found` if the convenience is not worth the complexity.

habbes: I've changed it to 'No options found'.

Issue #687, opened 2016-11-29
When navigating to a text reference in a document from another document it should scroll down to the reference


daneryl:

vasyugan:
I was't aware of this bug report, sorry for the duplication in #2237  . But I would have questions why this is "low priority". My understanding is that the ability to have cross references is one of the core features of Uwazi. 
This is one of the things I am really struggling with at the moment, trying to use Uwazi to build up a collection of quotes from various documents about the same subject. Very often, when reading the quote one will want  to quickly check out the context, and at this moment this takes the step of finding it in PDF documents which in our case are often over a hundred pages.

RafaPolit:
- [ ] The Text reference should also be a URL

vasyugan:
Why has this been closed? Last I checked, clicking a reference still doesn't take one to the referenced paragraph but only to the beginning of the document.

txau:
@vasyugan while double checking this functionality I realized we have a bug. I'm opening a new issue (https://github.com/huridocs/uwazi/issues/3016) and keep track of this requirement on that other issue as well.

PR #2354, opened 2019-05-27
Scroll to text reference when document opened in relationships view

### LINKED ISSUE #2237: "view" should take one to the referenced paragraph, not just the document
Here is a reference I created:
Screenshot_2019-03-14 Free • FPIC
When I click "view", I would expect to be taken to the referenced paragraph, but what I end up is this:
Screenshot_2019-03-14 PROTOCOLO CP Y CPLI DEL PUEBLO NASA CERRO TIJERAS • Page 1 • FPIC
So, that's the beginning of the document, not the referenced paragraph. That's quite impractical.

habbes:
fixes #2237 and #687

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Library/components/Doc.js
+++ b/app/react/Library/components/Doc.js
@@ -57,7 +57,7 @@ export class Doc extends Component {
   }
 
   render() {
-    const { className, additionalText } = this.props;
+    const { className, additionalText, targetReference } = this.props;
     const doc = this.props.doc.toJS();
     const { sharedId, file, processed } = doc;
 
@@ -66,7 +66,7 @@ export class Doc extends Component {
       itemConnections = this.getConnections(doc.connections);
     }
 
-    const buttons = (<div><ViewDocButton file={file} sharedId={sharedId} processed={processed} storeKey={this.props.storeKey}/></div>);
+    const buttons = (<div><ViewDocButton file={file} sharedId={sharedId} processed={processed} storeKey={this.props.storeKey} targetReference={targetReference}/></div>);
 
     return (
       <Item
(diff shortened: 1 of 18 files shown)


habbes:
Instead of using a timeout delay to scroll the document to the location in the document containing the text reference, I added a `documentLoaded` event when the document is ready in the document viewer. When clicking the "View" button in the text reference from another document, the target document will be loaded, and the `documentLoaded` event is used to know when to start scrolling.

daneryl:
@habbes working good, what do you think it will take to change this to a url state instead of a redux state ? this way we do not need the events, simpler code in theory, and text references can be linked and shared, @txau @RafaPolit

RafaPolit:
We are going to implement links for the text relationships, so it is shareable with others.

habbes:
@joan @RafaPolit I've refactored the feature to use url parameters that contain information necessary to scroll to the text reference. It's ready for QA.

daneryl:
@habbes, @RafaPolit i thinks we should use only reference id as a url parameter, is there a reason to use multiple parameters on the url to do this ?

habbes:
@daneryl the function that does the scrolling uses the range start to decided which page to scroll to. I don't know if there reference is in the store when the page is loaded, in that case maybe we could just use the id to look for the correct reference in the store to get the additional properties.

daneryl:
@habbes take a look and check if you can do it that way pls.

habbes:
@daneryl, I've pushed an update. Now you only need to pass the reference id in the url (using the `ref` query parameter).

PR #2359, opened 2019-05-31
Public form

### LINKED ISSUE #1392: Idea: Submission of documents (and other info) for non-authorised users
So far, only authorised users can submit a document, including associated metadata.

Would it be possible to submit a document as a non-authorised user, adding metadata which would not necessarily be made public (information about submitter, including contact details) upon approval of document.

Use case: https://minorityforum.uwazi.io/en/page/l24m3lni1m8n4fz9gz80n3ik9

konzz:
fixes #1392 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


daneryl:
@konz some issues i found:

1.- putting the wrong number in the captcha and submitting results on a success message (the response from the server is an error and the entity is not created so only the feedback is wrong)

2.- when no Main File field or the field is empty, the form does not get reset after submit and the request stays at 'pending'

3.- both main file and attachments have the same label "File"

daneryl:
@konzz the issues prevoiously found are all fixed, found another one, all entities are created with the same entity template, even if the PublicForm is configured with another one, in the case of the testing fixtures always as "Mecanismo", is this intended ?

app/react/Markdown/components/PublicForm.js
```diff
@@ -0,0 +1,120 @@
...
+import Immutable from 'immutable';
+import { MetadataFormFields, validator } from 'app/Metadata';
+import { LocalForm, actions, Control } from 'react-redux-form';
+import { Captcha } from 'app/ReactReduxForms';
+import { Translate } from 'app/I18N';
+import { publicSubmit } from 'app/Uploads/actions/uploadsActions';
+import { bindActionCreators } from 'redux';
+import { FormGroup } from 'app/Forms';
+
+export class PublicForm extends Component {
```
RafaPolit: I like it that the test is using WrappedComponent to access the connected exported component.  It that is the case, why the need to export the base component?

app/react/Markdown/components/PublicForm.js
```diff
@@ -0,0 +1,120 @@
...
+
+PublicForm.propTypes = {
+  file: PropTypes.bool.isRequired,
+  attachments: PropTypes.bool.isRequired,
+  template: PropTypes.instanceOf(Immutable.Map).isRequired,
+  thesauris: PropTypes.instanceOf(Immutable.List).isRequired,
+  submit: PropTypes.func.isRequired,
+};
+
+export const mapStateToProps = (state, props) => ({
```
RafaPolit: I think there is enough 'logic' in this to merit a test.  Also, since these functions are usually exported only for testing purposes, I was expecting to find such tests.

RafaPolit:
## Functionality Issues

- [ ] The form is only partially being reset after send.  The fields that are not being reset are:
- Number fields
- Date
- Date Range
- Multi Date
- Multi Date Range

- [ ] Links are not being interpreted correctly, they are stored in the DB as [object Object]

- [ ] Range date fields are not being stored correctly either (neither single, nor multi), but it is working for non-ranged dates.

- [ ] I am having errors with the Geolocation field getting registered as string instead of Array.  Probably the same [object Object] issue as from links?

- [ ] Attachments are not being attached

## Usability Issues

- I like the fun aspect of the question "Are you a robot?", but will answering correctly prove I am one?  The question should probably be something like: "Are you not a robot?, answer this simple math operation".  I don't know, something more 'clear', even though I love the idea. 
- We have been using the word 'document' to signal the 'main' file of an entity, yet here the form lists it as 'File'.
- Just like with the "import CSV", the file input for both File and Attachments is left to the system's representation.  WIth the CSV I thought it not really nice, but it's a very 'admin' issue, so it can live like that.  But this is a very public form, could we use the same masking we have used for the entity edit?

konzz:
All the functionality issues should be fixed now.
About the "usability issues":
- Renamed Captcha label to "Captcha".
- Renamed File to "Document".
- I don't think using masking for the file input is good, actually, our masking is bad, it's just an icon instead of a label and the input itself is hidden, it looks nice (?), but public users are not used to it since they have no access to edit entities, also the  mask hides witch file you have selected, or how many in the case of the attachments field. I know it doesn't look nice, but I rather leave the standard browser file input than put a mask that you select a file but there is no feedback that you selected anything. We also could work in a Mask that shows what file you have selected, but I think it's outside of the scope of this issue. And I think that this same argument applies to the import CSV.

konzz:
## Public Form
The public form allows any Uwazi visitor submit a form that will create an unpublished entity. You'll need to first define a Template for this form, and then you can use the form in any page with the following code

```
<PublicForm template="ID_OF_THE_TEMPLATE" />
```

Optionally you can add file or attachments field to the form like this:

```
<PublicForm template="ID_OF_THE_TEMPLATE" file attachments />
```

RafaPolit:
> We also could work in a Mask that shows what file you have selected, but I think it's outside of the scope of this issue. And I think that this same argument applies to the import CSV.
I agree!

PR #2371, opened 2019-06-11
geolocation parser for csv import

daneryl:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/csv/typeParsers.js
+++ b/app/api/csv/typeParsers.js
@@ -1,9 +1,12 @@
 import entities from 'api/entities';
 import thesauris from 'api/thesauris';
-import { unique, emptyString } from 'api/utils/filters';
 import url from 'url';
 
+import { unique, emptyString } from 'api/utils/filters';
+import geolocation from './typeParsers/geolocation.js';
+
 export default {
+  geolocation,
   async default(entityToImport, templateProperty) {
     return this.text(entityToImport, templateProperty);
   },
--- a/app/api/csv/typeParsers/geolocation.js
+++ b/app/api/csv/typeParsers/geolocation.js
@@ -1,0 +1,10 @@
+const geolocation = async (entityToImport, templateProperty) => {
+  const [lat, lon] = entityToImport[templateProperty.name].split('|');
+  if (lat && lon) {
+    return [{ lat, lon }];
+  }
+
+  return null;
+};
+
+export default geolocation;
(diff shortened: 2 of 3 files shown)


app/api/csv/typeParsers/geolocation.js
```diff
@@ -0,0 +1,10 @@
+const geolocation = (entityToImport, templateProperty) => {
+  const [lat, lon] = entityToImport[templateProperty.name].split('|');
+  if (lat && lon) {
```
RafaPolit: Well, there is actually a place on earth with lat: 0 and lon: 0.  The chances are slim, but if anyone is using the geographic 0, 0... this will return null.

app/api/csv/typeParsers.js
```diff
@@ -1,9 +1,12 @@
 import entities from 'api/entities';
 import thesauris from 'api/thesauris';
-import { unique, emptyString } from 'api/utils/filters';
 import url from 'url';
 
+import { unique, emptyString } from 'api/utils/filters';
+import geolocation from './typeParsers/geolocation.js';
+
 export default {
+  geolocation,
```
RafaPolit: All other types are async functions, yet this one is not.  Will the "importEntity" line 13 that 'awaits' for this not fail here?
https://javascript.info/async-await

app/api/csv/typeParsers/geolocation.js
RafaPolit: Actually, any lat OR lon at 0 will return null, so chances are much less slim... this needs correcting.

app/api/csv/typeParsers.js
RafaPolit: Perhaps the missing implementation test on the typeParsers might solve the doubt?

daneryl: i think you can await any function as the tests demonstrates, but i will make it an async function for clarity

app/api/csv/typeParsers/geolocation.js
daneryl: @RafaPolit this was working properly because '0' evaluates to true, anyway good catch, i wrote a a test anyways to demonstrate.

PR #2376, opened 2019-06-13
Throwing an error when there is an indexation error instead of a console

konzz:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


daneryl:
@konzz after testing this i realised that this is stoping the whole indexation process, that is why we had there the console log, i do not think we want that ? we should use  errorLog.error(err) or     debugLog.debug(err), what you think ?

daneryl:
@konzz there is failing tests on search.spec

PR #2404, opened 2019-07-01
Throwing an error when trying to save a template with swaped names

### LINKED ISSUE #2349: Swap property names in a template throws a mongo error
Swapping the names of two existing properties in a template throws an error like:
`MongoError: Cannot update 'metadata.people' and 'metadata.people' at the same time`.

It's not clear what problems this may cause, but the template does get updated but probably not the entities metadata.

konzz:
fixes #2349 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/templates/templates.js
+++ b/app/api/templates/templates.js
@@ -2,6 +2,7 @@ import entities from 'api/entities';
 import request from 'shared/JSONRequest.js';
 import translations from 'api/i18n/translations';
 import validateTemplate from 'api/templates/validateTemplate';
+import createError from 'api/utils/Error';
 
 import { db_url as dbURL } from '../config/database.js';
 import { generateNamesAndIds, getUpdatedNames, getDeletedProperties } from './utils';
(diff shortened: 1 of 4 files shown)


app/api/templates/templates.js
```diff
@@ -49,43 +50,57 @@ const updateTranslation = (currentTemplate, template) => {
...
+      const templateContentProperties = template.properties.filter(p => p.content);
+      const toRemoveValues = {};
+      currentTemplateContentProperties.forEach((prop) => {
+        const sameProperty = templateContentProperties.find(p => p.id === prop.id);
+        if (sameProperty && sameProperty.content !== prop.content) {
+          toRemoveValues[sameProperty.name] = prop.type === 'multiselect' ? [] : '';
+        }
+      });
+      if (Object.keys(toRemoveValues).length === 0) {
+        return;
```
RafaPolit: I think that the consistent returns is a good practice, not only for eslint purposes.  If lit 98 is returning something, this line should also return something other than 'undefined'.  If removeValuesFromEntities is asynchronous, perhaps Promise.resolve() is enough.  This is much clear when you refactor functions to use async-await.

RafaPolit:
Try: just delete a property and recerate it

Does it still fail or give an error?

(this is a note to self)

daneryl:
@RafaPolit 

> Try: just delete a property and recerate it

tried that and still gives the error, is this the expected behaviour @konzz ? if yes LGTM

RafaPolit:
This has been talked between the group. We are not completely happy with throwing that error in some cases where it could be worked around, but it is a remote case.

Current solution: just don't save the template, and do the actual changes you need.

PR #2407, opened 2019-07-03
Charts enhancements

daneryl:
PieChart support to show labels instead of tooltip via showLabels="true" prop
PieChart accepts innerRadius and outerRadius props
PieChart automatically filters '0' values
PieChart/BarChart support for max categories to show via maxCategories prop
BarChart vertical layout support via layout='vertical' prop

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Markdown/components/BarChart.js
+++ b/app/react/Markdown/components/BarChart.js
@@ -3,7 +3,15 @@ import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 import Immutable from 'immutable';
 
-import { ResponsiveContainer, BarChart, XAxis, YAxis, CartesianGrid, Bar, Tooltip } from 'recharts';
+import {
+  ResponsiveContainer,
+  BarChart,
+  XAxis,
+  YAxis,
+  CartesianGrid,
+  Bar,
+  Tooltip
+} from 'recharts';
 
 import { t } from 'app/I18N';
 import { populateOptions } from 'app/Library/helpers/libraryFilters';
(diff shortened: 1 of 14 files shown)


app/react/Markdown/components/PieChart.js
```diff
@@ -60,17 +60,25 @@ export const PieChartComponent = (props) => {
 
 PieChartComponent.defaultProps = {
   context: 'System',
+  innerRadius: '0',
+  outerRadius: '105',
   classname: '',
   colors: '#ffcc00,#ffd633,#ffe066,#ffeb99,#fff5cc',
   data: null,
+  showLabel: 'false',
```
RafaPolit: Mmmm... I think I agree with having this default as false, but then that would 'break' the existing Pies.  Any thoughts about that? Would it merit to revert the default?

app/react/Markdown/components/BarChart.js
```diff
@@ -51,21 +65,25 @@ export const BarChartComponent = (props) => {
 
 BarChartComponent.defaultProps = {
   context: 'System',
+  layout: 'horizontal',
+  maxCategories: '0',
```
RafaPolit: This will certainly break every pie and chart out there.  Should we default to something 'useful'?

RafaPolit:
- Append the 'others' at the end of the array
- If maxCategories has more than the values, it works, but appends a 0 value after the array (a quick fix is to move the zero sanitizing after the appending)

daneryl:
@RafaPolit proposed fixes implemented

PR #2413, opened 2019-07-08
ListChart and BarChart filterZero prop support.

daneryl:
PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Markdown/components/ListChart.js
+++ b/app/react/Markdown/components/ListChart.js
@@ -11,13 +11,18 @@ import MarkdownLink from './MarkdownLink';
 import markdownDatasets from '../markdownDatasets';
 
 export const ListChartComponent = (props) => {
-  const { property, data, classname, context, colors, thesauris } = props;
+  const { excludeZero, property, data, classname, context, colors, thesauris } = props;
   const sliceColors = colors.split(',');
 
   let output = <Loader/>;
 
   if (data) {
-    const formattedData = arrayUtils.sortValues(arrayUtils.formatDataForChart(data, property, thesauris, { context }));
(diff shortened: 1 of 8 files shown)


app/react/Markdown/components/ListChart.js
```diff
@@ -11,13 +11,18 @@ import MarkdownLink from './MarkdownLink';
...
+  const { filterZero, property, data, classname, context, colors, thesauris } = props;
   const sliceColors = colors.split(',');
 
   let output = <Loader/>;
 
   if (data) {
-    const formattedData = arrayUtils.sortValues(arrayUtils.formatDataForChart(data, property, thesauris, { context }));
+    const formattedData = arrayUtils.sortValues(
+      arrayUtils.formatDataForChart(data, property, thesauris, {
+        filterZero: Boolean(filterZero),
```
RafaPolit: The weird thing about this is that using a <ListChart filterZero="false"> in the client would actually eval to true.  But we have this in other places as well.  Not sure what should be the approach (if any?) to fix this.

daneryl: yeah its weird, we can not pass booleans from the markdown so for booleans in this components anything is true except not present, good enough for me, the only reason to put there the prop should be because you want it true i think ?

RafaPolit: Agreed.  As I said, this is not the only place this issue is relevant.  Just to keep in mind in case we ever come up with a more universal solution.

PR #2417, opened 2019-07-08
Allow the label of "Title" field to be changed

### LINKED ISSUE #2333: Allow admin to change the label of "Title" on entities
I am working on an Uwazi instance that has an entity called "Person". I'd like to be able to change the "Title" field to "Name" so that it's clear to the person entering the data what kind of info goes there (and to avoid having 2 fields - one for title and one for name).

habbes:
fixes #2333

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -124,7 +124,7 @@
       "jsx-a11y/no-static-element-interactions": "off",
       "jsx-a11y/click-events-have-key-events": "off",
       "jsx-a11y/mouse-events-have-key-events": "off",
-      "jsx-a11y/label-has-associated-control": ["warn"]
+      "jsx-a11y/label-has-associated-control": "off"
     },
     "overrides": [
     {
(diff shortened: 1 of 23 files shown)


habbes:
@RafaPolit 

- I've used the title's label as both the key and value in the translation context
- The title will be added to the template's translation context when the template is created or updated 
- I've tested how it works with existing templates that don't have the title in the template's context: you can't translate the title until you update the template. If it's acceptable for users to have to update their templates in order to translate the title label, then I guess we don't need a migration, otherwise we do.
- I think I should add validation to prevent the user from creating a field that has the same name as the title field

daneryl:
@habbes fix the lint on codeclimate pls.
there is a place where the 'Title' label still appears even if you changed it, when you do a search:
Screenshot 2019-07-16 at 09 15 06

did not find more of these, apart from that LGTM

daneryl:
@habbes same thing happening on the search but now its showing only the original name not the translations, caan you check ?

habbes:
Thanks @joan, I've fixed that too.

daneryl:
@habbes pls fix the cognitive complexity issue on codeclimate.

PR #2419, opened 2019-07-10
Remove unpublished entities from template thesauri

### LINKED ISSUE #2324: Relationship field shows unpublished entities.
The relationship field shows relationships with unpublished entities, and since it renders a link to those entities you can see all the info.

Inheriting information is based on this field, so anything related to inheriting would have this problem too.

related #2313

habbes:
fixes #2324

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/thesauris/specs/thesauris.spec.js
+++ b/app/api/thesauris/specs/thesauris.spec.js
@@ -37,12 +37,11 @@ describe('thesauris', () => {
       expect(dictionaties[4].type).toBe('template');
     });
 
-    it('should return all thesauris including unpublished documents if user', async () => {
-      const dictionaties = await thesauris.get(null, 'es', 'user');
+    it('should return only published documents for as options for template thesauri', async () => {
+      const dictionaties = await thesauris.get(null, 'es');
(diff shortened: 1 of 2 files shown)


habbes:
This fix simply excludes unpublished entities when the server send the thesauri list via `GET /api/thesauris`. This means the user won't see the unpublished entities as options for relationship fields when creating/updating an entity.

However, it has the following limitations:
- it does not cater for existing entities which already have relationship fields with unpublished entities. Or what happens when an entity is unpublished after entities have already been created pointing to it
- a script creating entities via the API directly (bypassing the UI) could simply add ids of unpublished entities in the relationship field, since there's no server-side validation preventing that from happening

These concerns have been discussed in more detail in the related issue (#2324).

We have to decided to discuss the concerns with the rest of the team before moving forward with this task.

habbes:
It is intended behaviour for unpublished entities to be included in the thesauri options when the user making the request is authenticated. This update made in this PR is therefore wrong.

PR #2421, opened 2019-07-11
Allow template to be used as its own relationship target

### LINKED ISSUE #2271: Be able to select the same template as target for a relationship field
If I am configuring template A, I could potentially have a relationship field that binds a relationship to entities of the same template.

For instance, if I have a template called "Person of interest", I could have two people that have a connection between them: one a victim, one a perpetrator, or configure them as accomplices.

But I cannot select 'myself' as the target template of a relationship metadata field.

habbes:
fixes #2271 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Templates/components/FormConfigRelationship.js
+++ b/app/react/Templates/components/FormConfigRelationship.js
@@ -19,7 +19,7 @@ export class FormConfigRelationship extends Component {
     const { index, data, formState, type, templates, relationTypes } = this.props;
     const property = data.properties[index];
 
-    const options = templates.toJS().filter(template => template._id !== data._id);
+    const options = templates.toJS();//.filter(template => template._id !== data._id);
     const labelError = formState.$form.errors[`properties.${index}.label.required`] || formState.$form.errors[`properties.${index}.label.duplicated`];
     const relationTypeError = formState.$form.errors[`properties.${index}.relationType.required`] && formState.$form.submitFailed;
     const inheritPropertyError = formState.$form.errors[`properties.${index}.inheritProperty.required`] && formState.$form.submitFailed;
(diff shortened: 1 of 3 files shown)


habbes:
So I went for an implementation that turned out to be close to effortless, just including the current template as an option in the relationship field in the template creator. [Here's the PR](https://github.com/huridocs/uwazi/pull/2421).

In the process I noticed the following:

- When creating relationships from an entity's relationships page rather than the template form, it was already possible to create relationships between entities of the same template, including an entity with itself
- Relationships between an entity and itself are not displayed on the graph. I'm looking into this. But I do have a question, the graph displays the current entity on the left, and other entities it's related to on the right. So in this case of a relationship with self, the entity should be displayed both at the left and right?
- Assuming A, B and C are entities of the same template, which have a relationship field to the same template. I create A with links to B and C, create B with links to C and create C without any links to other entities. When I publish these entities, for some reason, all of them end up having B and C as values of their relationship field. I don't know why. When I edit them and correct them after having published them, they seem to behave correctly. I'm investigating this issue.

RafaPolit:
The third point is, probably, the reason why we did this in the first place.  @txau ?  Let's talk about this.

RafaPolit:
Issue two is being recorded in: #2426

txau:
@RafaPolit maybe we did that so we don't have to look into the problem. But as @habbes stated:

> When I edit them and correct them after having published them, they seem to behave correctly.

So I suppose there has to be a way of doing this properly since the beginning.

habbes:
@txau 

> When I edit them and correct them after having published them, they seem to behave correctly.

Actually, after investigating the problem more I found that the correction was only temporary. The problem would surface again when you save another entity of the same templates. All the entities in that template end up having the same values/relationships in the relation field. It's also not triggered by publishing entities as I previously thought, the issue occurs when you save the entity, whether or not its published. I think it has something to do with the way hubs and relationships are synced to the metadata, somehow all the entities end up synced together.

RafaPolit:
Lets wait till we figure out the relationships paradox issue with same template assignment before merging this.

PR #2423, opened 2019-07-15
csv thesauri

### LINKED ISSUE #843: Allow users to select from existing HURIDOCS micro-thesauri for Uwazi "dictionaries"
How can we integrate HURIDOCS' existing micro-thesauri into Uwazi to make it easier for users to create Uwazi "dictionaries"? These are the ones that currently exist:

    Huridocs Index Terms
    Violations Typology
    Rights Typology
    Types of Acts
    Methods of Violence
    International Instruments
    Counting Units
    Civil Status
    Education
    Occupations (ILO Categories)
    Physical Descriptors
    Religions
    Ethnic Groups
    Languages
    Geographical Terms
    Types of Source Material
    Types of Locations
    Degrees of Involvement
    Source Connection to Information
    Types of Intervention
    Types of Relationships
    Types of Chain of Events
    Relevant Characteristics
    Types of Perpetrators
    Status as Victim
    Status as Perpetrator
    Types of Responses
    Attribution
    Types of Detention
    Whereabouts and Outside Contact During Detention
    Legal Counsel
    Types of Courts
    Types of Language Used in Court
    Autopsy Results
    Death Certificate
    Statements Signed
    Medical Attention
    Intent
    Sex
    Types of Addresses
    Violation Status
    Reliability
    Monitoring Status
    Impact on the Situation
    Intervention Status
    Priority
    Compensation
    Types of Dates
    Sexual Orientation

All available in Google spreadsheets, and in multiple languages: https://www.huridocs.org/resource/micro-thesauri/

daneryl:
fixes #

@txau, @RafaPolit, this implementation does not include UI, it works by using the command yarn import-csv, it can be tested using any of [huridocs micro-thesauri](https://www.huridocs.org/resource/micro-thesauri/), for them to work it is needed to remove the first row (the description) and export them to csv.

should i also implement UI on this PR ? i suggest we create another issue for that as the ux needs a litle discussion.

example:
```
yarn import-csv -f FilePath -t ThesauriId -T

### CODE REVIEW

**Comment by txau** [comment]
@daneryl ok, we can do it in another PR

**Comment by RafaPolit** [comment]
For QA:
- Test what happens if you import into a thesauri that already has values and what happens if those values have been assigned to certain entities.
- Check for duplication (of existing values) and duplicated values coming from the CSV itself

File: database/import_csv.js [hunk 1]
```diff
@@ -2,10 +2,16 @@ import CSVLoader from 'api/csv';
...
-const { template, username, language, file, stop } = require('yargs') // eslint-disable-line
+const { template, importThesauri, username, language, file, stop } = require('yargs') // eslint-disable-line
 .option('template', {
   alias: 't',
-  describe: '_id of a template',
+  describe: '_id of a template or thesauri',
+})
+.option('importThesauri', {
+  alias: 'T',
+  describe: 'flag to imprt a thesauri csv instead of entities',
```
habbes: There's a typo here, `imprt` instead of `import`

habbes:
@joan @daneryl  I've tested the branch

## what happens if you import into a thesauri that already has values?
All the existing thesauri values disappear and they're replaced by the ones imported from the csv

### what if the existing thesauri values were assigned to entities prior to the import?
It appears that the thesauri values are removed from the entities

## duplicates within the csv
If the CSV contains duplicate values, the thesaurus will contain duplicate values

## duplicate values in some, but not all of the languages
If there are two rows that have the same values in some, but not all of the languages, i.e. one of the language has different values in the two rows, then the duplicates will be imported like in the previous case, but in the case of the language without duplicates, one of the distinct value  will overwrite the other. This is likely due to the fact they use the first language (or English, I'm not sure) as the language key, with is the same in this case.

Example: assuming we import the following csv data into a thesaurus

English | Spanish | Portuguese
--------|---------|--------------
Value (en) | Value (es) | Value (pt)
Value (en) | Value (es) | Different value (pt)

When create an entity in the Portuguese language, you'll see `Different value (pt)` listed twice in the thesauri, and `Value (pt)` won't be shown

## what happens when a row has empty values in some language columns?
The thesauri will display empty values in those languages, but the value will be displayed correctly in the language where it is available

## what if there are rows with empty values across all columns?
The thesauri will include the empty values


**PS**: When I initially imported the "Types of acts" micro-thesaurus from Huridocs micro-thesauri collection, it was not translating Portuguese values. The reason was because the micro-thesauri spreadsheets use ~~**Portugese**~~ as the column name instead of **Portuguese**. Just  thought I'd mention this instead someone else gets confused by this.

I am not very familiar with the requirements of this feature so I cannot tell whether the cases I have reported are bugs or expected behaviour.

daneryl:
@habbes @RafaPolit 

we did not have a clear requeriments for these details, lets talk about it.

i think everything basically here needs a fix:
 - do not delete previous values on the thesauri on import.
 - because of the key being used for the translations im assuming we can not allow duplicates.
 - fix [Object object] on translations when language column is empty.
 - get rid of values that are empty.
 - we have a list of languages that i we use to do the matching, they have to match exactly, im assuming here no typos.

do we agree on these changes ? you guys have any more suggestions ?

daneryl:
>  - fix [Object object] on translations when language column is empty.
>  - get rid of values that are empty.

this is not that simple, this happens when a value on the language that is being used as main language is empty, as @habbes said this will result on empty translation keys. the options here are either validate every single value on the language that is being used to import is not blank and it has no duplicated values, after this, and then its not that much of a problem having the translations with duplicated values, technically speaking.

habbes:
@daneryl I agree with your suggestions.

How does Uwazi determine the main language? Is it always "English", or is it configurable? Or is it just the currently active language?

daneryl:
@habbes with the cli tool you can pass the language, by default is english, the idea when this has a interface will be to use the language you are in as the main language.

daneryl:
- [ ]  - do not delete previous values on the thesauri on import.
- [ ]  - do not allow blank values
- [ ]  - do not allow duplicated values only on main language

RafaPolit:
Awaiting improvements on the Thesauri model to prevent blanks within the model itself, not in the import feature.

Please check if we are allowing empty translations of phrases.  If this is the case, we need to open a issue about that.

daneryl:
> Awaiting improvements on the Thesauri model to prevent blanks within the model itself, not in the import feature.

this is ready, im using [ajv](https://github.com/epoberezkin/ajv) hoping we can use this to start adding more complex validations to models, supports async validations so it is possible to compare things to already existing data in the database.

> Please check if we are allowing empty translations of phrases. If this is the case, we need to open a issue about that.

it id definitely allowed, but not from the translation route as we have some basic type validation on the route level, we should move this to model validations? should we have both ? @RafaPolit @habbes @konzz

habbes:
@joan when you talk about model validation are you referring to validation at the level of Mongoose models or validation in methods like `thesauris.save()`?

In any case, if the API routes are not the only methods  by which data is entered, then I think it's best to prioritize validation in the method that serves as the common entry point. For example, if you use `thesauris.save()` for both the `POST /thesauris` route and the method for CSV imports, then I would suggest having the validation directly applied to the arguments of the `thesauris.save()`. I think this is more important than the the API route validation. It's okay to have both I guess, but if the request body from the API is passed to the `thesauris.save()` method as-is/unchanged, then it feels redundant. But in the case where the API route manipulates the request body before passing it to the underlying method, then maybe in that case it makes sense to have 2 layers of validation.

daneryl:
@habbes, yes im talking about our own model validation on the common entry points, this was implemented on `thesaris.save()`, i agree that we should apply it only there to not have it redundant, will change this for the thesauri, for some other routes there is no "model" but some params like _id or language, for those the route validation is still useful, my proposal is we have both libraries now, Joi to validate on the route level the types of the params and ajv to validate data models on the model layer, we can migrate this slowly as we go.

habbes:
@joan that proposal sounds good to me.

Issue #2353, opened 2019-05-20
CSV import UI v2.0


RafaPolit:
- [ ] Decide what to do in responsive view.  Right now, both the create entity and upload CSV buttons are not showing in cellphone-size view
- [ ] Upgrade the file select input to an Uwazi button (like the add file for attachments) rather than defaulting to system-styled input
- [ ] Actually create the instructions, as it takes to an empty wiki page

PR #2424, opened 2019-07-15
when importing a zip, csv should be import.csv

### LINKED ISSUE #2310: Import CSV 2.0

daneryl:
fixes #2310 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


habbes:
@daneryl I've tested again and now it seems to work fine.

I found the icon used for the error message to be strange, but I don't think that's a major issue
Screenshot 2019-07-19 at 15 31 12

I also tried using a file named `import.csv` that was actually an image and not an a csv file and it showed the following error:
Screenshot 2019-07-19 at 15 34 53

I guess that's okay as well.

And finally, I created a template with a required field and used imported a csv file that does not have  that field. The import worked without reporting any issues. And the entities were create successfully, without the required field. Of course I don't think this is an issue with the import feature, I just think it means we don't have server-side validation of entities against the template schema.

If the error icon and the handling of non-csv file are note considerable issues, then this feature looks good to me.

daneryl:
@habbes, thanks, there is an issue to work on the ui for the import csv #2353, maybe we should include the icon that you mention, anyways that a generic, the only thing i wanted to fix here is the no handling errors situation. specific entities validation etc its a whole other topic.

habbes:
@daneryl then in that it's good enough for me.

daneryl:
@habbes, @RafaPolit, @txau  maybe a small validation to check if the csv its actually a csv ? but dont know if we can do that properly

PR #2441, opened 2019-07-22
Query, Repeat, Value and EntityLink components for Pages

konzz:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


konzz:
Proposed help:

### Query component:
This component allows you to do API request to fetch data you may need, for example:

`<Query name="entities" url="search?limit=10&order=desc&sort=creationDate"/>`

Will search for the last 10 entities created and will store them in the "entities" dataset to be consumed. You can access this dataset with `Value` or `Repeat` for example.

### Value component:
This component passes a given value to his child components or prints it if it has no children. For example:
```
<Query name="entities" url="search?limit=10&order=desc&sort=creationDate"/>
<Value path="entities.rows.0.title"/>
```
This example will request 10 entities and then print the title of the first one.

### Repeat component:
The purpose of this component is to iterate over data and print his contents for each entry, for example:

```
<Query name="entities" url="search?limit=10&order=desc&sort=creationDate"/>
<ul>
   <Repeat path="entities.rows">
     <li><Value path="title"/></li>
   </Repeat>
</ul>
```

WIP

konzz:
@RafaPolit @txau @daneryl
### EntityLink implementation:

Right now EntityLink is a very basic component, that expects an Entity object in the property key "entity" to generate a Link to that Entity. Because of this reason, when used in the Pages it needs to be wrapped in a Value component that is connected to the context and passes the entity down, like in the example:

```
<Value propkey="entity"><EntityLink>Some text</EntityLink></Value>
```

Another option is to adopt the philosophy that new components that we develop for the pages are aware of the context and can grab values directly, but this won't work with already existing components in Uwazi, or be able to use these new components like EntityLink in other parts of Uwazi that are not pages.

RafaPolit:
We have agreed that we are going to assume components have a context available to them.  We will later on develop decorators to allow components to be used outside pages.

txau:
@konzz I think it would be better to put the help in the issue instead of the PR

PR #2443, opened 2019-07-24
deleting a template deletes also references on other template properties

### LINKED ISSUE #2429: Cannot delete Templates used as relationships targets without any feedback
Templates that have been assigned as Relationship fields in other templates cannot be deleted.  Still, there is no input to the user as to what they have to do in order to get the template deleted.

The question is: how much feedback does the user need?  Do we tell them where it's being used and how?  Does this include listing all the entities that may be assigned to this template?  Is it more a 'generic' info and the user needs to find out where.  Something in the lines of: "The template is being used as relationship target in other templates". 

This is somewhat less useful, but could be much easier to implement that first and then evolve into the more detailed feedback?

daneryl:
fixes #2429 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/templates/specs/fixtures.js
+++ b/app/api/templates/specs/fixtures.js
@@ -1,40 +1,74 @@
 /* eslint-disable */
 import db from 'api/utils/testing_db';
+import { templateTypes } from 'shared/templateTypes';
+
 const templateToBeEditedId = db.id();
 const templateToBeDeleted = '589af97080fc0b23471d67f1';
 const templateWithContents = db.id();
 const swapTemplate = db.id();
 export default {
   templates: [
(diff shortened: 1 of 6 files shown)


RafaPolit:
Deleting the template now 'frees' the pointing property to point to template 'any'.  I think this opens up a lot of possible errors.

We should actually delete the template's property.

daneryl:
> Deleting the template now 'frees' the pointing property to point to template 'any'. I think this opens up a lot of possible errors.
> 
> We should actually delete the template's property.

@RafaPolit done pls check.

Issue #2313, opened 2019-04-11
Inheriting geolocation shows unpublished items in the map


txau:
The library map is showing a couple of non-public geolocation points that inherited by a public entity, even if logged out. 

I'm not sure if this is the desired behavior, ideally these points are not shown in the public map.

@RafaPolit @konzz

txau:
Selecting "Any entity" will cause this bug in the configuration that leads to a library error:

![Screenshot from 2019-04-11 19-27-16](https://user-images.githubusercontent.com/489831/55977964-df6c3480-5c8f-11e9-8954-01c71ff45e19.png)

PR #2445, opened 2019-07-24
Remove private entities from inherited properties when the user is not logged in

### LINKED ISSUE #2324: Relationship field shows unpublished entities.

habbes:
fixes #2324  #2313 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/entities/routes.js
+++ b/app/api/entities/routes.js
@@ -79,6 +79,10 @@ export default (app) => {
           res.json({});
           return;
         }
+        if (!req.user && _entities[0].relationships) {
+          const entity = _entities[0];
+          entity.relationships = entity.relationships.filter(rel => rel.entityData.published);
+        }
         res.json({ rows: _entities });
       })
       .catch(next);
--- a/app/api/entities/specs/routes.spec.js
+++ b/app/api/entities/specs/routes.spec.js
@@ -206,6 +206,37 @@ describe('entities', () => {
         .catch(catchErrors(done));
       });
     });
+
+    describe('when the document has unpublished relationships and the user is unauthenticated', () => {
+      it('should remove private relationships from response', async () => {
+        spyOn(entities, 'getWithRelationships').and.returnValue(Promise.resolve([
+          {
+            sharedId: 'e1',
+            published: true,
+            relationships: [
+              { entityData: { sharedId: 'e1', published: true } },
+              { entityData: { sharedId: 'e2', published: false } },
+              { entityData: { sharedId: 'e3', published: true } }
+            ]
+          }
+        ]));
+
+        const req = {
+          query: { _id: 'sharedId' },
+          language: 'lang'
+        };
+
+        const { rows: [result] } = await routes.get('/api/entities', req);
+        expect(result).toEqual({
+          sharedId: 'e1',
+          published: true,
+          relationships: [
+            { entityData: { sharedId: 'e1', published: true } },
+            { entityData: { sharedId: 'e3', published: true } }
+          ]
+        });
+      });
+    });
   });
 
   // describe('/api/entities/get_raw_page', () => {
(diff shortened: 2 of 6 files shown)


habbes:
The updates excludes private entities from being included in inherited metadata when search for entities with geolocation (for the map view) if the user is not logged in. It also removed private relationships when fetching an entity for the entity viewer.

habbes:
There's one outstanding issue:
The document count on the map view still counts entities with inherited geo fields from private entities, even if the private entities are not included in the payload. I think this happens because the search made elastic search returned all entities with inherited geo fields. I'm not sure how to solve this without having to resort to checking all the search results and ensuring that non of their geo fields are empty before returning the response to the client.

Screenshot 2019-07-24 at 16 36 48

habbes:
I fixed the issue with the document count in the following way:
- I do some post-processing of the geolocation search results before sending them from the server. This post-processing removes all the entities which do not have a geolocations (includes those which have empty inherited geolocations)
- I manually set the `totalRows` of the results to match the length of `rows` array. This is because the geolocation search is designed to return all the rows and not just a subset, so it's expected that `totalRows` will be equal to the size of the `rows` array.

I had discussed this approach with @joan and we did not find a better way to solve this. This is because the elastic search query cannot filter out inherited properties based on whether the target entities are private or public, these entities are only fetched after the elastic search query is complete.

daneryl:
@habbes can not test it, blank state + creating and publishing a single document results in an error.

`TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at entityHasGeolocation (/Users/daneryl/workspace/uwazi/app/api/search/search.js:329:12)
    at results.rows.forEach._row (/Users/daneryl/workspace/uwazi/app/api/search/search.js:361:9)
    at Array.forEach (<anonymous>)
    at processGeolocationResults (/Users/daneryl/workspace/uwazi/app/api/search/search.js:344:16)
    at Object.searchGeolocations (/Users/daneryl/workspace/uwazi/app/api/search/search.js:379:17)
    at process._tickCallback (internal/process/next_tick.js:68:7)`

habbes:
@daneryl could you explain more about the document you created? I tried to blank-state and created then published a document but did not see any errors. Looking at the error message, it seems likely the it encountered an entity that does not have `metadata`. But I don't know how that could be the case if you started from a blank state. If the issue is a potential missing `metadata` field, then I can fix that but I can't tell for sure yet cause I've not managed to reproduce it.

daneryl:
@habbes yes sorry, the error occurs after publishing the document and going to the library.

the Document is a default one only with title.

habbes:
@daneryl I have been able to reproduce it, and it does indeed seem that the entity with the default Document template is created without `metadata` field

Screenshot 2019-07-26 at 17 56 37

daneryl:
@habbes that means any entity can be created without metadata so we probably have those already on live instances, check for the existence of the property for now.

habbes:
I was wondering why this entity was part of the geolocation search results for elastic search. I've found out why. The geolocation search finds all geolocation fields by scanning all templates properties, and then it filters the elastic query to include only entities with those fields. But since the Uwazi is a blank-state, it has not templates with geolocation fields, so it has no fields to filter out. This means that it returns all entities, including those without metadata.

daneryl:
@habbes ok that sounds terrible :( can we fix quick ? returning an empty array if the case is we do not have templates with geolocation ?

habbes:
@joan I guess that's a better fix. I fixed by removing entities that don't have metadata since I was already removing entities that don't have geolocations (during post-processing).

But I do have some recurring failed e2e tests, so those are probably not flaky tests. I'm going to look into them as well.

habbes:
@daneryl the tests passed (I didn't change anything).

PR #2449, opened 2019-07-26
do not allow nesting groups on filters

### LINKED ISSUE #2448: Group Filters are not working
Groups in filters are not working as expected. Te reproduce:
- Create a group
- Create a second group
- Drag de second group inside the first

At this point, you can no longer delete the child group, nor the parent one.  You also cannot edit the names anymore of child groups.

daneryl:
fixes #2448 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Settings/components/FiltersForm.js
+++ b/app/react/Settings/components/FiltersForm.js
@@ -28,6 +28,16 @@ export class FiltersForm extends Component {
   }
 
   activesChange(items) {
+    items.forEach((item) => {
+      // eslint-disable-next-line
+      item.items = item.items.filter((subitem) => {
+        if (subitem.items) {
+          items.push(subitem);
+          return false;
+        }
+        return true;
+      });
+    });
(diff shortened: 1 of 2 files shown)


RafaPolit:
Nice!  What will happen to an instance that already has subnested filters and that cannot delete them?  Should we migrate or what other approach do you suggest @daneryl ?

daneryl:
> Nice! What will happen to an instance that already has subnested filters and that cannot delete them? Should we migrate or what other approach do you suggest @daneryl ?

going to qa that scenario before we think about a migration.

daneryl:
@RafaPolit if they have that scenario the filters will not show on the filters anyways, it can be fixed dragging the nested group outside of the parent group, should we migrate this ?

PR #2504, opened 2019-09-03
Sort thesauri filters by aggregated count

### LINKED ISSUE #2481: Aggregations for Thesauri no longer displaying ordered from higher to lower values
Aggregations for Thesauri-based properties, no longer show the aggregated results in order from higher count to lower count when the group is collapsed.

Recently, a patch was made to allow for alphabetical display of certain results, perhaps this broke the display for thesauri values when collapsed?  Entity-based aggregations are behaving as expected.

habbes:
fixes #2481

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Forms/components/MultiSelect.js
+++ b/app/react/Forms/components/MultiSelect.js
@@ -86,7 +86,7 @@ export default class MultiSelect extends Component {
   }
 
   sort(options, optionsValue, optionsLabel, isSubGroup = false) {
-    let sortedOptions = options.sort((a, b) => {
+    const sortedOptions = options.sort((a, b) => {
       let sorting = 0;
       if (!this.state.showAll) {
         sorting = (this.checked(b) || this.anyChildChecked(b)) - (this.checked(a) || this.anyChildChecked(a));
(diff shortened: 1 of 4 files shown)


habbes:
I probably broke the sorting when I removed sorting for thesauri select fields. It actually broke tests (both unit and e2e) at that time, but I updated the tests not being aware that I was supposed to preserve aggregation-based sorting on the library filters.

So it was not the case that we didn't have e2e tests covering that. But I have gone ahead and added more unit and e2e tests that specifically test the sorting order.

habbes:
Here is the specification I have attempted to adhere to:

## Thesauri filters
- they are sorted by descending aggregation count
- if two items have the same aggregation count, they are sorted alphabetically using their labels
- if there are selected items, they are displayed first
- when the list is expanded, selected items are not moved to the top

## Thesauri multi-select fields when creating/updating entities
- they are sorted in the same order they appear in the thesauri settings.

## Entities filters (relationship fields)
- they are sorted by descending aggregation count by default
- alphabetical order is used as tie breaker
- if there are selected values, they are displayed first
- when the filter list is expanded (by clicking the show more button), they are sorted alphabetically regardless of their aggregation count or whether they're selected

## Entities multiselect (relationship) fields when creating/updating entities
- they are sorted alphabetically

RafaPolit:
LGTM

PR #2506, opened 2019-09-04
Document reload (server-side render) and Document reupload issues

konzz:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Metadata/actions/actions.js
+++ b/app/react/Metadata/actions/actions.js
@@ -97,7 +97,7 @@ export function reuploadDocument(docId, file, docSharedId, __reducerKey) {
     .on('response', ({ body }) => {
       const _file = { filename: body.filename, size: body.size, originalname: body.originalname };
       dispatch({ type: types.REUPLOAD_COMPLETE, doc: docId, file: _file, __reducerKey });
-      api.get(new RequestParams({ docSharedId }))
+      api.get(new RequestParams({ sharedId: docSharedId }))
       .then(([doc]) => {
         dispatch({ type: libraryTypes.UPDATE_DOCUMENT, doc, __reducerKey });
         dispatch({ type: libraryTypes.UNSELECT_ALL_DOCUMENTS, __reducerKey });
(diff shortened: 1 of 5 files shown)


RafaPolit:
@konzz I have added a proof of concept of the document reload issue.

## Problem

- The ViewDocument constructor is setting the RouteHandler.renderedFromServer to false, to trigger the route handler to execute the getClientState
- This reaches without headers, and therefore, fails for any document that requires being logged in.  That is: all uploads, but also every document on a private instance

I think this is an anti pattern that should be avoided, but did try to 'fix' the issue using the same approach without good results.

## What this was trying to accomplish

This was meant to trigger the saving of the pdfInfo, which lists each document's page characters.  This 'cooking' of data can only be done in client, server can't do this right now.

## Proof of concept

I have tried to fix this via the Viewer componentDidMount with the following criteria:
- the ViewDocument requestState already works on client: there is a series of "if"s that would make client to create this if you happen to navigate to the document from a different page, so that is good... no need for two-times-calling the fetch of data.
- componentDidMount executes only the first time the component is rendered, it is not executed in server
- I have created an if that allows for this extra procedure not to happen on client if the document is empty (you arrived from a different page and the request state has not loaded doc data), and it is also not executed if the server-side-served doc already has pdfInfo (it has already been processed)
- I only executes when: the doc is server-side-served and it doesn't have a pdfInfo property

In that case, it refetches everything that the ViewDocument requires and puts it back on the store.

This may be overshooting, maybe just the doc is needed, but I was afraid that text highlighting and other similar things could go wrong.

Smells:
- I would have liked this to be on the ViewDocument, and not the Viewer.  But the ViewDocument just puts what it requests in the store, renders once, and is not concerned about any future loaded docs.  Perhaps fetching the same values directly from the store instead of the props could accomplish the same thing and could be much better solved?

What remains to be tested:
- That target documents on text references work as expected
- That non-pre-rendered target documents load correctly?  not sure this ever worked, but it's important to try
- Anything else you can think of.  This is a big change... e2e's are passing

If you agree with this approach, we should add tests for this on the Viewer -> componentDidMount

PR #2516, opened 2019-09-09
2515 superagent language

### LINKED ISSUE #2515: superagent calls are not setting the Content-Language header
This is only affecting reuploads for the time being.  If you reupload a file, it's grabbing the cookie value.  In a multi-tab scenario, this was leading to the PDF being uploaded to the wrong language.

The other superagent calls either take care of languages through other procedures, or don't care about the language at all.  Examples:
- attachments: it sends the entity ID (not shared ID) which already limits it to a particular language (this could be an issue if we ever combine languages into a single entity)
- upload: the main upload sends to all languages, so it makes no difference which language you are in
- thesauri and entity uploads via CSV: again, language defined by CSV structure, not language of current 

Still, perhaps for consistency, all calls should include the Content-Language, just "in case"? @daneryl  ?

At any rate, I'm only fixing the reupload for the time being.

RafaPolit:
fixes #2515

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Metadata/actions/actions.js
+++ b/app/react/Metadata/actions/actions.js
@@ -84,11 +84,12 @@ export function loadTemplate(form, template) {
 }
 
 export function reuploadDocument(docId, file, docSharedId, __reducerKey) {
-  return (dispatch) => {
+  return (dispatch, getState) => {
     dispatch({ type: types.START_REUPLOAD_DOCUMENT, doc: docId });
     superagent.post(`${APIURL}reupload`)
     .set('Accept', 'application/json')
     .set('X-Requested-With', 'XMLHttpRequest')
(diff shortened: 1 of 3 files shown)


habbes:
The issue is fixed. The upload now goes to the correct entity language.

However, during the upload I see the following error message appear, although it does not seem to affect the upload itself:

Screenshot 2019-09-10 at 10 42 42
Screenshot 2019-09-10 at 10 43 04

habbes:
I just realised that those errors are fixed by a previous PR, so this one just needs to be updated.

PR #2549, opened 2019-09-30
Allow entity viewer side panel to be closed

### LINKED ISSUE #2462: Relationships filters side panel cant be closed
When viewing an entity there is no way to close/reopen the relationships filter side panel, this is a huge problem in mobile because it just covers all the entity info.

This problem only happens in the entity viewer, but not in the document viewer.

habbes:
fixes #2462 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/App/scss/layout/_content.scss
+++ b/app/react/App/scss/layout/_content.scss
@@ -69,9 +69,6 @@
 }
 .content-header-entity {
   z-index: 4;
-  // @media(min-width: 1024px) {
-  //   z-index: 7;
-  // }
 }
 
 @media(min-width: 1024px) {
--- a/app/react/App/scss/layout/_content.scss
+++ b/app/react/App/scss/layout/_content.scss
@@ -68,10 +68,10 @@
   z-index: 4;
 }
 .content-header-entity {
-  z-index: 5;
-  @media(min-width: 1024px) {
-    z-index: 7;
-  }
+  z-index: 4;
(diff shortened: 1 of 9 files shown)


habbes:
I initially planned to get reuse the `ViewMetadataPanel` that's used in the document `Viewer` component, however that did approach did not prove straightforward and I abandoned it. So I just modifed the `EntityViewer` as necessary to get the feature to work, but you're likely to see a some bits of almost duplicate code.

habbes:
Here's the reason why I abandoned the approach of re-using `ViewMetadataPanel`:

The difficulties were that the EntitiViewer use the Sidepanel component directly and neither used ViewMetadataPanel nor DocumentSidePanel. The layout of the page was slightly different from that of Viewer. The state keys were different. The buttons were located at different places (in the document Viewer the buttons are in the sidepanel,   in the entity Viewer the buttons are below the content. The metadata form (when you click on edit) appears in the main area on the entity viewer, but appears in the sidepanel on the doc viewer. Also, the tabs displayed on the two pages are also different. So with all these differences I arrived at a conclusion that the two were just meant to work completely differently. But I should have asked this before concluding. In the end, I reasoned that since the task at hand was just to make the sidepanel closeable, it would not be a good idea to change how the entity viewer works fundamentally without the task explicitly specifying to do so.

PR #2554, opened 2019-10-03
2533 error when wrong id page

### LINKED ISSUE #2533: calling route /page/wrongid throws an error.
If you have an old URL for a page ID no longer found in the database, the pageViewer is trying to access content of an undefined object.

This should show a 404.

gabriel-piles:
fixes #2533 
PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

Acceptance criteria:
- [ ] A page in the menu with this wrong link "/page/page_not_found" shows a 404

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,4 +22,5 @@ custom_uploads/*
 !uploads/index.html
 !attachments/index.html
 !custom_uploads/index.html
+.idea
 
--- a/app/api/pages/routes.js
+++ b/app/api/pages/routes.js
@@ -43,6 +43,20 @@ export default (app) => {
     }
   );
 
+  app.get(
+    '/api/page',
+
+    validation.validateRequest(Joi.object().keys({
+      sharedId: Joi.string(),
+    }).required(), 'query'),
+
+    (req, res, next) => {
+      pages.getById(req.query.sharedId, req.language)
+      .then(res.json.bind(res))
+      .catch(next);
+    }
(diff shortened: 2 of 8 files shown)


habbes:
Hey @gabriel-piles. I was wondering why you create a new route to fetch pages when the the existing `GET api/pages` can already fetch a single page when the `sharedId` parameter is provided.

gabriel-piles:
> Hey @gabriel-piles. I was wondering why you create a new route to fetch pages when the the existing `GET api/pages` can already fetch a single page when the `sharedId` parameter is provided.

I'm not familiar with the code jet, if you can discuss it with @konzz, It would be nice.

habbes:
@gabriel-piles well I don't think it's a major issue. I just assumed using the existing route would still work. But your approach works out okay too, I assumed maybe you had a specific reason for not re-using the existing route. Maybe @konzz could share his view. Otherwise, I think it's okay to merge.

konzz:
The pages endpoint is used to list the pages too, and when the page is not found (because does not exists) it returns an empty array. We wanted it to return a 404 instead, but if we do that in the /pages endpoint, whenever you go to the pages list in settings for the first time (when its empty) you get a 404

habbes:
@konzz thanks, that makes sense.

PR #2555, opened 2019-10-03
fix typo in "together"

vorburger:
cherry-picking https://github.com/huridocs/uwazi/pull/2553
from master to develop, where I'm assuming it should also be?


konzz:
Hi @vorburger, we usually do a weekly merge to master, I'm closing the PR, but thanks for the PR

PR #2567, opened 2019-10-08
2561 latitude greater than 90

### LINKED ISSUE #2561: Error when inserting a latitude greater than 90
Error when inserting a latitude greater than 90 when editing a geolocation field.

Screen Recording 2019-10-07 at 11 47 42

gabriel-piles:
fixes #2561

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Forms/components/Geolocation.js
```diff
@@ -39,11 +44,30 @@ export default class Geolocation extends Component {
...
   latChange(e) {
+    if (!e.target.value) {
+      this.setState({ currentLatitude: e.target.value });
+      return;
+    }
+
+    let latitude = parseFloat(e.target.value);
+
+    // The latitude is a value between -90 degrees and 90 degrees
+    // The page breaks if the map is refreshed with a latitude out of range
```
RafaPolit: We try to avoid comments (unless really indispensable) on the code.  These are usually better described in the test that proves this concept.

Imagine a condition where this changes, and someone changes the value of the range to -100 to 100, but then forgets to update the comment?  Then they would get disconnected (or if text is inserted between the comment and the code).

This can also happen in the tests, but it provides less clutter.

app/react/Forms/components/Geolocation.js
```diff
@@ -61,32 +85,43 @@ export default class Geolocation extends Component {
   render() {
     const markers = [];
     const { value: [value] } = this.props;
+    const latitude = value ? value.lat : '';
+    const longitude = value ? value.lon : '';
```
RafaPolit: This shouldn't be needed.  There is a defaultProps defined on the component for this exact scenario.  What is triggering this to not have the correct values?

app/react/Forms/components/specs/Geolocation.spec.js
```diff
@@ -35,19 +35,51 @@ describe('Geolocation', () => {
...
     it('should call onChange with the new value', () => {
-      const lonInput = component.find('input').at(1);
       lonInput.simulate('change', { target: { value: '28' } });
       expect(props.onChange).toHaveBeenCalledWith([{ lat: 32.18, lon: 28, label: 'home' }, props.value[1]]);
     });
+
+    it('should not call onChange when empty value', () => {
+      lonInput.simulate('change', { target: { value: '' } });
+      expect(props.onChange).not.toHaveBeenCalled();
+    });
```
RafaPolit: And how will the user just 'remove' the values?  These scenarios where a user inputs a value and then are never able to delete it, should not happen.  How does a user just 'remove' the geolocation altogether?

RafaPolit:
Also, is this not also true for longitude beyond 180 degrees?  If so, perhaps it´s worth doing both approaches at once?

gabriel-piles:
The clear coordinates option is not working properly 

output

gabriel-piles:
> Also, is this not also true for longitude beyond 180 degrees? If so, perhaps it´s worth doing both approaches at once?

The longitude does not throw an error with values out of range

app/react/Forms/components/Geolocation.js
gabriel-piles: Comments removed. Thank you!

app/react/Forms/components/Geolocation.js
gabriel-piles: Now the code is a little bit different, now is used always the getInputValues method to get the props

app/react/Forms/components/specs/Geolocation.spec.js
gabriel-piles: The user can empty the input box, so some code is in place to keep it working properly.

app/react/Forms/components/Geolocation.js
```diff
@@ -39,13 +45,21 @@ export default class Geolocation extends Component {
   }
 
   latChange(e) {
-    const { lon, label } = this.getInputValues();
-    this.onChange({ lat: parseFloat(e.target.value), lon, label });
+    let latitude = e.target.value ? parseFloat(e.target.value) : '';
+    latitude = latitude && latitude < -90 ? -90 : latitude;
```
RafaPolit: Apparently this is the wrong limit.  While the system will accept +90, it won't accept -90.

Negative latitude of -90 still breaks the rendering!  This is more complex than the positive value limiting, as anything above 90 should be 90, but we cannot use integers with the negative value, as limiting to -89 will leave a full degree of error (albeit ti would always be in the Antarctica!).

So, lets use 5 decimal points (about 1 meter worth of distance) and go with -89.99999 for anything above that value.

gabriel-piles: Thank you Rafa, I change it now.

PR #2569, opened 2019-10-09
Parse and beautify more routes in activity log

### LINKED ISSUE #2528: Extend the Audit Log parser / beautifier
Audit Log entries mostly show RAW data in the current state of the audit log.

The log is recording all the necessary data to provide useful info to the users, but these routes / action combinations need to be "beautified".

There is already a working example with the entity / document routes, but this needs to be expanded for a lot of usual routes like relationships, pages, and many others.  

This is designed to compute at run-time, so there is no need of data migrations.  Just keep on expanding the parser with new routes until most everything has a semantic presence in the log, instead of the RAW data.

habbes:
fixes #2528 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/activitylog/activitylogMiddleware.js
+++ b/app/api/activitylog/activitylogMiddleware.js
@@ -2,7 +2,11 @@ import activitylog from './activitylog';
 
 const ignorelist = [
   'POST/api/users',
-  'POST/api/login'
+  'POST/api/login',
+  'POST/api/contact',
+  'POST/api/unlockaccount',
+  'POST/api/resetpassword',
+  'POST/api/recoverpassword'
 ];
 
 export default (req, res, next) => {
--- a/app/api/activitylog/activitylogParser.js
+++ b/app/api/activitylog/activitylogParser.js
@@ -1,5 +1,6 @@
 import templates from 'api/templates';
 import entities from 'api/entities';
+import semanticSearchModel from 'api/semanticsearch/model';
 import { allLanguages } from 'shared/languagesList';
 
 const methods = {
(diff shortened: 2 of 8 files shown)


habbes:
Still a work in progress.

habbes:
I believed I have "beautified" all the relevant routes for the activity log.
Here are some things to note when reviewing this:
- I excluded the following routes from the activity log because they contain sensitive data: `/api/unlockaccount`, `/api/recoverpassword`, `/api/resetpassword`
- I also excluded `/api/contact` from the activity log. But maybe it was intentional to include it in the activity log? What do you think? Should I revert this change?
- I **did not** exclude `/api/users/new`, I think it's useful to know when new users were added. However this route might include the initial password that the admin set for the new user. What do you think?
- The activity log does not save the body of upload requests, this means that it's not possible which document was getting uploaded to, or which attachment was getting uploaded. In such cases I only populated the `description` of the semantic log, but not the `name`.
- I used generic descriptions for bulk operations (i.e. bulk delete, entities multiupdate, bulk relationships, settings) and did not populate the `name` field, even when the operation included only 1 or 0 elements
- I "beautified" `/api/references` logs but did not add a lot of details to them since they are not used anymore since `/api/relationships/bulk` was introduced
- removed trailing slashes from semantic search calls from the client, which prevented the urls from being picked up by the beautifier

daneryl:
LGTM, some logs will need improvement in the future like settings and relationships, but for that we will probably need to improve how this endpoints work, for instance if we split the settings  we will be able to get better log info here, but for now i think this is great !, @RafaPolit ??

PR #2571, opened 2019-10-10
Allow Typescript in Uwazi

bdittes:
This PR configures babel, webpack and jest to handle .ts files in Uwazi.

Changes are quite light, except:
 - all runners (dev-server, production build etc.) need to run through babel (they probably should have done that even before this change)
 --> 'build-production' now creates the prod/ dir with the full compiled and webpack'd app
 --> dev-server and ad-hoc runs (like migrate) use babel-node

The first upgraded js file is app/shared/languagesList.ts (in order to test client and server integration).

Tested: yarn test, yarn hot, yarn e2e

PR checklist:
- [ ] Update READ.me ? no need
- [ ] Update API documentation ? no need

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


daneryl:
@bdittes this is great thanks!

>  Changes are quite light, except:
> 
> all runners (dev-server, production build etc.) need to run through babel (they probably should have done that even before this change)
> --> 'build-production' now creates the prod/ dir with the full compiled and webpack'd app
> --> dev-server and ad-hoc runs (like migrate) use babel-node

all runners were running through babel using babel-register instead of babel-node through the `run.js` script, probably for the change for ts this will work `require("@babel/register")({ extensions: ['.js', '.jsx', '.ts'] });`
dont know what is good practice here, babel-node or register, if we go to babel-babe node i guess we need to remove the run.js script as it is redundant, if we go to register we can remove babel-node and just configure it on the require.

i think we should add on this pr support for [eslint on typescript](https://www.npmjs.com/package/@typescript-eslint/parser) ? maybe think about the rules or ruleset for ts later, what you think ?

bdittes:
Sounds good, I'll make up my mind about node VS register and add basic eslint rules.

PR #2573, opened 2019-10-11
support vscode debugging of jest and hot

bdittes:
Configures the repository to allow vscode import resolution, debug jest tests and attach to an instance started with 'yarn hot'.

PR checklist:
- [ ] Update READ.me ? no need
- [ ] Update API documentation ? no need

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


daneryl:
@bdittes jest debug not working, it requires we have babel-node on launch.json, it is working for me if i remove "runtimeExecutable" and "runtimeArgs", so i guess instead running jest, will this work on ts filess properly as jest uses babel-jest and babel will be configured with ts-preset ? or we just add babel-node as dependency ?

bdittes:
Removed babel-node. I'll ckeck the debug in the typescript PR after this is merged, but yarn test worked out if the box there, so it should be fine.

PR #2575, opened 2019-10-11
Use $DBHOST (if set) instead 127.0.0.1 in database/*.sh (fixes #2564)

### LINKED ISSUE #2564: Database Shell scripts should respect $DBHOST environment variable
The `*.sh` scripts under [`database/`](https://github.com/huridocs/uwazi/tree/development/database/) don't seem to respect the `DBHOST` environment variable, and instead [hard-code `127.0.0.1` as the default](https://github.com/huridocs/uwazi/search?q=127.0.0.1&unscoped_q=127.0.0.1)?

This `DBHOST` does seem to be used at run-time though, according to [app/api/config/database.js](https://github.com/huridocs/uwazi/blob/development/app/api/config/database.js); and my PR #2563 proposes to officially document it. Would it therefore make sense to have the DB scripts support it as well out of the box?

This is a similar problem to  the older #1592 from @fititnt

This came up in the context of https://github.com/huridocs/uwazi/pull/2559.

@konzz @txau

vorburger:
fixes #2564

@fititnt @konzz @txau review?

### DIFF
--- a/database/admin_user.sh
+++ b/database/admin_user.sh
@@ -3,6 +3,9 @@
 parent_path=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 cd "$parent_path"
 
-echo -e "\n\nDeleting ${1:-uwazi_development} database"
-mongo -host ${2:-127.0.0.1} ${1:-uwazi_development} --eval "db.users.drop()"
-mongorestore -h ${2:-127.0.0.1} blank_state/admin_user/ --db=${1:-uwazi_development}
+DB=${1:-uwazi_development}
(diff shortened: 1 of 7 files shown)


fititnt:
(Semi related comment https://github.com/huridocs/uwazi/pull/2559#issuecomment-541265594)

Did not stopped to do the full test (need to prepare again my full environment), but looking at the files, these changes should still work and not break old behavior.

Also, a different PR only for this change seems better than do on the #2559, since if make it more flexible on other places.

Let's wait for @konzz @txau etc just try this one just to be sure, in special if it breaks the CI in a deterministic way (aka regression compared to what was before)

PR #2588, opened 2019-10-15
Handle mongoose validation errors

### LINKED ISSUE #2531: Avoid logging as unhandled Error, Entities title required validation.

habbes:
fixes #2531 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/csv/csvLoader.js
+++ b/app/api/csv/csvLoader.js
@@ -45,7 +45,8 @@ export default class CSVLoader extends EventEmitter {
     .read();
 
     if (Object.keys(this._errors).length === 1) {
-      throw this._errors[Object.keys(this._errors)[0]];
+      const firstKey = Object.keys(this._errors)[0];
+      throw this._errors[firstKey];
     }
 
     if (Object.keys(this._errors).length) {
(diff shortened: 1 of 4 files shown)


daneryl:
@habbes when an error occurs, the error handler is throwing an error also and terminating node process, this is also happening on dev, not something added here but it is a good place to solve it if you have time take a look on that pls.

habbes:
@daneryl could you provide an example of an error that takes down the node and how to reproduce it?

daneryl:
@habbes you can replicate with a simple csv like this, an entity with blank title

```
Title,
title 1,
,
title 2,

habbes:
@daneryl  I found that the error was caused by a bug in the `CSVLoader` which cause the module to throw `undefined` as the error. So the error handler ended up being called with an undefined value, and trying to access its properties eventually caused errors inside the error handled (e.g. trying to access `error.json` when `error` is `undefined`).

I have already fixed the bug in the `CSVLoader` that cause the error.

I could prevent similar errors from blowing up in the error handler by setting `error` to some default `Error` object when it comes in as undefined, but I wonder whether this could lead into critical bugs remaining hidden. What do you think?

habbes:
@daneryl also, is there any other place where you found such an error that causes the process to terminate?

habbes:
@daneryl when an undefined error is caught, the error handler generates a new Error with message `undefined error occurred`, since this error will receive code 500, it will be logged and the stack trace will allow us to investigate the source of the bug.

PR #2593, opened 2019-10-17
Display error when saving data while server is down

### LINKED ISSUE #2503: Saving a page while the server was down didn't provide proper user feedback
While working on a page, the server side went down and I didn't receive a proper error message when trying to save a page. This could be affecting other areas of Uwazi as well.

This bug is particularly evil since users get the impression of their changes being saved, when actually they didn't.

habbes:
fixes #2503

Displays a "Could not reach server. Please try again later." When a request fails when due failed to connection to the server (the server could be unreachable, or the client could be offline).

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/utils/api.js
+++ b/app/react/utils/api.js
@@ -16,7 +16,7 @@ const doneLoading = (data) => {
   return data;
 };
 
-const isOtherApiError = error => error.status && ![500, 404, 401].includes(error.status);
+const isNonUsualApiError = error => error.status && ![500, 404, 401].includes(error.status);
 
 const handleErrorStatus = (error) => {
   if (error.status === 401) {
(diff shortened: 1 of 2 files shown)


habbes:
Not ready for merging. The solution does not work on Firefox.

habbes:
This is now ready for review.

`fetch` throws a `TypeError` when it cannot reach the server or validate the CORS. This error does not have a `status` or any other HTTP-related properties, that's why it was not being caught by existing error handler. The existing error handler also did not display an fall-back error message when the error did not match any of the conditions it was checking.

So this PR now displays a "Server could not be reached" error when a `TypeError` is caught, and it displays a generic "An error has occurred" message when an unknown error is caught.

PR #2606, opened 2019-10-24
Allow id when new entity

gabriel-piles:
To check if a document was already uploaded by the crawler, the entity id is used.


gabriel-piles:
Too much differences with the development branch

Issue #2594, opened 2019-10-18
validate geolocation lat/lon range values on the server 


daneryl:
Latitudes should range from -90 to 90.
Longitudes should range from -180 to 180.

daneryl:
related #2596

daneryl:
this is still not properly validated, the new validation only validates that lat and lon are numbers.

daneryl:
this is actually covered already with current validation.

PR #2622, opened 2019-10-31
Templates and Entity metadata validation

### LINKED ISSUE #2596: Documentation and validation of metadata properties.
in order to use our api there is no way to know the structure needed to be used, for static properties we have the api documentation but for dynamic ones we have nothing, i think we should:

- document and validate template properties configuration
- document and validate entity metadata properties according to the template property configuration.

i think it is more important that we validate this data because even with documentation the data can be sent wrongly formatted, being saved to the db without any error and breaking the application unless detected and changed. this validation could also be used as documentation for people using the api and getting feedback errors on how their data should be structured.

habbes:
fixes #2596 #2594

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


bdittes:
Hi! Reading through this change I'm worried that we're introducing a lot of validators: Joi, PropTypes, TypeScript and now Ajv.

Could you look into https://github.com/YousefED/typescript-json-schema#readme ? I think this would allow us to use Typescript interfaces as the main definition and then generate json-schema (for Ajv) out of that. Then we can replace Joi by Ajv and PropTypes by typescript and we'd have all interfaces defined only in Typescript?

habbes:
This is now ready for review.

I've created a helper function called `wrapValidation` in utils. It allows you to create a separate object for validation purposes and then wraps it on top of your main object like a middleware.

E.g., assume we have an `entitiesValidator` object which implements validation for `entities.js`. For each method of `entities.js` that you want to validate, you need to create a validation method with the same name in `entitiesValidator`, the method should throw an error if validation fails and should not throw if validation succeeds. The method can also be asynchronous, in which case it should return a promise that rejects with an error if validation fails. If validation succeeded, the wrapper will call the underlying `entities.js` method.

```js
const entities = {
   save(data) {
      return db.save(data);
   },
  get(id) {
    return db.get(id);
  }
};

const entitiesValidator = {
   save(data) {
       if (!isValid(data.property)) {
           throw ValidationError;
       }
  }
};

const wrappedEntities = wrapValidation(entitiesValidator, entities);

// use the method as usual, and validation will be applied
wrappedEntities.save(data);
// this method will be called without validation
wrappedEntities.get(data);
```

habbes:
I also made use of the Ajv library to create JSON schema-based validations for entities and templates, similar to what already existed for thesauri. I focused the validation mostly on template properties and entity metadata. However I have relaxed the validation to some extent to prevent crashing existing servers in case they have some data/properties that are not documented or that I did not know about. For example, if incoming data includes fields that are not specified in the spec, this will not cause an error.

app/api/entities/entitiesValidator.js
```diff
@@ -0,0 +1,9 @@
+import { validateEntity } from './entitySchema';
+
+const validator = {
+  save(entity) {
```
bdittes: I don't fully understand the purpose of this construct. If we need to provide a per-function validation helper here, why not just call validateEntity() in the entities.js functions?

app/api/entities/entities.js
```diff
@@ -1,18 +1,34 @@
...
 import templates from 'api/templates/templates';
 import path from 'path';
 import PDF from 'api/upload/PDF';
 import paths from 'api/config/paths';
+import { wrapValidation } from 'api/utils/wrapValidation';
 
-import { deleteFiles } from '../utils/files.js';
+import { deleteFiles } from '../utils/files';
 import model from './entitiesModel';
+import validator from './entitiesValidator';
```
bdittes: does it happen in the rest of Uwazi that model.save() is called directly? I.e. should validation also happen as part of the model?

app/api/entities/entitySchema.js
```diff
@@ -0,0 +1,304 @@
...
+});
+
+const schema = {
+  $schema: 'http://json-schema.org/schema#',
+  $async: true,
+  type: 'object',
+  required: ['title'],
+  metadataMatchesTemplateProperties: true,
+  definitions: {
+    dateRange: {
```
bdittes: I would prefer to pull out definitions into separate schema objects so they can be used in other schemas as well

const dateRangeSchema = { type: ...}
...
const schema = { ...
	definitions: { dateRangeSchema }
...

app/api/templates/templateSchema.js
```diff
@@ -0,0 +1,171 @@
...
+    }
+    return true;
+  }
+});
+
+const schema = {
+  $schema: 'http://json-schema.org/schema#',
+  $async: true,
+  definitions: {
+    property: {
```
bdittes: also extract as separate const propertySchema, please

app/shared/templateTypes.js
```diff
@@ -1,13 +1,20 @@
...
+  multiselect: 'multiselect',
+  nested: 'nested',
+  numeric: 'numeric',
+  preview: 'preview',
   relationship: 'relationship',
+  select: 'select',
+  text: 'text',
 };
 
 export { templateTypes };
```
bdittes: nit: why not default?

app/api/entities/entitySchema.js
```diff
@@ -0,0 +1,304 @@
...
+          {
+            type: 'array',
+            items: {
+              $ref: '#/definitions/dateRange'
+            }
+          },
+          {
+            type: 'object',
+            required: ['label', 'url'],
+            properties: {
```
bdittes: this could be an extracted definitions/linkFieldSchema

app/api/entities/entitySchema.js
```diff
@@ -0,0 +1,304 @@
...
+              label: { type: 'string', minLength: 1 },
+              url: { type: 'string', minLength: 1 }
+            }
+          },
+          {
+            type: 'array',
+            items: {
+              type: 'object',
+              required: ['lon', 'lat'],
+              properties: {
```
bdittes: this could be an extracted definitions/geolocationSchema

app/api/entities/entitiesValidator.js
habbes: That's a good question. The rationale was to allow the developer to write and test the functionality of `entities.save()` without having to worry about validation, and to have that taken care of separately. This was an experiment and I can't say that it's any better than the alternative. I also tried this to allow me to add validation without the burden of updating every single unit test case, since many tests only use a subset of the data that is relevant to the behaviour being tested and this subset will often fail validation. But I will have to update tests either way, so this may not have been as thoughtful an idea.

app/shared/templateTypes.js
habbes: What do you mean by this @bdittes ?

app/api/entities/entities.js
habbes: Since the `entities.save()` is the central means by with entities are saved across Uwazi, I think it should handle validation to ensure that entities that are stored in the db are always valid regardless of whether they were created via the API, through a file import or through some other process in Uwazi.

app/shared/templateTypes.js
bdittes: the file is called templateTypes and only exports one thing, so why not 'export default templateTypes'?

app/api/entities/entitySchema.js
habbes: A separate const like other suggestions? Or a definition embedded in the main schema?

app/shared/templateTypes.js
habbes: Ooh I see now. Yeah, I could make it a default export.

app/api/entities/entitiesValidator.js
habbes: Just to add a clarification on the tests, I still need to update the tests of external modules which call `entities.save()` since external modules work with the "wrapped" version of `entities`, which performs validation.

app/shared/templateTypes.js
daneryl: the idea with this was actually stop using export defaults to have naming consistent across files, we had bad experiences with this in the past.

app/api/entities/entitiesValidator.js
habbes: @joan @konzz @RafaPolit what is your view on this approach?

app/shared/templateTypes.js
bdittes: Aha! That's fine too. Sam mentioned you might be collecting some of these in a style guide? ;)

app/api/entities/entitySchema.js
bdittes: I would prefer to separate out all definitions generally.

app/api/entities/entitiesValidator.js
habbes: I fixed the unit tests, most of them were due to mocking not working as expected with the wrapped object, I updated the implementation of `wrapValidation` to fix it. Maybe this a red flag, too?

bdittes: The extra bind is not a red flag in itself, but I do think the wrapValidation generally adds complexity and is not adding enough benefit to make up for that. So I would vote to remove it...

habbes:
I have removed the `wrapValidation` and separate objects for handling validation following @bdittes request. Now I have gone back to calling the validation method inside the method being validated as usual.

Unless there are any other issues or requests, this PR is done.

app/api/entities/entitiesValidator.js
habbes: I've removed `wrapValidation`.

Issue #2613, opened 2019-10-28
Add delay before retrying semantic search calls that have errors


habbes:
When Uwazi makes a semantic search call to the ML server and it gets an error, it retries the search. The error could be due to the semantic search server failing or shut down or anything else, but Uwazi will keep retrying the searches indefinitely and almost immediately until its get a successful response. This can overload the servers and lead to crashes. So we need to add a delay before retrying to prevent this.

PR #2629, opened 2019-11-06
Semantic search fixes and improvements

### LINKED ISSUE #2584: Cannot stop running semantic search
Start a couple of semantic searches, then try to stop one of them while it's still running.
Instead of the search getting stopped, you get the following error message instead:

Screenshot 2019-10-14 at 16 16 24

habbes:
fixes #2584 #2613 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


habbes:
This PR adds the following fixes/improvements:

### Fixes stop/resume functionality
(see #2584). I think this feature was inadvertently broken when the client-side API requests were overhauled and `RequestParams` introduced. Also, now when you click the stop/resume button, it does not redirect you to the search's page.

### Add delay before retry on error
(see #2613)
I've added a 10 second delay before semantic search retries searching after an error occurs (e.g. can't connect to semantic search server). To reproduce this, start a semantic search in Uwazi then shutdown the semantic search server, then check the Uwazi logs. Before this PR, you'll seen an error logged incensantly without breaks because Uwazi keeps retrying immediately, with this PR, you'll notice a ~10s delay in between the error logs.
I think 10s is a lot, but if the error is not resolved immediately, then a smaller delay could be too frequent. And I thought adding an exponentially increasing delay would be over kill. Let me know if should reduce it (or increase it).

### Start a pending search when one search is stopped
Uwazi processes a limited amount of semantic searches concurrently (currently only up to 3). If more than 3 searches are pending, the extra ones will wait for one to finish. Now if you stop an ongoing search, a pending one will be picked up and processed. Previously, pending searches needed to wait until an ongoing one finishes or gets an error before they could get picked up.

PR #2631, opened 2019-11-14
Metadata object

### LINKED ISSUE #2111: Unify viewers for Documents and Entities under the same
Unify viewers for Documents and Entities under the same route
The default route should be document/id/title
But we will keep the old route for entities entity/id working so we dont break old links to those entities

### LINKED ISSUE #2630: Metadata with denormalized values
- [x] Upgrade mongodb
- [x] Update elkastic search
    - [x] Elastic search mapping
    - [x] Add suffix to properties
    - [x] Update mapping
    - [x] Update existing pages
    - [x] Imports CSV work with prop suffix
- [x] Query Builder
    - [x] Agregations
    - [x] Queries
    - [x] Remove mongo pre queries
- [x] Formater
    - [x] Update representation
- [x] Forms
    - [x] ~~Update models~~
    - [x] Wrap / Unwrap values
- [x] Handle propagation of label updates for entities
- [x] Handle propagation of icon updates for entities
- [ ] Handle propagation of label updates for translations
- [x] Unify Entity and Documents viwer under the same route
- [x] Rebase Q_Q

### LINKED ISSUE #2658: Can't save entities with custom nested field
A partner is receiving a "validation error" when trying to edit an already existing entity with a geolocation property. When viewing the metadata, the geolocation field is present, when editing it the geolocation field disappears, so probably that field is the culprit.

Image Pasted at 2019-12-11 10-48

konzz:
fixes #2630 
fixes #2111 
fixes #2658
PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


RafaPolit:
- [ ] Previously, a missing PDF would show a blank PDF page, but the metadata was still visible.  Now I get a promise rejection error and I am left with a forever-loading blue dots, no metadata info, and I get 200% CPU usage on (not sure if by the client or by the server, which are running on the same machine) while this is shown.

RafaPolit:
- [ ] If I edit the ICON from an entity, this is not propagated to the relationship fields pointing to this entity, which results in ICONs not showing on cards of relationship fields.

RafaPolit:
- [ ] Inheritance is not working on the cards... It shows the original related entity name, but not the inherited value.  The side panel, on the other hand, does show the correct inherited property!  Are they both not using the same formatter?

daneryl:
@RafaPolit 

> Inheritance is not working on the cards... It shows the original related entity name, but not the inherited value. The side panel, on the other hand, does show the correct inherited property! Are they both not using the same formatter?

this is an issue on development, i have created an issue for it #2661.

RafaPolit:
- [ ] When I assign a relationship field to inherit property A from template X, but then I delete property A from the template X without reassigning the inheritance, I'm presented with a blank screen on library and uploads that are showing those entities (this may be happening in DEV also)

RafaPolit:
I have yet to check this, but in case @daneryl or @konzz check this before we start on Monday:
- Has "unpulished" metadata values have been taken into account?

Scenario:
- I create two entities, entity A has Entity B as metadata for one of its properties
- I publish entity A, but not entity B.

Would data from entity B be returned to the client as part of requesting entity A?

Prior to this refactor, I believe it did return, but it was a benign ID, so no info was able to get inferred.  Now, we should find a way to hide this unpublished data if the user is not logged in.

This maybe already have been developed for, I'll test it on Monday, but in case you guys get to this before I do and already know there is an issue here, maybe you can start brainstorming a solution?  Thanks.

daneryl:
@RafaPolit 
> When I assign a relationship field to inherit property A from template X, but then I delete property A from the template X without reassigning the inheritance, I'm presented with a blank screen on library and uploads that are showing those entities (this may be happening in DEV also)

also happening on DEV, issue created #2668

konzz:
> * [x]  Previously, a missing PDF would show a blank PDF page, but the metadata was still visible.  Now I get a promise rejection error and I am left with a forever-loading blue dots, no metadata info, and I get 200% CPU usage on (not sure if by the client or by the server, which are running on the same machine) while this is shown.

We decided to dont Implement a fix about this in this branch

konzz:
@RafaPolit Yep, the scenario about the "unpublished metadata" is happening. It renders titles of unpublished entities, and a link poionting to them that throws a 404.

RafaPolit:
- [ ] Please ensure that #2661 gets addressed as well

PR #2634, opened 2019-11-19
2511-extensive graphs in page e2e tests

### LINKED ISSUE #2511: Create more extensive Page and Graphs-in-Page e2e Tests

elric-wamugu:
fixes #2511

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


RafaPolit:
@elric-wamugu Please take a look at the codeclimate issues.  Most are repeated code, it is a good exercise to try and abstract those (sometimes it is not possible).  But the arrow function return + assignment is not something that should exist.  Either assign and return afterwards if that is the pattern you need or remove the return if the function is only meant to change the value and not return it.

nightmare/paths/graphs.spec.js
```diff
@@ -0,0 +1,146 @@
+/*eslint max-len: ["error", 500], */
```
daneryl: @elric-wamugu this is no needed, file has less lines than the default configured on eslint.

nightmare/paths/graphs.spec.js
```diff
@@ -0,0 +1,146 @@
...
+        .catch(catchErrors(done));
+      })
+      .then(() => { done(); })
+      .catch(catchErrors(done));
+    });
+
+    it('should display Pie chart graph in edited page', (done) => {
+      nightmare
+      .evaluate(() => document.querySelector('div.panel-body.page-viewer.document-viewer > div.alert.alert-info a').href)
+      .then(link => nightmare.goto(link))
```
daneryl: @elric-wamugu why an evaluate to get href and goto instead of clicking the anchor itself ?

nightmare/paths/graphs.spec.js
```diff
@@ -0,0 +1,146 @@
...
+        return nightmare
+        .wait('#app > div.content > div > div > main div.markdown-viewer')
+        .getInnerHtml('#app > div.content > div > div > main div.markdown-viewer')
+        .then((html) => {
+          expect(html).toContain('class="PieChart "');
+          expect(html).toContain('class="recharts-responsive-container"');
+        });
+      })
+      .then(() => { done(); })
+      .catch(catchErrors(done));
```
daneryl: @elric-wamugu i think what we want to test here is that graphs + dataset actually works, we should test somehow that the graph its being rendered properly based on the data, is there any option to do this inspecting the html, if not then maybe we can try making a screenshot and using jest snapshot to compare images, what do you think ?

nightmare/paths/graphs.spec.js
```diff
@@ -0,0 +1,146 @@
...
+
+        return nightmare
+        .write(localSelectors.pageContentsInput, graphs.pieChart)
+        .waitToClick(localSelectors.savePageButton)
+        .wait('div.panel-body.page-viewer.document-viewer > div.alert.alert-info:first-of-type')
+        .getInnerText('div.panel-body.page-viewer.document-viewer > div.alert.alert-info:first-of-type')
+        .then((text) => {
+          expect(text).toContain('/page');
+          expect(text).toContain('(view page)');
+        })
```
daneryl: @elric-wamugu i think this is testing implementation details without need, this functionality you are testing here its already covered by clicking this link in the next test, especially if you use something like clickLink action.

nightmare/paths/graphs.spec.js
elric-wamugu: @daneryl could you explain more on exactly what we need to test for/lookout for on this 'test somehow that the graph its being rendered properly based on the data'?

PR #2663, opened 2019-12-13
2604 escape quotes

### LINKED ISSUE #2604: Relationships search field fails until you complete the quotes in quoted text
We recently expanded the document searching for documents to connect in the relationships tab to not only work with exact matches, but to allow normal partial finds.

In order to replicate the exact match functionality, using quotes works as expected (mostly) by typing: "name of document".  The problem is that, as soon as you type the first ("), you get an error of query_shard_exception due to the unterminated quotes.  It keeps giving the error while you type, until you close the quote, and then it yields the expected results.

fnocetti:
fixes #2604 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/search/specs/search.spec.js
+++ b/app/api/search/specs/search.spec.js
@@ -120,21 +120,22 @@ describe('search', () => {
       });
 
       describe('when quotes are present', () => {
-        it('should not fail if they are paired', done => {
-          search
-            .searchSnippets('"text"', ids.batmanFinishes, 'es')
-            .then(() => {
-              done();
-            })
-            .catch(catchErrors(done));
+        it('should not fail if they are paired', async () => {
+          await expect(
+            search.searchSnippets('"text"', ids.batmanFinishes, 'es')
+          ).resolves.toBeDefined();
         });
-        it('should not fail if they are not paired', done => {
-          search
-            .searchSnippets('"text', ids.batmanFinishes, 'es')
-            .then(() => {
-              done();
-            })
-            .catch(catchErrors(done));
+
+        it('should not fail if they are not paired', async () => {
+          await expect(
+            search.searchSnippets('"text', ids.batmanFinishes, 'es')
+          ).resolves.toBeDefined();
+        });
+
+        it('should not fail if there are escaped quotes', async () => {
+          await expect(
+            search.searchSnippets('"this text \\"includes\\" quotes"', ids.batmanFinishes, 'es')
+          ).resolves.toBeDefined();
         });
       });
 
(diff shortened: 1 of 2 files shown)


RafaPolit:
@fnocetti Please also fix the codeclimate issues.  Unnecessary escaped characters should be fairly simple to fix.

The while assignment may be a bit more time consuming, but it's also important to fix it.

Thanks.

RafaPolit:
If the PR is ready for QA, please, remember to tag it with the "Ready for QA" label.

daneryl:
@fnocetti this escapes when the number of chars is odd, is this the discussed solution, it was never clear to me, this definitely fixes the crash but what other effects have ? if this was already talked about and its clear this LFTM and i will merge.

RafaPolit:
I will repeat the point I made for NOT escaping. If I plan to search for a very specific name, for instance, my target search is:
`”Paper on documenting Rights Violatons”`

The first thing I’ll write is the Quote.  Then I’ll right a lot of words... finally I will write the closing quote.

On this scenario, if the quote is escaped, I’ll get no hits whatsoever until I finish the entire thing! 

If the starting quote is omitted until y close it, I’ll get a lot of hits and probably the right document long before I end writing the sentence.  I think that is considerably better than having no results until the very end.

I think escaping is the wrong approach.

fnocetti:
@daneryl at the beginning the idea was to fix the query whenever it was known to fail, that's why it escapes when there's an odd amount of quotes. From all the posible strategies at the end we decided to escape every quote when in this scenario. As for the side effects: it'll mess with the exact match searches at some point. 

@RafaPolit I agree with you and personally don't think escaping is the best approach. 

Please, let's hold the merge and have a brief discussion.

app/api/search/specs/search.spec.js
```diff
@@ -119,6 +119,25 @@ describe('search', () => {
         });
       });
 
+      describe('when quotes are present', () => {
+        it('should not fail if they are paired', done => {
+          search
```
daneryl: @fnocetti pls change the specs so they use async/await, also it is always a good idea to expect something. based on jest [docs](https://jestjs.io/docs/en/asynchronous#resolves-rejects) this should work and fail when there is a rejection.

```
await expect(
  search.searchSnippets('"text"', ids.batmanFinishes, 'es')
).resolves.toBeDefined();

fnocetti:
I implemented @daneryl changes and also improved the replacement algorithm to prevent escaping already escaped quotes, which would make the search fail in some scenarios.

PR #2670, opened 2019-12-17
changes to improve client testsuite performance

daneryl:
improves client testsuite by around 30 - 50 seconds on my machine.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -86,7 +86,7 @@
     "react/no-array-index-key": ["warn"],
     "react/jsx-first-prop-new-line": ["warn"],
     "react/no-unused-state": ["warn"],
-    "react/jsx-wrap-multilines": ["warn"],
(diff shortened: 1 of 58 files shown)


RafaPolit:
There are a few `<i className="fa" ...>` cases that are not using the Icon component.

Maybe it's worth replacing them as part of moving the `loadIcons` function.  Most are found in the FilterSuggestions component.

daneryl:
@RafaPolit 

> There are a few <i className="fa" ...> cases that are not using the Icon component.
Maybe it's worth replacing them as part of moving the loadIcons function. Most are found in the FilterSuggestions component.

done and good catch, those icons were not working not even on development branch, some others i removed because they are not being used.

RafaPolit:
LGTM.  Merging changes, and if all tests pass, I'll merge this today.  Sorry for the delay.

PR #2677, opened 2019-12-19
2668 inheritance deletion bug

### LINKED ISSUE #2668: app breaks when deleting an inherited property from the original template
When I assign a relationship field to inherit property A from template X, but then I delete property A from the template X without reassigning the inheritance, I'm presented with a blank screen on library and uploads that are showing those entities (this may be happening in DEV also)

konzz:
fixes #2668 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/templates/specs/fixtures.js
+++ b/app/api/templates/specs/fixtures.js
@@ -1,4 +1,9 @@
-/* eslint-disable */
+/**
+ * /* eslint-disable
+ *
+ * @format
+ */
+
 import db from 'api/utils/testing_db';
 import { templateTypes } from 'shared/templateTypes';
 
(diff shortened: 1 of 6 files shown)


app/api/templates/specs/templates.spec.js
```diff
@@ -411,4 +413,19 @@ describe('templates', () => {
       }
     });
   });
+
+  describe('canDeleteProperty()', () => {
+    it('shoul return false if the property is been inherited by others', async () => {
```
fnocetti: There's a typo in the word "should"

app/api/templates/specs/fixtures.js
```diff
@@ -50,41 +61,65 @@ export default {
...
-        { id: '2', type: templateTypes.select, name: 'select', label: 'Select' }
-      ]
-    }
+        { id: '2', type: templateTypes.select, name: 'select', label: 'Select' },
+      ],
+    },
+    {
+      _id: templateToBeInherited,
+      name: 'template to be inherited',
+      commonProperties: [{ name: 'title', label: 'Title', type: 'text' }],
```
fnocetti: Shouldn't we use templateTypes.text instead of the literal?

PR #2678, opened 2019-12-19
sync not returned anymore by default on settings

### LINKED ISSUE #2540: Instances with sync configured, fail to save settings from the UI
Some sync instances reporting problems with synced instances.

daneryl:
fixes #2540

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/sync/syncWorker.js
+++ b/app/api/sync/syncWorker.js
@@ -68,7 +68,7 @@ export default {
   },
 
   async start(interval) {
-    const { sync } = await settings.get(true);
+    const { sync } = await settings.get({}, { sync: 1 });
     if (sync && sync.active) {
       const syncs = await syncsModel.find();
       if (syncs.length === 0) {
--- a/app/api/sync/syncWorker.js
+++ b/app/api/sync/syncWorker.js
@@ -68,7 +68,7 @@ export default {
   },
 
   async start(interval) {
-    const { sync } = await settings.get();
+    const { sync } = await settings.get(true);
     if (sync && sync.active) {
       const syncs = await syncsModel.find();
       if (syncs.length === 0) {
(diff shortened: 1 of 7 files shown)


app/api/sync/syncWorker.js
```diff
@@ -67,7 +67,7 @@ export default {
   },
 
   async start(interval) {
-    const { sync } = await settings.get();
+    const { sync } = await settings.get(true);
```
RafaPolit: Should we start a coding practice to pass semantic objects rather than this standalone values?

For example, if this would be: `settings.get({ selectPrivateValues: true})`, I believe this would much more readable.

I also think it helps passing this 'config' arguments in a much better way, if you latter on need a new param, you just add the key, instead of needing to change the number of arguments on the function (but this is kind of secondary to the main legibility point).

What do you guys think? cc @daneryl @konzz @gabriel-piles @fnocetti 
Maybe this is worth discussing in the tech call?

daneryl: @RafaPolit i agree, we use the string notation '+field' but we should use maybe the object notation also supported,
{ field: 1, noReturnField: 0}, as for other scenarios, we can rely on Typescript to know what things are.

daneryl: @RafaPolit ill change it to be more readable.

daneryl: @RafaPolit i did change the settings get method to be more standard, now you actually have to ask for properties that are marked as select: false, explicitly.

fnocetti:
@daneryl I'm still experiencing the issue as described in #2540.
Steps to reproduce:

1. Configure data sync in the instance.
2. Log in as an administrator
3. Try to edit any setting (i.e. The collection name)

Result: "ValidationError: "sync" is not allowed"

daneryl:
@fnocetti im not able to replicate, can you guide me on how to get the validation error ?

PR #2685, opened 2020-01-03
2660 disable add relationship button 2

### LINKED ISSUE #2660: When there are no Rleationship Types, I can still add a Relationship field to template
When defining a template, if I don't have any Relation Types, the Relationship property is greyed out, and I cannot drag it into the template.

But... I can still click the "+" sign and it will get assigned to the template, potentially causing a lot of errors.

This should, ideally, be also tested in an E2E.

fnocetti:
fixes #2660

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Templates/components/PropertyOption.js
+++ b/app/react/Templates/components/PropertyOption.js
@@ -18,7 +18,9 @@ export class PropertyOption extends Component {
   }
 
   addProperty() {
-    this.props.addProperty({ label: this.props.label, type: this.props.type });
+    if (!this.props.disabled) {
+      this.props.addProperty({ label: this.props.label, type: this.props.type });
+    }
   }
 
   render() {
--- a/app/react/Templates/components/PropertyOption.js
+++ b/app/react/Templates/components/PropertyOption.js
@@ -18,19 +18,19 @@ export class PropertyOption extends Component {
   }
 
   addProperty() {
-    if (!this.props.disabled) {
-      this.props.addProperty({ label: this.props.label, type: this.props.type });
+    const { disabled, label, type, addProperty: addPropertyAction } = this.props;
+    if (!disabled) {
+      addPropertyAction({ label, type });
     }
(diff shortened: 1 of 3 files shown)


app/react/Templates/components/PropertyOption.js
```diff
@@ -15,7 +15,9 @@ export class PropertyOption extends Component {
   }
 
   addProperty() {
-    this.props.addProperty({ label: this.props.label, type: this.props.type });
+    if (!this.props.disabled) {
```
RafaPolit: Please ensure that the ESLINT plugin is working for your IDE, this should not be committed with this warnings.

app/react/Templates/components/specs/PropertyOption.spec.js
```diff
@@ -32,6 +32,24 @@ describe('PropertyOption', () => {
...
+      expect(props.addProperty).not.toHaveBeenCalled();
+    });
+
+    it('should add a property when clicked', () => {
+      const props = { label: 'test', type: 'optionType', addProperty: jasmine.createSpy().and.returnValue({}) };
+      TestComponent = wrapInTestContext(dragSourceOption);
+      component = renderComponent(TestComponent, props);
+      const button = TestUtils.findRenderedDOMComponentWithTag(component, 'button');
+      TestUtils.Simulate.click(button);
+      expect(props.addProperty).toHaveBeenCalled();
```
RafaPolit: While line 42: not.toHaveBeenCalled may be enough, just 'toHaveBeenCalled' when there are specific parameters passed to the function call is not comprehensive enough.  This test would be green if I removed the argument for the call.  

Please ensure that this is a toHaveBeenCalledWith that tests the relevant arguments are passed to the function.

PR #2689, opened 2020-01-08
Fix assets path for swagger on production build

### LINKED ISSUE #2649: Unable to access public API commands
When I go to https://documentos.memoriayciudadania.org/api/ I expect to see a list of available API commands for this Uwazi instance, but instead I see a white screen with the error message:

> Cannot GET /api/

fnocetti:
fixes #2649 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/swagger/swaggerconfig.js
+++ b/app/api/swagger/swaggerconfig.js
@@ -13,6 +13,7 @@ export default (app) => {
       description: `Uwazi is an open-source solution for building and sharing document collections.
       <br>Remember that using the "Try it out" functionality will execute the requests over your local instalation!`
     },
+    host: 'localhost:3000',
     basePath: '/api',
     tags: [
       { name: 'attachments' },
--- a/app/api/swagger/swaggerconfig.js
+++ b/app/api/swagger/swaggerconfig.js
@@ -3,6 +3,8 @@ import swaggerJSDoc from 'swagger-jsdoc';
 
 import path from 'path';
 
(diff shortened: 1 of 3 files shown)


app/api/swagger/swaggerconfig.js
```diff
@@ -11,7 +13,6 @@ export default (app) => {
       description: `Uwazi is an open-source solution for building and sharing document collections.
       <br>Remember that using the "Try it out" functionality will execute the requests over your local instalation!`
```
RafaPolit: This suggests that "trying out the functionality" will execute that in localhost:3000 local environment.

Does removing the host key mean this will now get executed remotely?  Do we really want that change?

app/api/swagger/swaggerconfig.js
```diff
@@ -11,7 +13,6 @@ export default (app) => {
...
     },
-    host: 'localhost:3000',
```
RafaPolit: Not sure this is the wanted behavior.  Check other comment.

fnocetti: The issue reported Swagger to fail when accessed on a production build using a FQDN. We should discuss if we want the "Try it out" functionality to work on a production deployment. Not removing the host implies that any Try it out action on a domain different than localhost won't work.
I'll double check it, this is still in progress.

PR #2693, opened 2020-01-14
post metadata-object nits

bdittes:
Some nits Sam and I discovered while working on top of metadata-object

PR checklist:
- [x] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/entities/metadataValidators.js
```diff
@@ -57,7 +57,8 @@ const validateRequiredProperty = (property, value) => {
 const isValidLinkField = value =>
   isString(value.label) && value.label && isString(value.url) && value.url;
 
-const validateNumericProperty = value => isNumber(value) || value === '';
+const validateNumericProperty = value =>
+  isNumber(value) || value === '' || (isString(value) && `${parseInt(value, 10)}` === value);
```
RafaPolit: I am not really sure why this is needed.  Why would we want to allow strings that hold numbers?  This will force us to sanitize this data either upon save, or everywhere else where you want to do operations with these values.  Can you please elaborate on the reasoning behind this?

Thanks.

app/api/templates/templates.js
```diff
@@ -203,7 +203,7 @@ export default {
       const entitySelects = [];
       return Promise.all(
         selects.map(select =>
-          request.get(`${dbURL}/${select.content}`).then(result => {
+          request.get(`${dbHost}/${select.content}`).then(result => {
```
RafaPolit: DBHOST is a process.env optional value, which is then defaulted to "localhost" on standard installations.  Where is this default happening if this variable is imported directly?

app/api/entities/metadataValidators.js
bdittes: This is the current state in planinternational. Not having this breaks any save action there.

app/api/templates/templates.js
bdittes: This just fixes the fact that dbURL was undefined ;)

I added the default to database.js

app/api/entities/metadataValidators.js
RafaPolit: Yes, because CSV import treated everything as string.

I am concerned to end up with mixed data of strings and, moving forward, numbers.  Perhaps the correct approach is to do a migration that every numeric field is sanitized to a number?

@txau  ?

RafaPolit: We could merge it like this and do the migration further down the road and remove this validation to make it stricter.  That is another possible route.

bdittes: I would prefer to not pile another migration onto metadata-object :)

PR #2704, opened 2020-01-16
Increased elasticsearch max index to 2k

konzz:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/database/elastic_mapping.js
+++ b/database/elastic_mapping.js
@@ -5,6 +5,7 @@ import languages from '../app/shared/languages';
 
 const config = {
   settings: {
+    'index.mapping.total_fields.limit': 2000,
     analysis: {
       char_filter: {
         remove_annotation: {


database/elastic_mapping.js
```diff
@@ -5,6 +5,7 @@ import languages from '../app/shared/languages';
 
 const config = {
   settings: {
+    'index.mapping.total_fields.limit': 2000,
```
RafaPolit: Do we really want a global upping of this limit, or on a 'per collection' case whenever needed?

As I understand it, this could impact performance or, at the very least, resource allocation, right?

konzz: The documentation says the limit is there to prevent mappings from becoming too large and impact performance, but it's not clear if just by increasing the limit it impacts the performance, or increasing the limit and mappings becoming too large then is what impacts the performance.
So maybe the instances that do not use the extra limit won't add performance problems.
It would be great to have a way of measuring this.

PR #2729, opened 2020-01-22
Fixed issue for id comparation

### LINKED ISSUE #2717: Problem creating a relationship to an entity type, and enabled inheriting of geolocation property (it won't let me save this configuration)

### LINKED ISSUE #2727: Adding a field to a template - Error message
Screenshot 2020-01-22 at 11 30 50

- Error happening while adding a field to a template
- It happens whatever the nature of the field added is
- It is happening on a very specific template, while replicating the action on another template, the field addition is done successfully
- Private instance

konzz:
fixes #2717
fixed #2727

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/templates/specs/templateSchema.spec.js
+++ b/app/api/templates/specs/templateSchema.spec.js
@@ -172,7 +172,7 @@ describe('template schema', () => {
       }
     });
 
-    it('should be ok with other scenarios', async () => {
+    it("should work with ID's stored as either strings or mongo ID's", async () => {
       const template = {
         _id: templateToBeInherited,
         name: 'changed name',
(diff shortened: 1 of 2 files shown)


app/api/templates/specs/templateSchema.spec.js
```diff
@@ -167,5 +171,25 @@ describe('template schema', () => {
         expect(e).toBeInstanceOf(Ajv.ValidationError);
       }
     });
+
+    it('should be ok with other scenarios', async () => {
```
RafaPolit: Which scenario exactly? "Other scenarios" makes it really hard to know what it is this test is doing.

konzz: Rafa, I don't agree that the test should describe specific implementations of the code, instead, it should describe functionality.
I agree that "other scenarios" may be a poor description, but the idea is a test that ensures that the validation is not going to throw an error when you are simply updating the template.

PR #2738, opened 2020-01-24
Add Husky to manage git hooks & start off with a few straightforward pre-commit/push checks.

samschaevitz:
To prevent the introduction of merge conflicts or untyped typscript (e.g.), this PR adds:
- a new node module, [Husky](https://github.com/typicode/husky)
- a new no-commit-master.sh script that ensures one isn't currently trying to commit to the master branch directly
- a new no-merge-conflicts.sh that ensures there aren't any outstanding merge conflicts introduced in the commit.
- a husky configuration in package.json that also ensures tests are passing before commits are pushed to origin

PR checklist:
- [X] Update READ.me ?
- [n/a ] Update API documentation ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


samschaevitz:
(this should only have the last 6 commits, not the other 206... /cries in git)

README.md
```diff
@@ -125,6 +125,24 @@ Uwazi supports the following environment variables to customize its deployment
...
+
+Fast checks go here.
+
+- Develop on a branch other than `Master`.
+
+#### Pre-Push
+
+Slower checks that are still useful go here.
+
+- All objects delared in TypeScript (`.ts`) files should have types assigned to them.
```
konzz: declared*

samschaevitz: Fixed, along with all other typos & lint errors in the file (thanks Code Spell Checker!)

samschaevitz:
Great! Makes sense. Definitely don't want to slow down dev workflows unnecessarily.

Note, I have added a filter flag to only ever run tests on files that are relevant to those that have been changed in this commit diff. Did you want me to remove that altogether? Its speed is of course a function of how many files have been touched.

With respect to pushing with failing tests that there's also a possibility (without changing files) to skip the hooks if you know they'll not pass.  https://stackoverflow.com/questions/7230820/skip-git-commit-hooks.

I'll remove the full yarn test and leave the pre-commit relevant-tests-only run for now. Let me know if you'd like that removed, too.

hooks/run-relevant-tests.sh
```diff
@@ -0,0 +1,35 @@
...
+
+    yarn prettier "$FILE"
+
+    if [ "$?" == 0 ]; then
+        echo "\t\033[32mPrettier  Passed: $FILE\033[0m"
+    else
+        echo "\t\033[41mPrettier Failed: $FILE\033[0m"
+        PASS=false
+    fi
+done
```
daneryl: @samschaevitz this is running eslint on the whole application per file changed, prettier stage also prettifies all the app, either we change/remove yarn eslint/prettier scripts or we use the eslint/prettier executable directly, also i think we should use $STAGED_FILES in a single command instead of an iteration ? what do you think.

daneryl:
@samschaevitz i think in favor of speed we should remove the pre-push hook (the idea here is that all this is going to run also on circleci for every push), i think the pre-commit can be useful.

PR #2768, opened 2020-02-13
One up polish

rustedgrail:
A collection of UI improvements to polish the one-up review flow.

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/I18N/components/I18NLink.tsx
```diff
@@ -1,50 +1,73 @@
...
+            onClick(e);
+            I18NLink.navigate(to);
+          },
+          title: confirmTitle,
+          message: confirmMessage,
+        });
+      } else {
+        onClick(e);
+        I18NLink.navigate(to);
+      }
```
daneryl: @bdittes @rustedgrail i think we should not add more functionality to i18NLink, what a bout an encapsulation of the Confirm functionality, i think this is much needed actually not only for this but we have this context.confirm thing all around the app, [this](https://itnext.io/add-confirmation-dialog-to-react-events-f50a40d9a30d) is a good option to encapsulate it on a react component, but maybe for this particular case its not clear how to use it, what do you think ?

bdittes: Hi! David and I both looked at it and concluded that it's a very elaborate solution (which I don't even fully understand since it has to intercept custom event handlers...). Also, the changes to I18NLink are not massive, so I would prefer to merge as-is.

PR #2801, opened 2020-03-05
Adding a notice box on the one up page

rustedgrail:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Review/scss/review.scss
+++ b/app/react/Review/scss/review.scss
@@ -24,6 +24,15 @@ $header-height: 44px;
   padding: 7px 10px;
 }
 
+.get-started {
+  height: 25px;
+  width: 85px;
+  margin-top: 15px;
+  padding: 2px;
+  font: Roboto;
+  border-radius: 2px;
+}
+
 .btn-toggle-on {
   background-color: #ccc;
   &:focus {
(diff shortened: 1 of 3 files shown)


app/react/Review/scss/review.scss
```diff
@@ -24,6 +24,15 @@ $header-height: 44px;
   padding: 7px 10px;
 }
 
+.get-started {
```
rustedgrail: Should we override btn or btn-primary instead of creating a new class?

app/react/Thesauri/Notice.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { Component } from 'react';
+import { Icon } from 'app/UI';
+
+function learnIcon() {
+  return (
+    <svg width="29" height="29" viewBox="0 0 29 29" fill="none" xmlns="http://www.w3.org/2000/svg">
```
rustedgrail: Does this need to be added to font awesome as a custom icon? Font awesome needs a single svg path, so we would have to regenerate this.

app/react/Thesauri/Notice.tsx
```diff
@@ -0,0 +1,59 @@
...
+    super(props);
+    this.state = {
+      isHidden: false,
+    };
+
+    this.close = this.close.bind(this);
+  }
+
+  close() {
+    this.setState({ isHidden: true });
```
rustedgrail: Should this be saved somewhere or is it okay if the notice returns whenever the page is refreshed?

PR #2810, opened 2020-03-10
migration 19 fix

### LINKED ISSUE #2809: migration 19 throws error when runed from yarn hot
when starting uwazi, files collection on the database is created and after that migration 19 tries to rename uploads to files, at this point files already exists and throws an error.

daneryl:
fixes #2809 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/migrations/migrations/19-rename-uploads-to-files/index.js
+++ b/app/api/migrations/migrations/19-rename-uploads-to-files/index.js
@@ -8,14 +8,20 @@ export default {
   async up(db) {
     process.stdout.write(`${this.name}...\r\n`);
 
-    try {
-      await db.collection('uploads').rename('files');
-    } catch (e) {
-      if (e.message !== 'source namespace does not exist') {
-        throw e;
+    const uploadsExist = (await db.listCollections({ name: 'uploads' }).toArray()).length;
+    if (uploadsExist) {
+      try {
+        await db.collection('files').drop();
+      } catch (e) {} //eslint-disable-line
+
+      try {
+        await db.collection('uploads').rename('files');
+      } catch (e) {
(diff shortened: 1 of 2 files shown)


app/api/migrations/migrations/19-rename-uploads-to-files/specs/19-rename-uploads-to-files.spec.js
```diff
@@ -41,4 +41,9 @@ describe('migration rename-uploads-to-files', () => {
     await testingDB.mongodb.collection('uploads').drop();
     await migration.up(testingDB.mongodb);
   });
+
+  it('should not fail when uploads collection does not exist', async () => {
```
RafaPolit: This description does not match the test.

app/api/migrations/migrations/19-rename-uploads-to-files/index.js
```diff
@@ -8,6 +8,10 @@ export default {
   async up(db) {
     process.stdout.write(`${this.name}...\r\n`);
 
+    try {
+      await db.collection('files').drop();
```
RafaPolit: This is not idempotent, which has been an 'unspoken' rule of migrations.  Not sure if it merits here to go the extra effort to ensure that a second run will not forever delete the already-migrated files?

daneryl: @RafaPolit changed it so it only runs the migration when has not ran before, this assumes that if uploads collection does not exist then the migration has already ran.

RafaPolit:
LGTM

PR #2813, opened 2020-03-12
Adding select all to the documentslist and a back button to the multiEditThesaurus

rustedgrail:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Layout/DocumentsList.js
+++ b/app/react/Layout/DocumentsList.js
@@ -20,6 +20,7 @@ class DocumentsList extends Component {
     super(props, context);
     this.state = { loading: false };
     this.clickOnDocument = this.clickOnDocument.bind(this);
+    this.selectAllDocuments = this.selectAllDocuments.bind(this);
   }
 
   componentWillReceiveProps() {
@@ -30,6 +31,13 @@ class DocumentsList extends Component {
     if (this.props.clickOnDocument) {
       this.props.clickOnDocument.apply(this, arguments);
     }
+  }
+
+  selectAllDocuments() {
+    this.props.documents.get('rows').forEach(doc => {
+      this.ctrlKey = true;
+      this.clickOnDocument(this, doc);
+    });
   }
 
   loadMoreDocuments(amount) {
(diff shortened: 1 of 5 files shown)


app/react/Layout/DocumentsList.js
```diff
@@ -119,6 +130,20 @@ class DocumentsList extends Component {
               stateProperty={this.props.sortButtonsStateProperty}
               storeKey={this.props.storeKey}
             />
+            <div className="select-all-documents">
```
rustedgrail: The select all works for the other document list views, but should this button only be added on the multiEditThesaurus page?

bdittes: I think it makes sense for all.

app/react/Layout/DocumentsList.js
```diff
@@ -32,6 +33,13 @@ class DocumentsList extends Component {
     }
   }
 
+  selectAllDocuments() {
+    this.props.documents.get('rows').forEach(doc => {
+      this.ctrlKey = true;
+      this.clickOnDocument(this, doc);
```
bdittes: Why not simply call this.props.selectDocuments once?

app/react/Layout/DocumentsList.js
```diff
@@ -69,6 +77,9 @@ class DocumentsList extends Component {
       connectionsGroups,
       LoadMoreButton,
       rowListZoomLevel,
+      location: {
```
bdittes: please use the existing state.library.sidepanel.multiEditOpts instead of reparsing from the query.

bdittes: Also, I think thesaurusName is not usually populated, so you'll need to look up the id in state.thesauris.find(t => t._id.toString() == opts.thesaurus)

app/react/Thesauri/ThesaurusCockpit.tsx
```diff
@@ -108,6 +108,7 @@ export class ThesaurusCockpitBase extends RouteHandler {
         ? new Date(+modelTime * 1000)
         : null;
     let status;
+    const multiEditUrl = `/library/?multiEditThesaurus=${thesaurus._id}&thesaurusName=${thesaurus.name}`;
```
bdittes: With the changes in DocumentsList, this shouldn't be needed.

PR #2816, opened 2020-03-13
2770 select no translations

### LINKED ISSUE #2770: Properties of type 'select' are not recognising translated list/thesaurus items
The list/thesaurus on this Uwazi instance has been translated into AR:
image

...but when you are in the AR interface, the items are in English:

image

...but the properties of type "multiselect" seems to pulling in the translated list items, as expected.

gabriel-piles:
fixes #2770

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Metadata/components/MetadataFormFields.js
```diff
@@ -160,7 +161,11 @@ export class MetadataFormFields extends Component {
                       model={model}
                       field={`metadata.${property.name}`}
                     />
-                    {t(templateID, property.label)}
+                    {templateID ? (
```
konzz: Is this if really necessary? are there scenarios where the template does not have _id?

gabriel-piles: When multiple entities with different templates and common properties are edited, the template id is void and the translations module throws an exception

PR #2828, opened 2020-03-20
2450 performance improvements

### LINKED ISSUE #2450: Performance issues detected when handling large amount of documents
Issues detected:
First  scenario:
A Document with a relationship field pointing to 500 entities of a different collection with around 76k entries. Takes around 22s~to render the entity.

The problem:
The MetadataFormater, for each of those relationships(500) goes over the 76k documents to transform the "sharedId" to the title. 500 * 76k = 38M

Possible solution:
Iterate over the 76k only once, translating the relationships in that iteration.

Second scenario:
Template a select / relationship / multiselect with 76k values.

The Problem:
Takes around 5s to render the form, even though we only render the top 5 options.
It will be also a problem when the user expands the list to see the other 76k options.

Possible solution:
Change the select to a lookup of the options instead of loading the 76k options.

konzz:
fixes #2450 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


gabriel-piles:
1. There is a 'type' error when entering the library without login. It can be seen in the Uwazi terminal

app/api/auth/authMiddleware.ts:6:13 - error TS2304: Cannot find name 'User'.

2. There is a validation error when editing two or more items at the same time in the quick labeling window.

To reproduce this the e2e fixtures can be used in the following window:
http://localhost:3000/es/library/?quickLabelThesaurus=5e3d1853deeb2652690a0c10&q=(allAggregations:!t,includeUnpublished:!t)

When two or more entities are edited, a validation error appears.

RafaPolit:
There is a bug on the button of 'x more' once you type a search in the input field.  Current behavior apparently also shares this error. 

Current behavior:
- You are seeing a filter showing the first top 5 values
- You see an arrow that reads '48 more' at the bottom
- If you type in something that would limit those results to even 1, the arrow still shows '48 more'

Expected behavior:
- The number of extra records should follow the actual available aggregations for that filter.

RafaPolit:
The whole labeling of the 'x more' and the 'xxxx total possible options' is utterly confusing!

- I have an aggregation that has 7532 possible options
- When I load the library, I see the first 5 and a label that says `196 more`  This is wrong
- When I click on the 'x more', only then am I presented with a `7532 total possible options, try searching`

Why do I see the 196 if there are 7532 options?  Should I not see the 7532 straight from the start?

I am also confused as to why and when it is that this label is shown.

Because, if I only have less than the 200 limit, that label does not show if I click on the 'x more' (which I think is perhaps correct), but then if I search something, this appears if I have the 'x more' clicked, but it doesn't if I have the 'x more' collapsed.  That is not correct in my opinion.

I think we need @simonfossom 's and @txau 's opinion here.

What I suggest is something much simpler than the very complex set of rules this now has as to when to show what.

- the 'x more' should read the 'true' amount of possible options, not the max showing based on the arbitrary limit imposed.
- If there are more than the limit of possible accessible options from the UI (arbitrary number... lets say 50? the current 200 limit?) we show that `try searching`  label, no matter if the 'more' is expanded or collapsed.  

My doubt is, what happens when you have already searched something? This value is listing the total amount of possible options, if I had no selection at all!  But what happens if my current search has more than 200 hits?  I still have no way of knowing how many search options would I get with my current phrase of search.

So I think that that value of `X Total possible options` should match what I have already narrowed down.  And if my narrowing down of the 7500+ records is now 1000, I should see 'still 1000 possible options, narrow down your search', or something like that.  Am I explaining myself correctly?  What I mean is: I believe that value should take into account my current search values, just like everything else on the aggregations.  Because now it is completely detached from the search terms.

In that way, if you have typed a perfect match of the value you want to filter, this will not say '7000+ possible options', that is just simply not true.

So, to append to the conditional shown above, I'd do:
- If the 'x more' is collapsed, we don't show the number on the extra label, as it will already be present in the 'x more' button.
- If it is expanded, we put the value.
- we remove the 'total' and 'possible' term, it's just '7500 options'.
- since it will change depending on the search term, the `possible` is no longer needed: if you search something that will limit those results to 1000, it will say `1000 options`, or `1000 remaining options` 

(we could even subtract the showing ones, so if we are already showing 5, it will say 995 additional options, try searching... and if you expand, it since you are seeing already 200 options, it will say 800 additional options, try searching)

I think this is much less confusing.

What I am 100% sure is that the number there HAS to follow the already filtered terms, or it makes no sense at all.

## Example

Just to exemplify the issue:

I am seeing at the 'Document' aggregation of the 'Paragraph' template:
performance-A01

How many documents do I have on those aggregations?  Seems like 196 + 5, right?  So, I expand those 196 and:
performance-A02

Wait... what? So I didn't have 201, I had 7532 total?  I didn't know that.  Ok, I'll search something very specific:
performance-A03
Ok.  I have a single value, showing, but it still says there is 196 more?  Sure... I'll click on that to see those 196 'extra' options!

performance-A04

What happened? No extra options, but I still have 7532 options?

This is really, really confusing, even for me that I know what is going on behind the scenes.

I strongly think we need to simplify this A LOT.

RafaPolit:
Why is the search box only half-width of the side panel?

simonfossom:
Yeap, completely agree on this with Rafa. 
Simply put, the aggregate of 'the rest of options' doesn't match what a user would expect.

txau:
@RafaPolit the search box always had that size. We can change it if @simonfossom agrees.

Regarding the other issue, I guess a quick fix would be displaying the correct total options at all times? Ie. if we have 8000 options:

- When everything is folded it has 5 options visible and the label says "7995 more"
- When the 50 top options are unfolded, the label says "7950 more. Try searching"

@konzz I though we reviewed that label and we already agreed on removing "total possible options"?

Also, the options should take into account other filters and searches. So if we have 8000 options but we have already filtered or search narrowing it down to 4000:

- When everything is folded it has 5 options visible and the label says "3995 more"
- When the 50 top options are unfolded, the label says "3950 more. Try searching"

txau:
Question, is this only affecting relationship filters? Is it also having any side effect on thesauri based filters?

RafaPolit:
> @RafaPolit the search box always had that size. We can change it if @simonfossom agrees.
> 
> Regarding the other issue, I guess a quick fix would be displaying the correct total options at all times? Ie. if we have 8000 options:
> 
> * When everything is folded it has 5 options visible and the label says "7995 more"
> * When the 50 top options are unfolded, the label says "7950 more. Try searching"
> 
> @konzz I though we reviewed that label and we already agreed on removing "total possible options"?
> 
> Also, the options should take into account other filters and searches. So if we have 8000 options but we have already filtered or search narrowing it down to 4000:
> 
> * When everything is folded it has 5 options visible and the label says "3995 more"
> * When the 50 top options are unfolded, the label says "3950 more. Try searching"

This is exactly what I proposed as well, but more concisely explained. Yes. I agree with this approach.

app/api/search/search.js
```diff
@@ -586,6 +699,105 @@ const instanceSearch = elasticIndex => ({
...
+
+  async autocomplete(searchTerm, language, templates = [], includeUnpublished = false) {
+    const publishedFilter = includeUnpublished ? undefined : { term: { published: true } };
+
+    const body = {
+      _source: {
+        include: ['title', 'template', 'sharedId', 'icon'],
+      },
+      from: 0,
+      size: 200,
```
RafaPolit, line 766: Again, fixed 200 value.

app/api/sync/specs/5e71fcc1a26d0868e434deba.jpg
```diff
@@ -0,0 +1 @@
+undefined
```
RafaPolit, line 1: Please, delete these residual files from mid-run tests.

app/api/sync/specs/5e71fd0194f4fa69deb15cca.jpg
```diff
@@ -0,0 +1 @@
+undefined
```
RafaPolit, line 1: Please, delete these residual files from mid-run tests.

app/api/search/search.js
```diff
@@ -173,65 +167,178 @@ function searchGeolocation(documentsQuery, templates) {
...
-    result._id = hit._id;
-    return result;
+const _sanitizeAgregationNames = aggregations => {
+  return Object.keys(aggregations).reduce((allAgregations, key) => {
+    const sanitizedKey = key.replace('.value', '');
+    return Object.assign(allAgregations, { [sanitizedKey]: aggregations[key] });
+  }, {});
+};
+
+const _getAggregationDictionary = async (aggregation, language, property, dictionaries) => {
```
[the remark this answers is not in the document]
konzz: It is a javascript convention to say the method is private, not a team convention.

app/api/search/documentQueryBuilder.js
```diff
@@ -260,6 +260,12 @@ export default function() {
           dictionaries,
           baseQuery
         );
+
+        baseQuery.aggregations.all.aggregations[`${property.name}__optionsCount`] = {
```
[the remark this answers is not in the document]
konzz: Dont know, need to try it.

konzz:
The number of total options when editing fields shows the number for all languages.

RafaPolit:
There is a small issue with when the "more options" is displayed:
Scenario A (working as expected):
- If there are more than the 5 available options, it correctly shows X more:
perf-issue-A01

- If I narrow down results (filter), and the resulting values are less than 6, then it correctly hides the 'xxx more' label:
perf-issue-A02

Scenario B (NOT working as expected):
- If there are more than the 5 available options, it correctly shows X more:
perf-issue-A01

- If I click on the 'more', it correctly extends the results up to the proposed limit, and correctly lists the 'YYY more. Narrow down results":
perf-issue-A03

HERE IS THE PROBLEM:
- if I then narrow down the results (with the "more" button ALREADY expanded), then the "XXX more" is not appropriately updated and, if my narrowed results are less than 6, then the label is not actually hidden:
perf-issue-A04

Scenario C (NOT working as expected):
- If there are more than the 5 available options, it correctly shows X more
- If, without expanding the 'more' button,  I filter by text and my narrowed results are more than the 5 I was currently seeing, then the 'more' disappears and I get 200 results displayed without having asked for the 'more' button
- Furthermore, if there were more than the  200 results, I no longer see the 'XXX more results' at all at the end of the list.

txau:
Idk how it was decided to change the label from "Try searching" to "Narrow down results" but I rather keep "Try searching". It is explicitly telling the user what to do, the other one is an open ended proposal.

RafaPolit:
> Idk how it was decided to change the label from "Try searching" to "Narrow down results" but I rather keep "Try searching". It is explicitly telling the user what to do, the other one is an open ended proposal.

You were in the call, and you agreed to this.  The question is: what happens if you have already searched something, still have hundreds of results, and you are presented with a link to "try searching"?  You just searched.  So, "Narrow down results" remains true whether you have already searched or not.  "Try searching" suggests you haven't yet tried that, and that may not be the case.

What would you suggest then? A conditional label that reads something before you searched and something different after you searched? I don't like that option.

RafaPolit:
Please make the "more", the "more options" and "Narrow down results" labels translatable. (the "more" should already be translatable as it is ussed in the library buttons?)

RafaPolit:
We have talked with Nancy and the term we are aiming for the button label is: "Narrow your search".
We are taking cue from wordings used in other platforms, like Google's Android: https://support.google.com/cloudsearch/answer/6172299?co=GENIE.Platform%3DAndroid&hl=en

Also, since this will be 'translatable', it could be made into anything the collection admin deems is the right wording for their use case.

app/api/search/metadataAggregations.js
```diff
@@ -1,10 +1,9 @@
-/** @format */
-
+import { preloadOptionsSearch } from 'shared/config';
```
RafaPolit, line 1: Please fix this minor eslint error.

app/api/search/metadataAggregations.js
```diff
@@ -148,23 +140,30 @@ const propertyToAggregation = (property, dictionaries, baseQuery, suggested = fa
         !match.bool.should[1].terms[path])
   );
   filters = filters.concat(baseQuery.query.bool.must);
+  return filters;
+};
+
+const getpath = (property, suggested) => {
```
RafaPolit, line 146: Please fix this minor eslint error.

app/api/auth/authMiddleware.ts
```diff
@@ -0,0 +1,25 @@
+import { Request, Response, NextFunction } from 'express';
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { User } from '../users/usersModel';
```
RafaPolit, line 3: Please, refer to this comment about the importing User type:
https://github.com/huridocs/uwazi/pull/2913/files#r425370222

fnocetti:
@konzz I'm sorry there are some new conflicts after my PR just being merged into dev. Let's talk tomorrow first time in my morning and solve them together. They look pretty straightforward.

app/api/auth/authMiddleware.ts
konzz: I prefer to have the code the way we develop everything else and disable a rule that is clearly bugged.

konzz:
@RafaPolit Rebased and ready again

RafaPolit:
I'm sorry.  This is still not working correctly.

# Issue 1

There seems to be an issue with the values, when you have a combination of types and text filter in the aggregations.

## how to replicate

Using `uwazi fixtures`:
- select to filter by 'Ordenes del presidente'
- note the Firmantes section that it shows 5 firmantes + 8 more
- type on `e` on the `search item` field
- now you have the same 5 firmantes but + 23 more

There is no way that, by limiting the results I am actually getting MORE options.

# Issue 2

There are conditions where you get negative "more" numbers

## how to replicate

Using `uwazi fixtures`:
- select to filter by 'Ordenes del presidente'
- note the Firmantes section that it shows 5 firmantes + 8 more
- type on `ec` on the `search item` field
- now you have only 2 options
- click on the `x` button
- you have the original 5, but it now reads `-2 more`

konzz:
@RafaPolit Issues 1 and 2 where caused by the same problem, the Options returned where not properly filter by having results, so the amount was wrong.

About Issue 3, the types for `cookie` and `socket.io`  were installed and in the package.json file, so I deleted my node_modules folder, reinstalled and don't get those errors, but I get others:

```
app/api/csv/csvLoader.ts:16:32 - error TS2507: Type 'typeof EventEmitter' is not a constructor function type.

16 export class CSVLoader extends EventEmitter {
                                  ~~~~~~~~~~~~

app/api/csv/csvLoader.ts:63:16 - error TS2339: Property 'emit' does not exist on type 'CSVLoader'.

63           this.emit('entityLoaded', entity);
                  ~~~~

app/api/csv/csvLoader.ts:68:14 - error TS2339: Property 'emit' does not exist on type 'CSVLoader'.

68         this.emit('loadError', e, toSafeName(row), index);
```

RafaPolit:
We have reviewed this together with @konzz and there persists some inconsistencies with the remote response and the local filters.  There are also some scenarios of collapsed / expanded that trigger weird conditions.

I'm reporting here what we found:

# Issue 1

Missmatch between elastic results and local filters

## To reproduce

- With Uwazi fuxtures, filter `Ordenes del presidente`
- In `Firmantes`, type `e`
- Remote server replies 8 possible values and you see 5 values + 3 more
- If you expand, you see 12 values that all have `e`s in them.  Server has omitted, among others `Cecilia`

# Issue 2

Incorrect `X more` when options were expanded and new filter results in less options.

## To reproduce

- With Uwazi Fixtures, filter `Ordenes del Presidente`
- In `Firmantes` expand the `8 more` values
- Filter for `Ce`
- You have one option left: `Cecilia`, yet, it says you have 12 more options (curiously enough, server replied with 0 matches)

# Issue 3

Negative numbers persist on some combinations

## To reproduce

- With Uwazi Fixtures, filter `Ordenes del Presidente`
- In `Firmantes` type `Diego`
- You have a single remaining option
- Clear the selection by clicking on the `x`
- You have 5 options, but the `x more` reads `-4 more`

RafaPolit:
When adding template filters, if a porperty filter persists from one selection to another (it is common to both, or, for example, it is a common filter and you select a template type that also has that property), total results (x more) are not updated until a condition change (ie. typing in the input, or navigating a new page and returning to library).

RafaPolit:
When trying to search any term for template properties assigned to "any entity or document" there is an error replied by the server.

PR #2833, opened 2020-03-23
1151 local favorites

### LINKED ISSUE #1151: Implement a 'per user' or 'favorites' folder to make collaborative document organising easier
_As a user, I would like to be able to separate the documents I am working on (organising, creating connections, etc) from the rest of the document library. This is especially helpful when I am working with a team of people._

RafaPolit:
fixes #1151

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


RafaPolit:
To test this functionality, you need to add a `"favorites": true` `feature` into the settings document (in the database, there is no current UI method to achieve this).

That is, the item 0 of the settings collection should have:
```
{
  ...
  "features": {
    "favorites": true
  }
  ...
}
```

After modifying the register in the DB, once it is loaded through the model, Uwazi settings route would put an extra `_id` key.   So, to fully test, after testing the regular functionality, ensure that 'settings' procedures: save, edit, etc., are working correctly.

app/react/Favorites/components/FavoriteBanner.tsx
```diff
@@ -0,0 +1,55 @@
...
+          className={`btn favoriteBanner ${selected ? 'selected' : ''}`}
+          onClick={this.toggleClick}
+          type="button"
+        >
+          <span className="tab-link-tooltip">Add / remove favorite</span>
+        </button>
+      );
+    }
+
+    return null;
```
daneryl: @RafaPolit does not this code produce a `Expected server HTML to contain a matching <div> in <div>` type of error, when reloading any library with already some favorites ?

app/react/Favorites/components/specs/FavoriteBanner.spec.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  beforeEach(() => {
+    localStorageMock.setItem('uwaziFavorites', 'sharedId1,sharedId4');
+    mockEvent = jasmine.createSpyObj('mockEvent', ['stopPropagation', 'preventDefault']);
+  });
+
+  it('should present a clickable banner', () => {
+    render();
+    expect(component).toMatchSnapshot();
+  });
```
daneryl: @RafaPolit we should avoid full component snapshots or snapshots at all, i do not think this has any value, do not reflect any intent, in some cases are too big to debug and usually test implementation details like the label in this case that is not relevant to the functionality.

app/react/Favorites/index.ts
```diff
@@ -0,0 +1,3 @@
+import { FavoriteBanner } from './components/FavoriteBanner';
+
+export { FavoriteBanner };
```
daneryl: @RafaPolit  what you think about 
`export { FavoriteBanner } from './components/FavoriteBanner';`
instead of the 2 lines, this way its shorter and in some cases avoids object binding problems on circular dependencies.

app/react/components/Elements/FeatureToggle.tsx
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export type OwnPropTypes = {
+  feature: string;
+};
+
+const FeatureToggle: React.FC<ComponentPropTypes> = ({
+  featureActivated,
+  children,
+}: ComponentPropTypes) => (featureActivated ? <React.Fragment>{children}</React.Fragment> : null);
```
daneryl: @RafaPolit i do not think a fragment is needed here ? just 
`featureActivated ? children : null`, or im missing something ?

RafaPolit: Wouldn't that fail if children are an array of elements instead of a single one?  I'll test.  Maybe I'm thinking about previous versions or I am completely mistaken.

app/react/Favorites/components/FavoriteBanner.tsx
RafaPolit: Not with 'some' favorites... if the collection has the feature enabled, it will show the warning.  I'll look into making this a bit better, but since the value is coming from localStorage, there is no way to fully avoid the issue of server rendering different than client.

app/react/components/Elements/FeatureToggle.tsx
RafaPolit: I cannot seem to get rid of the React.Fragment without having the TS fail on the assignment as return of the Component.  I could use some help.

app/react/Favorites/components/FavoriteBanner.tsx
RafaPolit: Actually, the React documentation is really good on this matter:
https://reactjs.org/docs/react-dom.html#hydrate

The 'isClient' implementation was completely avoided with a ComponentDidMount post-check.  This has the advantage of rendering in server the banners already if the functionality is enabled.  The downside is that it requires a re-render of all favored items afterwards, but that was already the case.

Nice catch!

app/react/Favorites/components/specs/FavoriteBanner.spec.tsx
RafaPolit: Mostly agree.  Everything of this component was already tested afterwards, so there was absolutely no need for this test.  Removed it and extended the selectors that the 'required' classes are included.

app/react/components/Elements/FeatureToggle.tsx
daneryl: its ok, i thought only removing the fragment will work, but its not worth it to spend the time if that is not the case. its ok like this.

mylaa:
@daneryl just a quick questions: is this deployed on production?

PR #2836, opened 2020-03-25
Enable coverage reporting by default and configure vscode to show it

bdittes:
PR checklist:
- [x] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


daneryl:
@bdittes i do not use vscode, pls @konzz or @RafaPolit review, but i think we do not want coverage by default because its too slow ?

.vscode/extensions.json
```diff
@@ -0,0 +1,24 @@
+{
+  "recommendations": [
```
fnocetti, line 2: @bdittes introduced a number of extensions suggestions for vscode that I find interesting. I've installed most of them (except for the git ones) and are good tools.
Being suggestions imply they will be recommended by the editor but you don't need to install them mandatorily. 

What you guys think? @RafaPolit @konzz @daneryl @elric-wamugu @gabriel-piles

.vscode/extensions.json
```diff
@@ -0,0 +1,24 @@
...
+    "eamodio.gitlens",
+    "shd101wyy.markdown-preview-enhanced",
+    "davidanson.vscode-markdownlint",
+    "leizongmin.node-module-intellisens",
+    "eg2.vscode-npm-script",
+    "christian-kohler.npm-intellisense",
+    "christian-kohler.path-intellisense",
+    "esbenp.prettier-vscode",
+    "shan.code-settings-sync",
+    "ms-vscode.vscode-typescript-tslint-plugin",
```
fnocetti, line 20: @bdittes doesn't this conflict with us using eslint?

fnocetti:
> @bdittes i do not use vscode, pls @konzz or @RafaPolit review, but i think we do not want coverage by default because its too slow ?

@daneryl It seems to add around 1 extra minute in my local env. As for GitHub builds it seems to add around 40secs to a a ~8mins build.

Maybe we can leave it disabled by default. Generating coverage would be as simple as `yarn test --coverage=true`

.vscode/extensions.json
daneryl: @fnocetti what about not actually include specific ide/editor related configs on the project? is this a common practice ? can the ppl interested maintain a separate repo ? im ok if its merged here just to check what the others think

fnocetti: @daneryl that was my first thought but since we are currently maintaining a .vscode folder with configs I wanted to know what you folks think...

RafaPolit:
To be discussed on the Tech Call.

fnocetti:
Given the conversations with the tech team, I'm closing this PR. We already have coverage in the pipeline and devs are used to generating coverage on demand when needed. As for the VS Code configs and recommendations, I'm keeping the branch alive for now. I'm taking ownership on the task of finding a good place for the recommendations as the devs using VS Code found them useful, and are taking a look at them, but no specific IDE/Editor is recommended for Uwazi development.

PR #2837, opened 2020-03-25
translation throws warning instead of fatal error

### LINKED ISSUE #2831: default filters render result in app crash
when property as default filter renders label on library/uploads throws a translation error and app crashes. tagging this as critical as we can not merge/deploy to master until this is fixed.

@gabriel-piles @txau @RafaPolit apart from the translation solution, maybe we should "relax" a little bit this error and throw it as a warning to prevent crashes but having some info on things that are not properly translated.

daneryl:
fixes #2831

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/I18N/t.js
+++ b/app/react/I18N/t.js
@@ -7,7 +7,8 @@ const testingEnvironment = process.env.NODE_ENV === 'test';
 
 const t = (contextId, key, _text, returnComponent = true) => {
   if (!contextId) {
-    throw new Error(`You cannot translate "${key}", because context id is "${contextId}"`);
+    // eslint-disable-next-line no-console
+    console.warn(`You cannot translate "${key}", because context id is "${contextId}"`);
   }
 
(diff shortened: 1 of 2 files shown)


app/react/I18N/t.js
```diff
@@ -7,7 +7,8 @@ const testingEnvironment = process.env.NODE_ENV === 'test';
 
 const t = (contextId, key, _text, returnComponent = true) => {
   if (!contextId) {
-    throw new Error(`You cannot translate "${key}", because context id is "${contextId}"`);
+    // eslint-disable-next-line no-console
+    console.warn(`You cannot translate "${key}", because context id is "${contextId}"`);
```
RafaPolit: It's a weird use of 'you', but I think it's OK.

PR #2843, opened 2020-03-31
1898 chart options

### LINKED ISSUE #1898: Fix charts legends
Some of our charts are failing to properly display the labels of the graphed values:

screenshot from 2018-09-15 08-15-14

### LINKED ISSUE #2434: Allow passing an arbitrary list of properties to charts
Sometimes users want to display arbitrary properties, ie, for comparison, instead of top categories or all of them.

### LINKED ISSUE #2839: Ability to sort alphabetically in charts
The most obvious example where this is needed is years in charts (example below). This just **has to** be sorted alphabetically.

Also it is expected soon for some partners.
 
image

RafaPolit:
fixes #2839 
fixes #2434
fixes #1898

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Charts/utils/arrayUtils.js
+++ b/app/react/Charts/utils/arrayUtils.js
@@ -65,6 +65,28 @@ const limitMaxCategories = (sortedCategories, maxCategories, aggregateOthers) =>
   return categories;
 };
 
+const determineRelevantCategories = (data, formatOptions) => {
+  const { excludeZero, pluckCategories = [] } = formatOptions;
+
+  let relevantCategories = data.filter(i => i.key !== 'missing');
+
+  if (excludeZero) {
+    relevantCategories = relevantCategories.filter(i => i.filtered.doc_count !== 0);
+  }
+
+  if (pluckCategories.length) {
+    relevantCategories = pluckCategories.reduce((results, category) => {
+      const matchingCategory = relevantCategories.find(c => c.label === category);
+      if (matchingCategory) {
+        results.push(matchingCategory);
+      }
+      return results;
+    }, []);
+  }
+
+  return relevantCategories;
+};
+
 const formatPayload = data =>
   data.map((item, index) => ({
     value: item.name,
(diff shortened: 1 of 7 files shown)


RafaPolit:
This development includes three BarChart additions:
- sort columns by either result values or label names, both ascending and descending
- pluck specific columns from the results as to limit to specific label values instead of only by max categories
- allow to map the labels to shorter names so that long labels don't break the UI

-------------

## Sort

This functionality will use the sort attribute of the BarChart component. It should be a valid JSON, and accepts two keys: by and order.

<BarChart sort='{ "by": "result", "order": "desc" }' />
Possible values:
by: result | label|none
order: asc | desc

Default sorting is by result desc, exactly as it has been working up till now. If you select label as the sorting property, order defaults to asc. If you pass none (or any other unrecognized value), no sorting will happen. The none value is particularly useful if combined with the new pluckCategories attribute that allows you to hand-pick which bars to show (#2434).

------------------------------------

## Pluck

The attribute is called pluckCategories and it takes a valid JSON array of category names or labels.

<BarChart ... pluckCategories='["UN", "NATO", "WHO"]' />
This allows to hand-pick which columns to display. For example, if you have years on the X axis, you may be interested in seeing the progress in each decade change, instead of every year. For this, you can ask for just pluckCategories='["1990", "2000", "2010", "2020"]' . This will omit all the values in between.

It is possible to force a particular order of columns instead of sorting them by one of the other available methods (by label or by result). To accomplish this, just put the plucked categories in the order you want them to appear and combine with the attribute sort='{ "by": "none"}' as described in #2839.

-----------------------------------------

## Map

The implementation, as discussed, is to provide a map of shortened values.  This will be passed in the `shortLabels` property as a JSON object:

```
<BarChart ... shortLabels='{ "A long label": "shrtLabel", "United Nations (UN)": "UN" }' />
```

Each key of the JSON object represents a label, it's value represents the shortened version to be displayed.  The value will appear in the X axis (or Y if layout is `vertical`), but the tooltip on hover (or click on mobile devices) will still show the extended name.

------------------------

This individual explanations are also to be found on each individual issue in case further discussion is needed.

daneryl:
@RafaPolit works great, tested i think everything, the only weird behaviour i found is when combining shortLabels with sort by label, sorting will use the original label but the short label will be the one that is shown not matching with the actual sorting.

daneryl:
> @RafaPolit works great, tested i think everything, the only weird behaviour i found is when combining shortLabels with sort by label, sorting will use the original label but the short label will be the one that is shown, not matching with the actual sorting.

this is gona be discussed on a separate issue

RafaPolit:
Sorting after shortLabels is being tracked here:
#2852

PR #2850, opened 2020-04-02
958 export csv

### LINKED ISSUE #958: Export of metadata/data from Uwazi to csv
A user just wrote: "It would be great if we could also export data/metadata stored in Uwazi database to csv so that we can use the data for other purposes."

Have asked for a bit more context and, if possible, to join discussion here, but if the team has questions, feel free to ask.

fnocetti:
fixes #958

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


RafaPolit:
Still very much ongoing review, but I am seeing a lot of :any types on the .ts files.  I understand it's the quickest way, but not sure if we should make an extra attempt to not do that unless there is no other way around?  It kind of defeats the purpose of the typing.  

On tests I think we can be much more lenient, but on actual code, I think it's important to try to capture the actual types of variables?  What do you think @joan @konzz?

fnocetti:
@RafaPolit I actually agree with you! Most of the any-typed variables refer to arbitrary objects for which I will need to define types/interfaces (because, right, it was quicker to implement). But now since the code has its iterated design I should be able to make strong typing my next priority after unit-testing.

As for the SearchResults type: I couldn't find a type for that in the application. Do you know if we have a type definition? If not I will continue to expand the one I'm using. It won't probably be complete enough to put it under the shared/types (or similar) directory, IMHO. Do you thing it's worth to define it as an application-wide type?

app/api/files/jsRoutes.js
```diff
@@ -132,4 +132,79 @@ export default app => {
...
+  const parseQueryProperty = (query, property) =>
+    query[property] ? JSON.parse(query[property]) : query[property];
+
+  const generateExportFileName = databaseName => `${databaseName}-${new Date().toISOString()}.csv`;
+
+  app.get(
+    '/api/export',
+
+    validation.validateRequest(
+      Joi.object().keys({
```
RafaPolit, line 145: @joan should we start using the new JSON schema here or keep JOI for old route files (even if the endpoint is new)?

app/api/files/jsRoutes.js
```diff
@@ -132,4 +132,79 @@ export default app => {
...
+          exporter
+            .export(results, req.query.types, fileStream, exporterOptions)
+            .then(() => {
+              fileStream.end(() => {
+                res.download(temporalFilePath, generateExportFileName(site_name), err => {
+                  if (err) {
+                    return handleError(err);
+                  }
+
+                  return fs.unlinkSync(temporalFilePath);
```
RafaPolit, line 198: Even if this could be fast, it is still a blocking procedure.  While using this in tests may work, please us non-sync unlink here.  Also, fs.unlink returns undefined, so be sure this is the expected behavior for success download.

app/react/Library/actions/exportActions.js
```diff
@@ -0,0 +1,76 @@
+import superagent from 'superagent';
```
RafaPolit, line 1: If this is a completely new file, should this be a TS file?

RafaPolit, line 1: If this is a completely new file, should this be a TS file?

app/react/Library/actions/exportActions.js
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export function extractFileName(contentDisposition) {
+  const startIndex = contentDisposition.indexOf('filename="') + 10;
+  const endIndex = contentDisposition.length - 1;
+  return contentDisposition.substring(startIndex, endIndex);
+}
+
+export function exportDocuments(storeKey) {
+  // eslint-disable-next-line max-statements
```
RafaPolit, line 38: Please refactor this to contain less statements instead of disabling the lint rule.

app/api/files/jsRoutes.js
```diff
@@ -132,4 +132,79 @@ export default app => {
...
+                });
+              });
+            })
+            .catch(e => {
+              fs.unlinkSync(temporalFilePath);
```
RafaPolit, line 203: Again, use regular unlink.

app/react/Library/reducers/exportReducer.js
```diff
@@ -0,0 +1,9 @@
+import { combineReducers } from 'redux';
+import createReducer from 'app/BasicReducer';
+
+export default combineReducers({
+  exportProcessing: createReducer('exportProcessing', false),
```
RafaPolit, line 5: I'm wondering if this should be better namespaced?  Perhaps exportCSVProcessing?  or exportLibraryProcessing?  Also, there should be consistent naming across the store, because the exportEntity store entry point is missleading.  This is exportSearchResults, or exportLibrary.  ExportEntity should be a 'single' entity export and reserved for a different flow.

app/api/csv/typeFormatters.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const formatFile = (fileName: string) => `/files/${fileName}`;
+const formatAttachment = (fileName: string, entityId: string) =>
+  `/api/attachments/download?_id=${entityId}&file=${fileName}`;
+
+const formatters: {
+  [key: string]: (field: any, options?: any) => string;
+} = {
+  select: (field: any) => (field && field[0] ? field[0].label : ''),
+  multiselect: (field: any) => field?.map((item: any) => formatters.select([item])).join('|') || '',
```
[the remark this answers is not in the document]
fnocetti: I was not doing the checks at first, but I got errors because I got rows in the results for which the template indicated there should be a field, while it was actually not set in the row. The OR operand is related to the optional chaining of map. It will make the whole instruction return an undefined value and the OR defaults it to the empty string.

Anyways, this is clearly not the place to do those checks as the formatters should only transform an existing field. I'll remove them from here and try to reproduce the failing scenarios.

app/react/Library/actions/exportActions.js
```diff
@@ -0,0 +1,76 @@
...
+  return (dispatch, getState) => {
+    const state = getState()[storeKey];
+    let currentFilters = state.filters;
+    const { search } = state;
+    currentFilters = currentFilters.toJS ? currentFilters.toJS() : currentFilters;
```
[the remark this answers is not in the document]
fnocetti: I replicated how the filters are treated in the search endpoint. I'll double check.

app/react/Library/actions/exportActions.js
fnocetti: Yes, it should. I'll work on it.

app/react/Library/reducers/exportReducer.js
fnocetti: Yes, makes sense. I'll take care of it.

app/api/files/jsRoutes.js
daneryl: @RafaPolit @fnocetti i think we should remove gradually the joi validations, for sure for new routes use the new one, example here:

https://github.com/huridocs/uwazi/blob/988b5ddd02705852babfe1a5b925bb7c9ace3b7e/app/api/files/routes.ts#L98-L106

app/api/csv/csvExporter.ts
```diff
@@ -0,0 +1,217 @@
...
+import { EventEmitter } from 'events';
+import * as csv from '@fast-csv/format';
+import templates from 'api/templates';
+import { PropertySchema } from 'shared/types/commonTypes';
+import { TemplateSchema } from 'shared/types/templateType';
+import translate, { getLocaleTranslation, getContext } from 'shared/translate';
+import translations from 'api/i18n/translations';
+import formatters from './typeFormatters';
+
+export type SearchResults = {
```
RafaPolit, line 11: Everything is exported: all functions, all types.  Is this really necessary? The specs only import about 12 of the 16 exports here, and external modules should probably only import the default.

app/api/csv/csvExporter.ts
```diff
@@ -0,0 +1,217 @@
...
+  name: string;
+  common: boolean;
+};
+
+export type TemplatesCache = {
+  [id: string]: TemplateSchema;
+};
+
+export const getTypes = (searchResults: SearchResults, typesWhitelist: string[] = []) =>
+  typesWhitelist.length > 0
```
RafaPolit, line 45: Unnecessary compare operator here: `typesWhitelis.length` is enough, as it will be true if > 0 and false if not.

app/api/csv/csvExporter.ts
```diff
@@ -0,0 +1,217 @@
...
+      (memo, template) => (template ? { ...memo, [template._id]: template } : memo),
+      {}
+    )
+  );
+
+export const notDuplicated = (collection: any) => (item: any) =>
+  collection.findIndex((i: any) => Object.keys(i).every(key => i[key] === item[key])) < 0;
+
+export const excludedProperties = (property: PropertySchema) =>
+  !['geolocation', 'preview', 'markdown', 'nested'].includes(property.type);
```
RafaPolit, line 67: Does this mean that geolocation properties are also not included?  Or are they processed elsewhere?

app/api/csv/csvExporter.ts
```diff
@@ -0,0 +1,217 @@
...
+    {
+      label: 'Template',
+      name: 'template',
+      common: true,
+    },
+  ].concat(headers);
+
+export const concatCommonHeaders = (headers: ExportHeader[]) =>
+  headers.concat([
+    { label: 'Geolocation', name: 'geolocation', common: true },
```
RafaPolit, line 102: While Documents, Attachments and Published are 'higher order' properties, the `_geolocation` properties reside within the metadata, they are not common to all templates, and there can be several with different labels for a single template.  I'm having a hard time following the process of geolocation.  I think there is something wrong in the concept here.

app/api/csv/specs/csvExporter.spec.ts
```diff
@@ -0,0 +1,576 @@
...
+      {
+        _id: '594bc3b0bee8b3829aea937d',
+        label: 'DOB',
+        type: 'date',
+        name: 'dob',
+        id: '87289f51-21fb-4ff6-bb8d-8b7d66d86526',
+      },
+    ],
+  },
+};
```
RafaPolit, line 126: Perhaps this fixtures are long enough to merit a 'fixtrues' file like in many other api tests?

app/react/Library/components/ExportButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { exportDocuments } from 'app/Library/actions/exportActions';
+import { ExportStore } from '../reducers/ExportStoreType';
+
+export type ExportButtonProps = {
+  processing: boolean;
+  storeKey: string;
+  exportDocuments: (keyStore: string) => any;
+};
+
+export class ExportButton extends Component<ExportButtonProps, {}> {
```
RafaPolit, line 16: Do not export the component if it is connected.  You can import the connected component in the test and either use the `.WrappedComponent` property of the component, or you can use '.dive()' methods to inject the store into the component.  That also allows you to test the mapStateToProps directly, which in this case are not tested at all.

You can check the dive method on `/app/react/components/Elements/specs/FeatureToggle.spec.tsx`

app/react/UI/Icon/export-csv.js
```diff
@@ -0,0 +1,15 @@
+/**
+ *
+ *
+ * @format
+ */
```
RafaPolit, line 5: This is no longer required, please remove.

app/react/UI/Icon/library.js
```diff
@@ -88,6 +88,7 @@ import { faUsers } from '@fortawesome/free-solid-svg-icons/faUsers';
 import { faUserTimes } from '@fortawesome/free-solid-svg-icons/faUserTimes';
 import { faHandPaper } from '@fortawesome/free-solid-svg-icons/faHandPaper';
 import { saveAndNext } from './save-and-next';
+import { exportCsv } from './export-csv';
```
RafaPolit, line 91: Was there really no icon in the entire font-awesome library that we had to create our own icon? Couldn't we have picked an icon from the library and added it here? Can you elaborate on how this happened? Thanks.

package.json
```diff
@@ -207,6 +208,8 @@
     "@types/enzyme": "3.10.0",
     "@types/enzyme-adapter-react-16": "1.0.2",
     "@types/express": "^4.17.2",
+    "@types/fetch-mock": "^7.3.2",
+    "@types/hoist-non-react-statics": "^3.3.1",
```
RafaPolit, line 212: If you require 3.3.1, please also update the hoist-non-react-statistics to the ^3.3.1 please, lets not have a discrepancy of types and installed version.

app/react/UI/Icon/library.js
fnocetti: It was requested by @simonfossom from his work in the new look and feel. There is a download/export icon that could be useful but the icon he proposed is a merge between a file download and a grid that better represent a CSV export. 
Maybe we can have a conversation about that if we prefer avoiding to create custom icons.

app/api/csv/csvExporter.ts
fnocetti: You're right. It also suggests that the testing strategy might be wrong. I propose you not exporting the elements that are not explicitly tested right away, and assuming the tech debt of refactoring this class.

app/api/csv/csvExporter.ts
fnocetti: The geolocation property is treated as a special case because I found there is some different logic going on in its displaying across the app. I may be totally wrong and had misunderstood the concept. Let's have a chat asap.

app/api/csv/csvExporter.ts
fnocetti: Answer: https://github.com/huridocs/uwazi/pull/2850#discussion_r412167988

package.json
fnocetti: Good catch. I don't actually don't require this types. I'm confused as how this happened, but I suspect the editor did something on behalf of me and I overlooked it. Removing it. Thanks.

fnocetti:
I found a bug that I need to take care of.

RafaPolit:
This LGTM.

PR #2857, opened 2020-04-06
Fix updtelogs without mongoid

daneryl:
in this PR the bug on updatelogs without mongoId its properly fixed, i had to do again migration 23 as migration 24 to remove any other entries with null mongoId created during this period, this was possible even with the guard condition because happened on delete calling updateMany:
https://github.com/huridocs/uwazi/blob/e7bd217dc688e968bbab5422f2c2e0f77fc15f06/app/api/odm/model.ts#L52-L58

whenever an empty array of affectedIds is passed this method, this actually creates an entry with only the $set object properties, this is because of the **upsert: true** that actually makes no sense in this query as it is supposed to update only, i also think this behaviour makes no sense on mongo itself, probably missing some bigger context, anyway this was the cause.

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/odm/model.ts
+++ b/app/api/odm/model.ts
@@ -31,14 +31,6 @@ class UpdateLogHelper {
   }
 
   upsertLogOne(doc: mongoose.Document, next: (err?: mongoose.NativeError) => void) {
-    // temporary prevent creating updateLogs without mongoid, log error
-    if (!doc._id) {
-      errorLog.error(
-        `ERROR, Update Log received a doc without _id:\n ${JSON.stringify(doc, null, ' ')}`
-      );
-      return next();
-    }
-    //
     const logData = { namespace: this.collectionName, mongoId: doc._id };
     return asyncPost(async () => {
       await updatelogsModel.findOneAndUpdate(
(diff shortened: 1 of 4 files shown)


app/api/odm/model.ts
```diff
@@ -53,7 +44,7 @@ class UpdateLogHelper {
     await updatelogsModel.updateMany(
       { mongoId: { $in: affectedIds }, namespace: this.collectionName },
       { $set: { timestamp: Date.now(), deleted } },
-      { upsert: true, lean: true }
+      { lean: true }
```
RafaPolit, line 47: This indeed fixes the problem.

However, this opens the question as to whether there is a chance that the affectedIds array was, indeed, populated with IDs that did not exist yet on the database.  That should trigger a 'create' (or upsert) with individual IDs of the non-found values.  Still, this is required to fix the problem.

We should keep an eye on whether we have reports of syncs not updating as expected.

RafaPolit:
Merge whenever the CI passes.

PR #2868, opened 2020-04-14
gelocation label now is optional

daneryl:
validation for gelocation fields assume label should be a string, this fails when the label does not exist. most instances have this type of data where geolocation does not have a string.

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/entities/metadataValidators.js
+++ b/app/api/entities/metadataValidators.js
@@ -44,8 +44,7 @@ const isValidDateRange = value => {
 
 const isValidSelect = value => isString(value);
 
-const isValidGeolocation = value =>
-  isString(value.label) && isNumber(value.lat) && isNumber(value.lon);
+const isValidGeolocation = value => isNumber(value.lat) && isNumber(value.lon);
 
 const validateRequiredProperty = (property, value) => {
   if (property.required) {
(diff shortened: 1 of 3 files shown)


app/api/entities/metadataValidators.js
```diff
@@ -44,8 +44,10 @@ const isValidDateRange = value => {
 
 const isValidSelect = value => isString(value);
 
-const isValidGeolocation = value =>
-  isString(value.label) && isNumber(value.lat) && isNumber(value.lon);
+const isValidGeolocation = value => {
+  const validLabel = value.label ? isString(value.label) : true;
```
RafaPolit, line 48: This will also pass as true for `label: ""`.  Do we want to accept that scenario? Or if the key label is present it has to actually contain data as string?

RafaPolit: Also, `label: 0` will pass this validation as well.   Perhaps we want to check that the key exists rather than this approach?

daneryl: @RafaPolit label: '' is ok that is what optional i think means, actually if not present we have more in depth code that puts label to ''.

best solution was not to check for the string here on the custom validation and rely on the already schema generic validation for metadata.

https://github.com/huridocs/uwazi/blob/305a45101ea93ba8ff7aee84739be42368d2b9b6/app/shared/types/commonSchemas.ts#L28-L37

RafaPolit: Great!

PR #2881, opened 2020-04-23
Layout fixes 2020 04 23

### LINKED ISSUE #2861: "Sort by" selector in library is not floating on top of the library
The "sort by" selector on library is no longer floating on top of the cards, it's pushing them further down to make space for the selector.

### LINKED ISSUE #2877: Map view CSS are broken
Map is being displayed with a wrong height. Also, initial zoom seems to be off showing a blank map.

Screenshot from 2020-04-21 07-42-13

Screenshot from 2020-04-21 07-43-04

RafaPolit:
fixes #2877 , fixes #2861

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


nightmare/helpers/connectionsDSL.js
```diff
@@ -1,216 +1,302 @@
...
   },
   sortBy(orderByText, done) {
     this.waitToClick(selectors.connections.sortMenu)
-    .evaluate((sortBy) => {
-      const helpers = document.__helpers;
-      const sortOptions = helpers.querySelectorAll('.sort-buttons .Dropdown-option a');
+      .evaluate(sortBy => {
+        const helpers = document.__helpers;
+        const sortOptions = helpers.querySelectorAll(
+          '.sort-buttons > div.Dropdown.order-by.is-active .Dropdown-option a'
```
RafaPolit, line 17: This is the only "real" change to the document, everything else is just Prettier doing its thing.

app/react/Library/components/SortButtons.js
```diff
@@ -153,41 +154,53 @@ export class SortButtons extends Component {
...
-                  onClick={() => this.handleClick('_score')}
-                >
-                  <span>{t('System', 'Search relevance')}</span>
-                </a>
-              </li>
-            )}
-            {additionalSorts}
-          </ul>
-        </div>
+        {this.renderDropdown(search, additionalSorts, { includeEvents: false })}
```
fnocetti, line 202: Was duplicating the dropdown as a placeholder the best strategy? I can be ignoring something, but couldn't this be fixed on the HTML/CSS?

RafaPolit: It’s a quick patch, because the approach to position the “| select all” button to the right of an absolutely positioned element is tricky.  More so because it’s size varies according to content.

Obviously there are better approaches, and even this one could just include the text of the select instead of all the structure. But the “reality” of this implementation is that it is an invisible text behind the floating text (which is configured to render identically to the floating element).

I do agree we need to rethink all that bar.

If you have any different approach for this scenario, I’m all ears!

RafaPolit: Let me elaborate a bit further: using an in-line element that servers as a placeholder for the floating (or absolutely positioned) element is perfectly valid and common approach. The fact that it holds the text and classes of the floating element to match its size is also the correct approach.

Where I agree is that there is no need to duplicate, hidden, the other <li> elements that are not the 'selected' one.  So, I can invest time in taking them away, but the duplication of the underlying sort button is, in my opinion, the correct approach.

It will take an extra effort to remove the 'non essential' <li> elements.  This can be done, for sure.

fnocetti: Agreed on going with this implementation.

fnocetti:
LGTM, just waiting for the tests to pass.

PR #2891, opened 2020-04-28
Dont display the preview field if no document associated

### LINKED ISSUE #2884: "Preview" property should not appear if no file uploaded.
Preview property renders a preview of the uploaded PDF.  Previously, a template was either entity or document, and preview could only be assigned to document templates.  Furthermore, the only way to 'start' a document entity was to upload a PDF, so preview was available always.  Still, I remember there was a 'no preview' file for those scenarios.

Since now any entity can be either just entity or document, and any template can hold a preview, there is a bug that Entities that have no PDF uploaded are rendering an 'undefined' preview.

If no PDF has been uploaded, just like other metadata properties, this should not show the preview label and value neither in the side panel nor in the cards.

Optionally, we could make it point to the previous 'no preview' IMG which was a route that catched this errors, perhaps that got broken somehow.

fnocetti:
fixes #2884 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Metadata/helpers/formater.js
+++ b/app/react/Metadata/helpers/formater.js
@@ -137,7 +137,11 @@ export default {
 
   preview(property, _value, _thesauris, { doc }) {
     const defaultDoc = doc.defaultDoc || {};
-    return this.multimedia(property, [{ value: `/api/files/${defaultDoc._id}.jpg` }], 'image');
+    return this.multimedia(
+      property,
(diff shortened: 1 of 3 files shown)


fnocetti:
I don't think it makes sense to fix the issues reported by codeclimate, but we can take a look.

PR #2895, opened 2020-04-30
2511 graphs e2e tests no puppeteer

### LINKED ISSUE #2511: Create more extensive Page and Graphs-in-Page e2e Tests

elric-wamugu:
fixes #2511

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


nightmare/paths/graphs.spec.js
```diff
@@ -0,0 +1,91 @@
...
+    it('should insert Bar chart graph in created page', async () => {
+      nightmare
+        .evaluate(
+          selector => document.querySelector(selector).value,
+          localSelectors.pageContentsInput
+        )
+        .then(response => {
+          expect(response).toBeTruthy();
+          expect(response).toContain('<Dataset />');
+        });
```
daneryl, line 67: @eric this should be async too instead of promise, actually the test probably finishes before the code inside then is executed.

elric-wamugu:
> @eric, looks good, ive made a comment on one of the tests, and important, you should not use specific time waits, this is prone to errors depending on the machine, pls wait for elements you know must be there for the page to be ready instead.

@daneryl The reason for the wait here is because of the graph's animation which happens after the render. Otherwise the screenshot output may happen when the animation has still not yet finished

daneryl:
> > @eric, looks good, ive made a comment on one of the tests, and important, you should not use specific time waits, this is prone to errors depending on the machine, pls wait for elements you know must be there for the page to be ready instead.
> 
> @daneryl The reason for the wait here is because of the graph's animation which happens after the render. Otherwise the screenshot output may happen when the animation has still not yet finished
> 
> The attached screenshot captures that scenario before and after just a 2 seconds wait, the red part showing the difference between the 2 captures
> graphs-spec-js-pages-path-graphs-in-page-should-display-bar-chart-graph-in-page-with-no-more-than-a-1-difference-1-diff

ok then, if there is no way of removing the animation on the test like we did on the rest of the app, this was via css, i guess this one is a js animation? anyways leave the wait there if that is the case but abstract it on an action with a semantic name to express the intent of the wait.

nightmare/paths/graphs.spec.js
```diff
@@ -0,0 +1,127 @@
...
+    it('should create a basic page', async () => {
+      await nightmare
+        .goto('http://localhost:3000/en/settings')
+        .clickLink('Pages')
+        .clickLink('Add page')
+        .wait('.page-creator')
+        .write(localSelectors.pageTitleInput, 'Page data viz')
+        .write(localSelectors.pageContentsInput, '</p><Dataset />')
+        .clickLink('Save')
+        .wait(1000);
```
daneryl, line 53: @eric remove this wait, a way to do it is to wait for the success notification and clicking on it to close it.

nightmare/paths/graphs.spec.js
```diff
@@ -0,0 +1,127 @@
...
+
+      await nightmare.getInnerText(localSelectors.createdPageLink).then(text => {
+        expect(text).toContain('(view page)');
+      });
```
daneryl, line 57: @eric lets avoid the use of 'then' when using awaits for consistency, use `const text = await nightmare.getInnerText` notation.

nightmare/paths/graphs.spec.js
```diff
@@ -0,0 +1,127 @@
...
+    it('should insert Bar chart graph in created page', async () => {
+      await nightmare
+        .evaluate(
+          selector => document.querySelector(selector).value,
+          localSelectors.pageContentsInput
+        )
+        .then(response => {
+          expect(response).toBeTruthy();
+          expect(response).toContain('<Dataset />');
+        });
```
daneryl, line 69: @eric lets avoid the use of 'then' when using awaits for consistency, use `const response = await nightmare.evaluate` notation.

nightmare/paths/graphs.spec.js
```diff
@@ -0,0 +1,127 @@
...
+            .screenshot()
+            .then(image => {
+              expect(image).toMatchImageSnapshot({
+                failureThreshold: 0.01,
+                failureThresholdType: 'percent',
+                allowSizeMismatch: true,
+              });
+            });
+        });
+    });
```
daneryl, line 97: @eric avoid use of then.

nightmare/paths/graphs.spec.js
```diff
@@ -0,0 +1,127 @@
...
+        .write(localSelectors.pageContentsInput, '</p><Dataset />')
+        .write(localSelectors.pageContentsInput, graphs.pieChart)
+        .write(localSelectors.pageContentsInput, graphs.listChart)
+        .click(localSelectors.savePageButton);
+    });
+
+    it('should display Pie-List chart graphs in page with no more than a 1% difference', async () => {
+      await nightmare
+        .evaluate(selector => document.querySelector(selector).href, localSelectors.createdPageLink)
+        .then(link => nightmare.goto(link));
```
daneryl, line 113: @eric we discussed about this, i do not remember why this is not just a clickLink("Label") ??

nightmare/paths/graphs.spec.js
```diff
@@ -0,0 +1,127 @@
...
+        .waitForGraphsAnimation()
+        .screenshot()
+        .then(image => {
+          expect(image).toMatchImageSnapshot({
+            failureThreshold: 0.01,
+            failureThresholdType: 'percent',
+            allowSizeMismatch: true,
+          });
+        });
+    });
```
daneryl, line 125: @eric avoid use of then.

nightmare/paths/graphs.spec.js
elric-wamugu: @joan the initial issue with clickLink was that it was opening a new window on links with the `target="_blank" attribute and Nightmare have no way yet to switch focus from the old to the new window. Some of the scope of that issue is documented here https://github.com/segmentio/nightmare/issues/804. I however, modified the `clickLink` action to remove the target attribute of an anchor element to avoid opening the link in a new window

daneryl:
@elric-wamugu looks good.

For some reason the tests are failing for me with 
```
Expected image to be the same size as the snapshot (1200x575), but was different (1200x578)
```
even though the allowSizeMismatch is set to true. maybe we should try to screenshot only the chart itself ? do not know if that is possible with nightmare.

daneryl:
We are not merging this, waiting to move to pupeteer to reimplement.

Issue #2894, opened 2020-04-30
Unify e2e fixutres and uwazi fixtures into a single collection


RafaPolit:
It would be useful for the e2e to have a more representative collection of data. We think using the current Fixtures, instead of the e2e fixtures would be a better route.

We can configure this as a submodule so that it doesn't get downloaded with the main repo.

Issue #1852, opened 2018-09-03
Google Fonts are being loaded in e2e testing


txau:
This is slowing down tests. We could either mock this call or server it locally instead of from a CDN.

PR #2898, opened 2020-05-01
e2e tests

### LINKED ISSUE #2905: Puppeter e2e testing scaffold
Have a couple of basic tests based on puppeteer passing in CI as a base for e2e revamping.

txau:
fixes #2905 #2894 #1852 #2950 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [x] Code review


daneryl:
@txau main problem i see now is that we actually need prod built in order to run this, if you try to run it without prod built, fixtures fail. should we allow to run this normally and with prod optionally ?

daneryl:
@txau im thinking that ideally yarn e2e, runs its own instance to run the e2e ? this way you can run the e2e with a single command and it takes care of everything ?
but apart from that im happy with the documentation or feedback options for now ...

daneryl:
@txau i have added a message when running e2e-puppeteer to remind that uwazi needs to run with specific db and index. i think for now that is enough.

e2e/suites/entity.test.ts
```diff
@@ -1,10 +1,12 @@
...
 
-import { adminLogin, logout } from './helpers/login';
-import proxyMock from './helpers/proxyMock';
+import { adminLogin, logout } from '../helpers/login';
+import proxyMock from '../helpers/proxyMock';
+import insertFixtures from '../helpers/insertFixtures';
 
 describe('Entities', () => {
   beforeAll(async () => {
+    await insertFixtures();
```
elric-wamugu, line 9: Hey @txau, why do we have to run the fixtures on every test suite? Maybe we should include the e2e fixtures script while calling the e2e script? I think this will save a lot of time while running the tests.

txau: Actually @daneryl suggested this approach so we know the suites always start at a clean state and can be run independently. Also because of that, we have spent some time optimizing the fixture insertion (ie, by using transpiled scripts) and we removed plenty of document content so indexation is very fast. I think the current times are acceptable?

daneryl: we already had this scenario on the nightmare e2e fixtures where some testSuites fail because they depend on some specific data that the previous test had changed.

current times are good and with transpiled version for running on ci are a lot better, this is definitely an improvement over the old e2e where this same step can take almost 30secs now its more around the 5 secs with the production version.

elric-wamugu: Ok. sounds good

PR #2906, opened 2020-05-07
Enabled public submit endpoint to send emails

konzz:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/contact/specs/contact.spec.js
```diff
@@ -27,7 +27,7 @@ describe('contact', () => {
         })
         .then(() => {
           expect(mailer.send).toHaveBeenCalledWith({
-            from: '"Uwazi" <no-reply@uwazi.io',
+            from: '"Uwazi" <no-reply@uwazi.io>',
```
RafaPolit, line 30: Not sure if this is the time to address this, but the fact that we send all our mails as "Uwazi" is kind of confusing.  People use a document management platform "powered by Uwazi", but a lot of users would expect the mail to come from "Plan International", or "CEJIL", not from "Uwazi", as 90% of the people using the platform (other than admins) have no idea what Uwazi is.

Could we just change this to the collection name?  But some collection names are whole books, so not sure.  Perhaps a configurable option (another?) in collection settings?  Should we do this here and now? (especially since this opens up the scenario to sending of a LOT more emails).

app/api/contact/contact.js
```diff
@@ -5,7 +5,7 @@ export default {
   async sendMessage({ email, name, message }) {
     const siteSettings = await settings.get();
     const mailOptions = {
-      from: '"Uwazi" <no-reply@uwazi.io',
+      from: '"Uwazi" <no-reply@uwazi.io>',
```
RafaPolit, line 8: Not sure if this is the time to address this, but the fact that we send all our mails as "Uwazi" is kind of confusing. People use a document management platform "powered by Uwazi", but a lot of users would expect the mail to come from "Plan International", or "CEJIL", not from "Uwazi", as 90% of the people using the platform (other than admins) have no idea what Uwazi is.

Could we just change this to the collection name? But some collection names are whole books, so not sure. Perhaps a configurable option (another?) in collection settings? Should we do this here and now? (especially since this opens up the scenario to sending of a LOT more emails).

app/api/files/jsRoutes.js
```diff
@@ -76,6 +85,11 @@ export default app => {
           socket(req).emit('documentProcessed', newEntity.sharedId);
         });
       }
+
+      if (req.body.mail) {
+        mailer.send({ from: `"${req.body.mail.sender}" <no-reply@uwazi.io>`, ...req.body.mail });
```
RafaPolit, line 90: The validation defines the "email" key object, yet here you use the "mail" key.  How is this even working?

app/api/files/jsRoutes.js
```diff
@@ -46,13 +47,21 @@ export default app => {
...
+      Joi.object().keys({
+        entity: saveSchema,
+        email: Joi.object().keys({
+          to: Joi.string(),
+          sender: Joi.string(),
+          text: Joi.string(),
+          html: Joi.string(),
+          subject: Joi.string(),
+        }),
+      })
```
daneryl, line 60: validateRequest already supports using json schema, please use that for new validations so we can progressively remove joi validations, example in the same file.  

https://github.com/huridocs/uwazi/blob/5a9d015835141b16187708f3f86d64f800a01970/app/api/files/jsRoutes.js#L151-L174

app/api/files/jsRoutes.js
konzz: it wasn't

app/api/contact/contact.js
konzz: I agree with the problem, but that hardcoded email is for the contact form, I was just adding a missing ">"

The public form will use a "From" coming from the custom JS for the custom public form.
Adding the option to setings will be nice, but since that is the only form using that functionality, I think that adding that option in settings should be done when we modify the PublicForm to be able to configure the send an email option, or we will be adding an. option to settings that no one can use except for the contact form but I think if we do it for the contact form, it should be in another issue to do not extend this one.

RafaPolit: Yes, I agree.  That's why I mentioned this was probably not the place for this discussion.

RafaPolit: Created issue #2912 to track this development.

app/api/files/jsRoutes.js
```diff
@@ -46,13 +46,23 @@ export default app => {
     '/api/public',
     multer().any(),
     captchaAuthorization(),
-    (req, _res, next) => {
-      req.body = JSON.parse(req.body.entity);
-      return next();
-    },
-    validation.validateRequest(saveSchema),
+    validation.validateRequest({
+      entity: {
```
RafaPolit, line 50: Every other route that uses the non-Joi validation has a structure of the type:
```
{
  porperties: {
    *key* (ie: entity): {
      poperites: {
        *key* (ie: title) : { type: 'string' }
      }
  }
}
```
Is this form a shorthand and the extra keys are no longer required?  If so, why are we using them on every other route?

app/api/files/jsRoutes.js
```diff
@@ -76,6 +86,11 @@ export default app => {
           socket(req).emit('documentProcessed', newEntity.sharedId);
         });
       }
+
+      if (req.body.email) {
```
RafaPolit, line 90: This 'if' suggests the possibility of the `email` not being present on the request, yet every test scenario injects the body.email key.  Please, extend the test cases to a scenario where that is not the case.

app/api/files/jsRoutes.js
```diff
@@ -76,6 +86,11 @@ export default app => {
...
+        mailer.send(JSON.parse(req.body.email));
```
RafaPolit, line 91: Should this somehow be wrapped in a try - catch scenario? What happens if the `email` json is incorrectly formed? The parse here will break and we will end up with an un-responded `res`.  Should there be more `next(err)` contemplated here?

Perhaps, even more critical, mailer is a Promise, which is been implemented here without an await, then or catch.  Is this prone to uncatched rejection scenarios?

app/api/files/jsRoutes.js
daneryl: this validation probably fails, the idea is to actually include in the validator which part of the req you are validating, body, query, params or any other in this case probably 
```
{
  properties: {
    body: {
      properties: {
        *key* (ie: title) : { type: 'string' }
      }
  }
}
```

also we already have an entity schema json defined that i think should be used if this is expected to be a full entity ?

konzz: The validation wasn't failing probably because without the "properties" key it just ignores everything and doesn't matter what you send.

app/api/files/jsRoutes.js
konzz: Done

app/api/files/jsRoutes.js
konzz: About what happens if the email JSON is malformed, I think that if we add validation to the request is to trust the structure of the data we receive, if we don't trust that structure we lack a lot of try-catch in all the app, but I guess the mailer can fail because of wrong configuration in settings, that I don't think we can have it validated.

app/api/files/jsRoutes.js
```diff
@@ -47,36 +47,66 @@ export default app => {
     multer().any(),
     captchaAuthorization(),
     (req, _res, next) => {
-      req.body = JSON.parse(req.body.entity);
-      return next();
+      req.body.entity = JSON.parse(req.body.entity);
```
RafaPolit, line 50: Already discussed with @konzz:
- wrap this in a try/catch to prevent any uncaught errors due to malformed requests.

konzz: Done

PR #2913, opened 2020-05-08
2903 intermitent test fix

### LINKED ISSUE #2903: Fix intermitent tests for CSV Export
- [x] Use supertest to test res.download and not route return
- [x] Check validation (specially `search` which is currently an array)
- [x] Test properly the file unlink, this probably means mocking the file generation naming.  Please ensure that the name can not be duplicated if tests are run in paralell
- [x] Test for correct passing of the user to the search function (perhaps @daneryl can help set this up quicker)
- [x] test that the exporter got called correctly with the desired params, even if its reply is mocked.  A good approach is to tailor the reply according to the arguments passed, in that way you are testing, with the response, the arguments passed.
- [x] Look if there is merit to create a separate file for the `api/export/` route?

fnocetti:
fixes #2903 

This PR also includes the migration of some files to TS, and some minor refactors.

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/files/exportRoutes.ts
```diff
@@ -0,0 +1,92 @@
...
+          // eslint-disable-next-line camelcase
+          ([results, { dateFormat, site_name }]) => {
+            const exporter = new CSVExporter();
+            const temporalFilePath = temporalFilesPath(
+              generateFileName({ originalname: 'export.csv' })
+            );
+            const fileStream = createWriteStream(temporalFilePath);
+            const exporterOptions = { dateFormat, language: req.language };
+
+            exporter
```
RafaPolit, line 74: why is this not an async route that awaits here for the exporter?  same for the promise all on line 63.  This callback hell will probably look much better, and there would not be to have two catches calling next.

app/api/files/jsRoutes.js
```diff
@@ -51,6 +47,7 @@ export default app => {
       return next();
     },
     validation.validateRequest(saveSchema),
+    // eslint-disable-next-line max-statements
```
fnocetti, line 50: I tried to reduce the statements in this function first, but it is pretty compact and is old code. I don't think it's worth fixing this at this point.
If you disagree and have an idea on how to achieve it, let me know and I'll be happy to do it! :)

RafaPolit: No, it's OK.  That is not even part of this development, and this file no longer contains anything you have developed, so no, it's OK.  Let's leave it as is.

app/api/files/specs/exportRoutes.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+  expect(mockCall[1]).toEqual(types);
+  expect(mockCall[2] instanceof Writable).toBe(true);
+  expect(mockCall[3]).toEqual(options);
+}
+
+describe('export routes', () => {
+  describe('/api/export', () => {
+    let exportMock: any;
+
+    beforeAll(() => {
```
RafaPolit, line 31: This is a weird scenario, and works because you have a single `it` on this test.  If someone else includes a previous `it` for whatever reason, the `exportMock.mock.calls[0]` on line 99 will fail as there is a single mock shared by all tests.

That is not a good practice.  My suggestion is to move this into the 'before each' so that you get the jest spy reset every time you run a test and, therefore, adding further tests (or running them on different orders!) would not affect this spy.

fnocetti: Makes total sense, good catch, thank you!

app/api/files/exportRoutes.ts
```diff
@@ -0,0 +1,86 @@
...
+import { User } from 'api/users/usersModel';
+import settings from 'api/settings';
+
+type Middleware = (
+  req: { user: User; language: any; query: any },
+  res: { download: any },
+  next: any
+) => Promise<void> | void;
+type Get = { (arg0: string, arg1: Middleware, arg2: Middleware): void };
+type App = { get: Get };
```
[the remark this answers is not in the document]
fnocetti: I duplicated the strategy used in another route because I needed the request to have those extra attributes, but I'm sure there's a better way using Express types.
I'll try to improve it.

fnocetti:
@RafaPolit @daneryl please check again.
I implemented the changes requested.
While improving the typings I followed the strategy that we talked and replicated what @konzz did in his PR to avoid/simplify conflicts later. You'll see several files modified because I had to covert two middlewares to TS and update their references (just like konz).

app/api/auth/authMiddleware.ts
```diff
@@ -0,0 +1,23 @@
+import { Request, Response, NextFunction } from 'express';
+
+declare global {
+  namespace Express {
+    export interface Request {
+      user: import('../users/usersModel').User;
```
fnocetti, line 6: This differs from konzz's PR.
Using the import statement avoids having to disable the rule for not used vars and it is the preferred way of importing external types to a global definition based on several articles.

app/api/files/specs/exportRoutes.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+      assertDownloaded(res);
+      expect(search.search).toHaveBeenCalledWith(
+        {
+          filters: '',
+          types: ['types'],
+          fields: '',
+          aggregations: '',
+          select: '',
+          unpublished: '',
+          includeUnpublished: '',
```
RafaPolit, line 92: I am not ultra convinced about just 'replicating' here the entire query values of lines 74 ~ 80.

I'll merge this as is for now, but perhaps a refactor is in order.

RafaPolit:
LGTM!

PR #2916, opened 2020-05-12
test out different notification depending on error message string

### LINKED ISSUE #2798: Users see "Could not reach server, please try again later" message after log-in
There are a number of Uwazi instances for which when I log-in, I see a message for about 5 seconds that says "Could not reach server, please try again later", but then everything works fine. This has been happening for about 2-3 months. I don't think it's causing any problems, but it is confusing to see this message when there is actually nothing wrong.

elric-wamugu:
fixes #2798

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


elric-wamugu:
@RafaPolit @daneryl @txau A workaround for the short WebSockets disconnect error is putting a small delay in milliseconds before reloading the landing page after a successful login. However, for instances with a large initial library load request that exceeds the set delay because of their database size (eg UPR-Info), I have set the user notification to a 'Please wait, loading...' message. 
Would this be a sufficient fix?

RafaPolit:
@elric-wamugu will research why is this not happening on language change.

elric-wamugu:
For more context in the differences between window.location.assign()/window.location.href vs window.location.replace()/window.location.replace - https://stackoverflow.com/questions/10302905/location-href-property-vs-location-assign-method#comment107863541_10302942 and https://www.samanthaming.com/tidbits/86-window-location-cheatsheet/#how-to-do-a-page-redirect

PR #2918, opened 2020-05-14
Library pagination

### LINKED ISSUE #2892: Library Pagination
We need to have pagination on the library to avoid reloading and rerendering the previous cards that are already loaded.

This is also need for: #2708 and #958

konzz:
fixes #2892

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/search/search.js
```diff
@@ -447,6 +448,10 @@ const instanceSearch = elasticIndex => ({
         documentsQuery.limit(query.limit);
       }
 
+      if (query.offset) {
+        documentsQuery.offset(query.offset);
```
RafaPolit, line 452: I can't seem to find any test covering an 'offset' scenario.  I am missing something?

app/api/search/specs/routes.spec.js
```diff
@@ -44,14 +44,10 @@ describe('search routes', () => {
       expect(response).toEqual(expectedResults);
     };
 
-    it('should have a validation schema', () => {
-      expect(routes.get.validation('/api/search')).toMatchSnapshot();
-    });
```
RafaPolit, line 49: It's good that we no longer test the validation structure.  But now we are not testing validation at all in this scenario.  How should we approach this?

app/react/Layout/DocumentsList.js
```diff
@@ -39,26 +39,28 @@ class DocumentsList extends Component {
     }
   }
 
-  loadMoreDocuments(amount) {
+  loadMoreDocuments(amount, offset) {
```
RafaPolit, line 42: Should we take into account duplicates?

Example:
- You load library with 30 documents sorted by recently added
- Someone adds a new entity
- You select 'load more'

The last of the 30 documents will now be the first of the second batch, leaving you with a duplicate card.

Should we address that?

fnocetti:
I tested the CSV Export and it is working fine after this changes. The exporter is built to export the maximum amount of rows allowed, and the offset value in this PR is optional and defaulted to 0.

app/api/search/specs/routes.spec.js
daneryl: @RafaPolit that is a good question, do you have any suggestion ?

1.  test validations that belong to the route as an integration test.
2.  test the validations itself separately, this makes more sense when reusing validations obviously, i think this particular one is duplicated now on the export csv.

app/react/Layout/DocumentsList.js
daneryl: confirmed this happens, and it gets really bad if someone published multiple documents at once, lets say 30, in that case the next batch load will be exactly the same documents as the first batch. 😂
how we can address that ? 
removing duplicated on client app will result on inconsistent load more numbers, you request for 30 more, 28 are loaded even though there is enough to load 30. which case do we prefer to have ? i to not have duplicated is better ? any other ideas @konzz @RafaPolit

app/api/search/specs/routes.spec.js
konzz: @joan @RafaPolit 
Test the validation will be testing a configuration not an implementation.
Those are the kind of tests that only break when you are intentionally changing the configuration and not because by mistake you break the functionality, the concept is basically the same as doing a snapshot of the config. 

Integration tests between the route model and database are ok but still not great, the thing is that the method, the path, and the validation of the route are a contract between the client and the server, and the best way to have that tested is with the e2e.

app/react/Library/helpers/specs/resquestState.spec.js
```diff
@@ -59,25 +60,6 @@ describe('static requestState()', () => {
...
-
-    expect(libraryHelpers.URLQueryToState).toHaveBeenCalledWith(
-      q,
-      templates,
-      thesauris,
-      relationTypes,
-      ['country']
-    );
-  });
-
```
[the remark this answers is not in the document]
konzz: Is testing the implementation of requestState expecting a helping method to be called.

app/react/Uploads/UploadsRoute.js
```diff
@@ -92,11 +95,18 @@ export default class Uploads extends RouteHandler {
...
-    if (nextProps.location.query.q !== this.props.location.query.q) {
-      return this.superComponentWillReceiveProps(nextProps);
+    const nextQuery = decodeQuery(nextProps.location.query);
+
+    if (!nextQuery.offset) {
+      this.emptyState();
+    }
+
+    if (this.urlHasChanged(nextProps)) {
+      this.getClientState(nextProps);
```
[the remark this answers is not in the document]
konzz: About the question of the difference between Library and Uploads, not sure if there is any, I've unified code, making Uploads use bits of code of the Library, didn't want to spend more time on it and did the obvious refactors, avoiding to do a full refactor of uploads request state.

About bugs on uploads, probably add conditionals there, but after testing different places to empty the list I think componentWillReceiveProps is the best place, but if you can suggest a different implementation I can look into it.

app/react/Library/helpers/requestState.js
```diff
@@ -57,8 +63,9 @@ export default function requestState(request, globalResources) {
         },
       };
 
+      const replaceOrAddEntities = Boolean(docsQuery.offset);
       return [
-        setReduxState(state),
+        setReduxState(state, replaceOrAddEntities),
```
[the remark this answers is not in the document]
konzz: updated the name

app/react/Library/reducers/specs/documentsReducer.spec.js
```diff
@@ -27,6 +27,24 @@ describe('documentsReducer', () => {
...
+      });
+
+      newState = documentsReducer(newState, {
+        type: types.ADD_DOCUMENTS,
+        documents: { rows: [documentTwo], totalRows: 10 },
+      });
+
+      expect(newState.toJS()).toEqual({ rows: [documentOne, documentTwo], totalRows: 10 });
+    });
+  });
```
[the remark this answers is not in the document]
konzz: Changed the total rows number in the second call to test it better.

app/react/Library/helpers/specs/resquestState.spec.js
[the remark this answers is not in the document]
daneryl: is this somehow covered on another test ? even if the test is a bad one ?

app/react/Uploads/UploadsRoute.js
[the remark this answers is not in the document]
daneryl: do not know any other places, but on componentWillReceiveProps whats obvious is that introduces a bug that we will trigger without knowing most likely.

app/api/search/specs/routes.spec.js
daneryl: maybe open this on a new issue to implement separately as this was tested already wrong before this ?

e2e are hard to maintain, i dont see why this tests should be there when the react app has nothing to do with this apart from calling the endpoint.

data endpoint validation is the server side concern, i agree that testing the validation by comparing the object is the same bad idea as the snapshot testing, integration tests using supertest should be quick to develop and you dont really need to go till the DB, just test some failing cases, the db integration tests are already there.

app/react/Uploads/UploadsRoute.js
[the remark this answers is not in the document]
daneryl: > About the question of the difference between Library and Uploads, not sure if there is any, I've unified code, making Uploads use bits of code of the Library, didn't want to spend more time on it and did the obvious refactors, avoiding to do a full refactor of uploads request state.

the componentWillReceiveProps of both are diferent, on Library there is no emptyState() like in uploads.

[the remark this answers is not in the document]
konzz: I see, yes I think I can make the same approach than in Library

app/api/search/specs/routes.spec.js
konzz: When we did the metadata object change we had validations and tests for them on the server-side of the app, but we had the app crashing in many places because the client was not properly updated, even if we add Supertest test, we can be in the same scenario.

konzz: I guess there are 2 things to test, the positive, that is that the client - backend can communicate, and the negative that is that you cant inject data that is malformed.

app/react/Library/helpers/specs/resquestState.spec.js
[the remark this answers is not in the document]
konzz: There are 2 other tests calling requestState, and that method is always called inside requestState.

app/react/Uploads/UploadsRoute.js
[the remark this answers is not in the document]
konzz: I've changed Uploads to use the same setReduxstate as Library and doesn't need to have that empty state there anymore.

app/react/Layout/DocumentsList.js
RafaPolit: The team has decided not to address this for the time being.

app/api/search/specs/routes.spec.js
RafaPolit: The full length of this discussion is to be discussed in #2934

app/api/search/search.js
konzz: Removed offset method since the query builder already had a method using the key "from" that is tested.

PR #2919, opened 2020-05-14
2075 mapbox hardcoded user key

mfacar:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/settings/settings.js
```diff
@@ -82,9 +84,15 @@ function removeTemplate(filters, templateId) {
   });
 }
 
+function setDefaults(storedSettings) {
+  const [first] = storedSettings;
```
RafaPolit, line 88: `[first]` is just `[settings]`.  MongoDB works with array of documents, but we are using just a single document.  So it's not a 'first' settings, as suggested by the naming, the first document contains the ONLY settings, so I think just naming them 'settings' is enough.

app/api/settings/specs/settings.spec.js
```diff
@@ -190,6 +190,11 @@ describe('settings', () => {
       const values = await settings.get();
       expect(values.publicFormDestination).not.toBeDefined();
     });
+
+    it('should return default values', async () => {
+      const values = await settings.get();
+      expect(values.mapTilerKey).toEqual('QiI1BlAJNMmZagsX5qp7');
+    });
```
RafaPolit, line 197: This test covers only the scenario of returning the default mapTilerKey.  But it does not test the fact that if the DB already contains such a key, it should return that.  Please, try to approach this as TDD: write the test that expects that, if the DB already has a value, to be that value instead of the default.  See the test fail, implement only the code required to make that small part fail.

app/react/Map/Map.js
```diff
@@ -321,6 +323,13 @@ export default class Map extends Component {
     return false;
   }
 
+  async replaceKeysMapStyleJson() {
+    const mapTemplateEntry = /{{MAP_TILER_KEY}}/g;
```
RafaPolit, line 327: I don't really understand this variable name?  should this be `mapTilerKeyPlaceholder`, or something in those lines?

app/react/Map/specs/Map.spec.js
```diff
@@ -218,6 +224,15 @@ describe('Map', () => {
...
     });
+
+    it('should read and set the map tiler key in the style', () => {
+      const expectedKey = 'abc';
+      settingsAPI.get.mockResolvedValue({ mapTilesKey: expectedKey });
+    });
+
+    it('prepare mapStyle from settings', () => {
+      expect(settingsAPI.get).toBeCalled();
+    });
```
RafaPolit, line 235: I take it this is a Work in Progress.  This tests really don't prove much.  The first one doesn't test anything at all, the second one is testing that the API got called, but nothing else.  Please, don't make this an implementation test, try to test that the componentDiDMount is actually achieving the results you want, and not that the implementation is calling this or other API endpoint.

app/react/Map/specs/Map.spec.js
```diff
@@ -368,4 +383,23 @@ describe('Map', () => {
...
+      render();
+      const style = {
+        sources: { openmaptiles: { url: 'URL?key={{MAP_TILER_KEY}}' } },
+        glyphs: 'URL?key={{MAP_TILER_KEY}}',
+      };
+      const expectedKey = 'XYZ';
+      settingsAPI.get.mockResolvedValue({ mapTilerKey: expectedKey });
+      await instance.replaceKeysMapStyleJson(style);
+      const stringifyStyle = JSON.stringify(instance.mapStyle);
+      expect(stringifyStyle).toEqual(expect.not.stringContaining('{{MAP_TILER_KEY}}'));
```
RafaPolit, line 398: this can be written as:
`expect(stringifyStyle).not.toContain('{{MAP_TILER_KEY}}')`

Please refer to: https://jestjs.io/docs/en/expect#tocontainitem

app/react/Map/specs/Map.spec.js
```diff
@@ -368,4 +383,23 @@ describe('Map', () => {
...
+      expect(instance.mapStyle.getIn(['sources', 'openmaptiles', 'url'])).toEqual(
+        expect.stringContaining(expectedKey)
+      );
+      expect(instance.mapStyle.getIn(['glyphs'])).toEqual(expect.stringContaining(expectedKey));
```
RafaPolit, line 402: These two are the right expects to have!

app/api/settings/settings.js
mfacar: 👍

app/api/settings/specs/settings.spec.js
mfacar: Thank you Rafa, that's true, the initial code covered this case without a test.
I've added a test that saves a mapTilerKey and then checks if get returns this one.

app/react/Map/Map.js
mfacar: mapTilerKeyPlaceholder is perfect 👍

app/react/Map/specs/Map.spec.js
mfacar: I'm sorry, the first test really doesn't have to be there :(
The second test, however, I did this intentionally, because I think I only have to check that the method that contains the logic is called because the replaceKeysMapStyleJson method has already a set of tests in its own context.
I think you are true and the test as it was written doesn't explain anything about the expected behavior, so, I've changed this one for a pretty small check for the expected changes.

app/react/Map/specs/Map.spec.js
mfacar: Great!, thanks for the reference 👍

app/react/Map/specs/Map.spec.js
mfacar: nice 👍

app/api/settings/specs/settings.spec.js
```diff
@@ -190,6 +190,26 @@ describe('settings', () => {
...
+
+    describe('if there is settings with no default mapTilerKey on the DB', () => {
+      it('should return the stored mapTilerKey', done => {
```
RafaPolit, line 200: We have moved away from the (done) + Promises approach, and now use the async-await model.  Please refactor this to use that, you will see how much 'cleaner' and readable it looks.

mfacar: 👍 you're right the code is better now

RafaPolit:
This is working as expected.  My one final doubt is if we should include some sort of explanation as to what this field is in the settings.  I'm not sure the admins will immediately pick what "MapTiler Key" is meant to configure.

What do you think @simonfossom ?

RafaPolit:
I'm merging this, I think we need to review the entire online help system on a different task.  I created this here: #2938

simonfossom:
@RafaPolit I don't know what this is about. If it's documented on mapbox or maptiler we could just provide a link to that part od documentation

PR #2931, opened 2020-05-22
2840 error reporting es data allowed exceeded exception

### LINKED ISSUE #2840: Better error reporting when the amount of data allowed for a field is exceeded
If a user tries to insert in a text field more data than allowed by ES restrictions (: bytes can be at most 32766), Uwazi throws an error simply saying "An error has occurred". These users have no way of knowing why this is happening so they can approach their data loading in a different way.

On the top of that, the error is only triggered when they exceed the request pay load, but not so much during the indexing error:
```
2020-03-25T17:22:46.466Z [localhost] ERROR Failed to index document 5e7b9353df968d20b9c4dbf1: {
 "type": "illegal_argument_exception",
 "reason": "Document contains at least one immense term in field=\"metadata.resumen.value.raw\" (whose UTF8 encoding is longer than the max length 32766), all of which were skipped.  Please correct the analyzer to not produce such terms.  The prefix of the first immense term is: '[72, 85, 82, 73, 68, 79, 67, 83, 32, 115, 101, 101, 107, 105, 110, 103, 32, 83, 111, 102, 116, 119, 97, 114, 101, 32, 68, 101, 118, 101]...', original message: bytes can be at most 32766 in length; got 491349",
 "caused_by": {
  "type": "max_bytes_length_exceeded_exception",
  "reason": "bytes can be at most 32766 in length; got 491349"
 }
}
2020-03-25T17:23:06.597Z [localhost] 
url: /api/entities
PayloadTooLargeError: request entity too large
```

So it is failing silently when the error is on ES indexation, but tossing an error when the error is due to browser payload. 

I haven't checked whether the content does not get updated or it gets truncated.

mfacar:
fixes #2840

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/search/entitiesIndex.js
```diff
@@ -7,7 +7,70 @@ import { entityDefaultDocument } from 'shared/entityDefaultDocument';
...
+    const field = matches[1].split(/\.(?=[^.]+$)/)[0];
+    const fieldPath = field.split('.');
+    truncateLongFieldValue(fieldPath, docData);
+    invalidFields.add(field);
+  });
+  await elastic.bulk({ body, requestTimeout: 40000 });
+  return `max_bytes_length_exceeded_exception. Invalid Fields: ${Array.from(invalidFields).join(
+    ', '
+  )}`;
+};
```
daneryl, line 47: @mfacar is this still needed even with the validation we added for the fields ?

mfacar: I thought the same, with the validation the code becomes unreachable, however, in my tests it turned out that this control throws an exception when the title of the entity is bigger too. 
I think it is better to add the validation at title field and delete this control, what do you think?

daneryl: yes i think the same, this is a lot of code for just a specific scenario, validation solves it. 👍

app/react/utils/api.js
```diff
@@ -20,15 +20,58 @@ const doneLoading = data => {
 
 const isNonUsualApiError = error => error.status && ![401, 404, 409, 500].includes(error.status);
 
+const errorMessages = [
+  {
+    key: /PayloadTooLargeError: request entity too large/g,
+    message: 'The request has too large data. Please review any long value property.',
```
RafaPolit, line 26: Sorry but again this is not correct English. :(

“Request data too large.  Please review text length in property values”

app/react/utils/api.js
```diff
@@ -20,15 +20,58 @@ const doneLoading = data => {
...
+  },
+  {
+    key: /max_bytes_length_exceeded_exception. Invalid Fields: (.+)/g,
+    message: 'The request has too large data. Please review the follow fields: {0} ',
```
RafaPolit, line 30: “Follow” is a verb “to go after someone”.  In this case you want them to review “the following fields”.   So:

“Request data too large. Please review the following fields: “

app/react/utils/api.js
```diff
@@ -20,15 +20,58 @@ const doneLoading = data => {
...
+  },
+  {
+    key: /ERROR Failed to index documents: (.+)/g,
+    message: 'The data have not saved completely, it failed at indexing: {0} ',
```
RafaPolit, line 34: For singular nouns, you use “has”.  So this should read something like: “The data has not been saved in its entirety” or something like that.

But here, what exactly does “completely” mean? I think we should keep as close as possible to the original error.

“Failed to index documents: {0}”

Why the different message?

app/api/search/entitiesIndex.js
mfacar: unreachable code was removed

app/react/utils/api.js
mfacar: this code was removed because it became unreachable after adding the ajv validation

app/react/utils/api.js
mfacar: I was trying to provide more information to the user, but I agree, it was confusing.

app/api/entities/metadataValidators.js
```diff
@@ -73,8 +73,23 @@ const validateGeolocationProperty = value =>
...
+    property.type === propertyTypes.text ||
+    property.type === propertyTypes.markdown
+  ) {
+    const LUCENE_BYTES_LIMIT = 32766;
+    const bytes = Buffer.from(JSON.stringify(value));
+    valid = bytes.length < LUCENE_BYTES_LIMIT;
+  }
+  return valid;
+};
+
```
RafaPolit, line 89: Why do we need the property type check?  Does the Lucene limit not apply to any field? What happens if someone inputs a Numeric field that exceeds this?  I know it's absurd, no one will do that, but I am concerned with the maintainability of this piece of code: if we add other types of fields, we have to always remember and add them here... it's the sort of thing that is prone to development error / forgetting.

Could this not be just a global rule? cc. @daneryl

app/api/entities/specs/entitySchema.spec.js
```diff
@@ -191,13 +202,22 @@ describe('entity schema', () => {
...
 
       describe('markdown property', () => {
         it('should fail if value is not a string', async () => {
           entity.metadata.markdown = [{ value: {} }];
           await expectError(customErrorMessages[propertyTypes.markdown], ".metadata['markdown']");
         });
+        it('should fail if value is a string that exceeds the lucene term byte-length limit', async () => {
+          entity.metadata.markdown = [{ value: largeField }];
+          await expectError('maximum field length exceeded', ".metadata['markdown']");
+        });
```
RafaPolit, line 220: While this, indeed, apparently checks for the text and markdown property, this tests will all pass if the "IF" testing the property type is removed.  This is a case of implementing something that is not really tested beforehand.

Where is the test that defines that that particular "IF" for the property type is required?  It doesn't exist.  So, please, define if it is needed and either test is or remove it.

Thanks.

app/react/utils/specs/api.spec.js
```diff
@@ -150,6 +172,21 @@ describe('api', () => {
...
 
+    describe('when request return a server error of too large request', () => {
+      it('should notify the error with this message', async () => {
+        const requestParams = new RequestParams({ key: 'test' }, { header: 'value' });
+        try {
+          await api.post('validation_error', requestParams);
+          fail('should throw error');
+        } catch (e) {
+          testNotificationDisplayed(
+            "validation failed:  .metadata['prop1'] should be string, .metadata['prop2'] is too longer,",
```
RafaPolit, line 183: Is there no way to differentiate that it, indeed, was a string, but it is too large?  This may mislead into interpreting it was not a string.

app/api/entities/metadataValidators.js
mfacar: I think that for other fields its own validations of format or type prevent to admit long fields and in that case, we could avoid this processing block

app/api/entities/specs/entitySchema.spec.js
mfacar: you're right @RafaPolit, this condition is not necessary by tests neither the running app

app/api/entities/metadataValidators.js
daneryl: looks good now 👍

app/react/utils/specs/api.spec.js
```diff
@@ -150,6 +172,21 @@ describe('api', () => {
       });
     });
 
+    describe('when request return a server error of too large request', () => {
+      it('should notify the error with this message', async () => {
+        const requestParams = new RequestParams({ key: 'test' }, { header: 'value' });
+        try {
+          await api.post('validation_error', requestParams);
```
RafaPolit, line 179: Is this a self contained thing within the test?  You are making an API call to "validation_error", which is overridden on line 58. I am confused.

app/react/utils/specs/api.spec.js
```diff
@@ -14,6 +14,23 @@ import loadingBar from 'app/App/LoadingProgressBar';
...
+    body: {
+      error: 'validation failed',
+      validations: [
+        {
+          dataPath: ".metadata['prop1']",
+          message: 'should be string',
+        },
+        {
+          dataPath: ".metadata['prop2']",
+          message: 'is too longer',
```
RafaPolit, line 28: Please fix this English error: "is too long".  "Longer" is a comparison operator.

app/api/search/specs/index.spec.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
```
RafaPolit, line 1: Please, remove this line, file no longer exceeds max-line limit.

Please, don't use this exception, we have talked about it, and we still want to know this file could be improved by refactoring, or other means to reduce max lines.

PR #2941, opened 2020-05-29
2927 prevent delete connections

### LINKED ISSUE #2927: Changing an entit

mfacar:
fixes #2927

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


RafaPolit:
There are conditions where things are not working as expected:

## How to reproduce

- Create templates A, B and C
- Create any relationship type, in this case "related"
- Create a relationship property in template A for type "related" and Template C
- Create an Entity "eA" in template A and and entity "eB" in template B
- Manually create a relationship between "eA" (with no label) to "eB" with type related
- Change the template of "eB" to "C"

Expected behavior: The relationship of eA to eB now fulfills the metadata property's conditions, so, therefore, the metadata property should get filled with eB.

Current behavior: No metadata value is created until a further edit of eA triggers the parsing of relationships to assign metadata properties.

The fix for this is not straightforward.  Perhaps @konzz can give an insight into the best way to achieve this?

RafaPolit:
There is another condition where this new approach fails:

## How to reproduce

- Create templates A, B, and C
- Create any relationship type, in this case "related"
- Create a relationship property in template A for type "related" and Template C
- Create an Entity "eB" in template B and an entity "eC" in template C
- Manually create a relationship between "eB" (with no label) to "eC" with type related
- Go to edit "eB" and change its template to A

Here's a tricky scenario.  Since there is a current relationship created between this entiy to eC, this meets the criteria for a metadata relationship pointing to eC.  But the relationship checkboxes are left empty, and, if you click save, you will even lose the original relationship (it now gets deleted).

This is more tricky, the user actually sees that the appropriate checkbox is not selected.  But we should actually, on template change, check for the existing relationships on the entity and populate the relationship checkboxes accordingly.

Again, @konzz may be in a better position to help with this.

RafaPolit:
Since we are introducing these new conditions by preventing deleting the connections this is a tricky decision.  Our current "delete all connections on template change" at least produces expected results.  If we just merge this "no delete" fix and not address the other two scenarios, we could end up with people having ghost relationships that don't show on metadata or relationships deleted upon template change on this second scenario.

I think we should address this all together before merging this.

What do you think @txau ?

txau:
@RafaPolit excellent analysis, ty. You are right, we should add the logic for properly updating connections on template changes. This could explain other problems we have had in the past with ghost relationships etc.

PR #2949, opened 2020-06-03
Fix for Map inherited Geolocations

### LINKED ISSUE #2948: Inherited Geolocation properties not working correctly in map
Hovering over a map's single entity (not a cluster) results in a 'cannot get property GET of undefined' error when the source of the map is an inherited Geolocation.

Particularly, this piece of code in `/app/react/Library/components/MapView.js`:
```
{marker.properties.inherited &&
  markerInfo(
    <EntityTitle
      context={marker.properties.context}
      entity={marker.properties.inheritedEntity}
    />
  )}
```

Here, `properties.inheritedEntity` is Undefined.

The error then moves into the EntityTitle which is not prepared to get an undefined Entity.

RafaPolit:
fixes #2948 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


RafaPolit:
To QA this, it is important to test single dots and cluster in maps, both coming from direct Geolocation properties and also coming from inherited Geolocations.  This later was the condition that started the issue.

app/api/search/search.js
```diff
@@ -416,10 +416,12 @@ const getInheritedEntitiesData = async (toFetchByTemplate, language, user) =>
       }
       return entities.get(query, {
         ...toFetchByTemplate[t].properties.reduce(
-          (memo, n) => Object.assign(memo, { [`metadata.${n}.value`]: 1 }),
+          (memo, n) =>
+            Object.assign(memo, { [`metadata.${n}.value`]: 1, [`metadata.${n}.label`]: 1 }),
           {}
         ),
         sharedId: 1,
+        title: 1,
```
konzz, line 424: is it possible to use "true" instead of "1" ?

RafaPolit: According to the documentation, yes, projections can be 1 or true.
https://docs.mongodb.com/v4.0/reference/method/db.collection.find/#projection

I do believe its more semantic.

PR #2957, opened 2020-06-08
2855 toc not visible on library view

### LINKED ISSUE #2855: ToC not visible on library view
ToC has it's own tab on the side-panel view of the library for each document.

Currently, it shows as "No table of Content" (which by the way is wrongly worded, it should read "No Table of Contents" - in plural) even if there is a ToC created for the document of that entity.

Originally reported in #2854

mfacar:
fixes #2855

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Documents/components/DocumentSidePanel.js
```diff
@@ -39,6 +40,20 @@ export class DocumentSidePanel extends Component {
     }
   }
 
+  setDefaultDocumentToC(isEntity, documents, language) {
+    let defaultDocumentToC = this.props.file.toc;
+
+    if (!isEntity) {
+      const defaultDocument = {
+        ...entityDefaultDocument(documents, language, 'en'),
```
RafaPolit, line 48: This is not correct: "en" is not the default language.  The default language is collection-dependent and it comes from the DB from settings -> languages -> the one where default === true.

It is probably already in the store somewhere, as a lot of things, including the language you are redirected to upon landing on any page, depend on this.

app/react/Documents/components/DocumentSidePanel.js
```diff
@@ -387,11 +403,13 @@ export const mapStateToProps = (state, ownProps) => {
   const references = ownProps.references
     ? viewerModule.selectors.parseReferences(ownProps.doc, ownProps.references)
     : relevantReferences;
+  const defaultLanguage = state.locale;
```
RafaPolit, line 406: This naming is confusing.   This is not the "defaultLanguage", this is the "currentLanguage".

app/react/Documents/components/DocumentSidePanel.js
```diff
@@ -387,11 +403,13 @@ export const mapStateToProps = (state, ownProps) => {
...
   return {
     references,
     excludeConnectionsTab: Boolean(ownProps.references),
     connectionsGroups: state.relationships.list.connectionsGroups,
     relationships: ownProps.references,
+    defaultLanguage,
```
RafaPolit, line 412: Where is this prop being used in the component?

app/api/i18n/systemKeys.js
```diff
@@ -142,10 +142,10 @@ export default [
   { key: 'No Relationships description', label: 'Relationships are bonds between entities.' },
   { key: 'No type', label: 'No type' },
   { key: 'No type', label: 'No type' },
-  { key: 'No Table of Content', label: 'No Table of Content' },
+  { key: 'No Table of Contents', label: 'No Table of Contents' },
```
RafaPolit, line 145: Good catch on the English error here!

app/react/Documents/components/specs/DocumentSidePanel.spec.js
```diff
@@ -108,6 +109,27 @@ describe('DocumentSidePanel', () => {
         ).toBe('metadata');
       });
     });
+
+    describe('when doc passed has toc', () => {
+      const toc = ['title1', 'title2', 'title3'];
+
+      it('if is a new entity should set toc of the loaded file', () => {
```
RafaPolit, line 116: Please, start all "it" statements with "should".

The way tests are meant to be read is:  `Description`: it `should... `.

In this case, if you concatenate the wordings:
> When doc passed has toc, it if is a new entity...

This is not correct.  What we want is:
> When doc passed has toc, it should set TOC of the loaded file (if doc is a new entity)

So, the 'it' should read: "should set the TOC of the loaded file (if doc is a new entity).

Ideally, "if"s are actually extra describes:
```
describe('when doc passed has ToC')
  describe('when doc is a new entity')
     it('should set ...')
```

This is the "preferred" approach, but sometimes that means nesting too much.  So, you decide here which approach you prefer, but every "it" should start with "should", no matter what.

app/react/Documents/components/specs/DocumentSidePanel.spec.js
```diff
@@ -108,6 +109,27 @@ describe('DocumentSidePanel', () => {
...
+        props.doc = Immutable.fromJS({ metadata: [], attachments: [], type: 'entity' });
+        props.file = { toc, pdfInfo: {} };
+        props.tab = 'toc';
+        render();
+        expect(component.find({ toc })).toHaveLength(1);
```
RafaPolit, line 121: The test says that "it should set the ToC of the loaded file"... yet the actual expect here only tests for length.  I think that the description is correct, but the tests needs to proof it is getting the "correct" ToC.

app/react/Documents/components/specs/DocumentSidePanel.spec.js
```diff
@@ -108,6 +109,27 @@ describe('DocumentSidePanel', () => {
...
+      });
+
+      it('if is a loaded entity should set the toc of the default document', () => {
+        const documents = [{ toc }];
+        entityDefaultDocument.entityDefaultDocument = jest.fn().mockReturnValue(documents[0]);
+        props.doc = Immutable.fromJS({ documents, attachments: [], type: 'entity' });
+        props.tab = 'toc';
+        render();
+        expect(component.find({ toc })).toHaveLength(1);
```
RafaPolit, line 130: Same here: please test not the length, but the actual expected default ToC.

app/react/Documents/components/specs/DocumentSidePanel.spec.js
```diff
@@ -108,6 +109,27 @@ describe('DocumentSidePanel', () => {
...
+
+      it('if is a new entity should set toc of the loaded file', () => {
+        props.doc = Immutable.fromJS({ metadata: [], attachments: [], type: 'entity' });
+        props.file = { toc, pdfInfo: {} };
+        props.tab = 'toc';
+        render();
+        expect(component.find({ toc })).toHaveLength(1);
+      });
+
+      it('if is a loaded entity should set the toc of the default document', () => {
```
RafaPolit, line 124: Please, apply the same principle as before: either nest the "if" in a new describe, or start with "should" and move the conditional to the end of the pharse.

app/react/Documents/components/specs/DocumentSidePanel.spec.js
```diff
@@ -185,5 +208,10 @@ describe('DocumentSidePanel', () => {
       const ownProps = {};
       expect(mapStateToProps(state, ownProps).connectionsGroups).toBe('connectionsGroups');
     });
+
+    it('should map default language', () => {
+      const ownProps = {};
+      expect(mapStateToProps(state, ownProps).defaultLanguage).toBe('es');
```
RafaPolit, line 214: As mentioned before, this is:
- the wrong wording, this is currentLanguage, not defaultLanguage
- I still can't see where this prop is actually used

RafaPolit:
I have added some things that need revisiting.  In the process, please also update the branch with the most current DEV commit.

Thanks.

app/react/Documents/components/DocumentSidePanel.js
RafaPolit: state.locale holds the 'current language' that the person has either selected or has been assigned automatically.  The entityDefaultLanguage expects three arguments: documents, entityLanguage and defaultLanguage.  EntityLanguage.

Since we have one entity per UI language, entityLanguage and stat.locacle are 100% of the time going to be the same value.  What it is expecting in the 3rd argument is the COLLECTION's default so that it can, if there is no match to the entity's language, try to pick the collection's default.

So you need to fetch it from state.collection (I'm almost sure this is always present) and, from languages, extract the one that has default: true.  That should be done in the mapStateToProps and you pass the language so that we prevent unnecessary rerenders.

app/api/files/files.ts
```diff
@@ -9,7 +9,7 @@ import { FileType } from '../../shared/types/fileType';
 export const files = {
   async save(file: FileType) {
     const savedFile = await model.save(await validateFile(file));
-    await search.indexEntities({ sharedId: file.entity }, '+fullText');
+    await search.indexEntities({ sharedId: file.entity || savedFile.entity }, '+fullText');
```
mfacar, line 12: Additionally to the visualization problem, it was missing to index the entities of the updated files. I've added a test to ensure this indexation.

fnocetti: @mfacar is `file.entity` ever used?

app/react/Documents/components/DocumentSidePanel.js
```diff
@@ -39,6 +40,20 @@ export class DocumentSidePanel extends Component {
     }
   }
 
+  setDefaultDocumentToC(isEntity, documents, language, currentLanguage) {
```
fnocetti, line 43: @mfacar this is a really minor thing but, can we change the name of this function to start with `get` instead of `set`?

app/api/files/files.ts
mfacar: Given the saved entity will always to have the entity value, I've removed file.entity

PR #2958, opened 2020-06-08
Return a 404 when a non existing file is requested

### LINKED ISSUE #2901: Undefined defaultDoc on viewer -> routeActions
We are receiving several error notifications on different instances with this message:
```
TypeError: Cannot read property '_id' of undefined
    at requestViewerState ([...]/app/react/Viewer/actions/routeActions.js:41:77)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

This means that the Viewer's `requestViewerState` procedure is returning `undefined` on the `defaultDoc` key for the `getDocument` process near line 30.

This could be the result of navigating to the URL of an Entity without document, but given the amount of errors we are getting, this does not seem to be the case.  We should look for what it triggering this, but we should also probably tackle the scenario of what to do if this happens.  Theoretically it shouldn't happen as that route should only be called for entities which already have a main document.

elric-wamugu:
fixes #2901

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Viewer/actions/documentActions.js
```diff
@@ -92,12 +92,11 @@ export function deleteDocument(doc) {
 
 export async function getDocument(requestParams, defaultLanguage, filename) {
   const [entity] = (await api.get('entities', requestParams)).json.rows;
-
   const defaultDoc = filename
     ? entity.documents.find(d => d.filename === filename)
     : entityDefaultDocument(entity.documents, entity.language, defaultLanguage);
 
-  entity.defaultDoc = defaultDoc;
+  entity.defaultDoc = defaultDoc || {};
```
RafaPolit, line 99: Please create a test for this.

daneryl: please @eric create a test for the fix.

daneryl:
@eric fix codeclimate issue please. ill merge after both issues have been solved 👍

app/react/Viewer/actions/specs/documentActions.spec.js
```diff
@@ -246,7 +248,8 @@ describe('documentActions', () => {
     describe('getDocument', () => {
       it('should return the document requested', async () => {
         const requestParams = new RequestParams({ sharedId: 'docWithPDFRdy' });
-        const doc = await actions.getDocument(requestParams);
+        const doc = await actions.getDocument(requestParams, 'en', 'filename');
+        expect(doc.documents[0]).toBeDefined();
```
elric-wamugu, line 252: @joan, I added this assert to check for the if the doc object is defined or not. Let me know if that's an enough test to check for that

daneryl: @eric i do not see how this is related to the default {} for defaultDoc, actually, if you remove the ```|| {}``` from https://github.com/huridocs/uwazi/blob/2e99e706791f1d370ae2cd0cdb4bccdc81ca2d69/app/react/Viewer/actions/documentActions.js#L98-L100

test is still green, meaning that it is not properly tested.

elric-wamugu: okay, thank you. On it

konzz:
@elric-wamugu is this PR ready to be QA again?

elric-wamugu:
> @elric-wamugu is this PR ready to be QA again?

@konzz yes it is

PR #2973, opened 2020-06-11
2964 contact form

### LINKED ISSUE #2964: Contact form validation is failing
email-validation

txau:
fixes #2964 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/contact/specs/routes.spec.js
```diff
@@ -1,37 +1,54 @@
...
+    describe('validation', () => {
+      it('should not validate with missing requried properties properties', async () => {
+        body = {};
+        const response = await request(app)
+          .post('/api/contact')
+          .send(body);
+        expect(response.text).toContain('validation failed');
+      });
+
+      it('e-mail should be an e-mail', async () => {
```
RafaPolit, line 45: As convention, `it` statements should start with `should`.  The test parser actually composes this as "Validation: it ...", so this should probably read: "should validate email as such".

app/api/contact/specs/routes.spec.js
```diff
@@ -1,37 +1,54 @@
...
+        body.email = 'wrong format';
+        const response = await request(app)
+          .post('/api/contact')
+          .send(body);
+        expect(response.text).toContain('validation failed');
```
RafaPolit, line 50: This is probably the wrong thing to expect.  Since you are missing the name and message, this test will pass even if you include no mail at all.  Or even if you put in a correct mail in the body!

So, there should be a specific failure, on top of the validation failed, that proves that email is actually being rejected.  For such a scenario, it's probably a good practice to include the other fields passing correctly and just expect the specific mail error you are expecting from AJV.

app/react/Markdown/components/specs/ContactForm.spec.js
```diff
@@ -23,13 +23,11 @@ describe('ContactForm', () => {
...
-      done();
-    };
+
     const component = shallow(<ContactForm {...props} />);
     component.instance().setState(formData);
-    const form = component.find('form');
-    form.simulate('submit', { preventDefault: () => {} });
+    component.find('form').simulate('submit', { preventDefault: () => {} });
+    expect(api.post).toHaveBeenCalledWith('contact', { data: formData, headers: {} });
+    done();
```
RafaPolit, line 31: On the previous implementation, this test actually tested, indirectly, that notify was being called.  I agree it may be somewhat weird.

With the new implementation though, there is nothing asynchronous in this function, so a `done()` in line with everything else has no effect.  This can be re-written to either include the async nature of the api call, or just remove the done(); from here and the argument done in line 20.

app/react/utils/api.js
```diff
@@ -30,7 +30,11 @@ const handleErrorStatus = error => {
   } else if (error.status === 500) {
     store.dispatch(notify('An error has occurred', 'danger'));
   } else if (isNonUsualApiError(error)) {
-    store.dispatch(notify(error.json.error, 'danger'));
+    if (error.json.prettyMessage) {
+      store.dispatch(notify(error.json.prettyMessage, 'danger'));
+    } else {
+      store.dispatch(notify(error.json.error, 'danger'));
+    }
```
RafaPolit, line 37: perhaps a good refactor here is the much simpler:
```
store.dispatch(notify(error.json.prettyMessage || error.json.error, 'danger'));
```

app/react/Markdown/components/ContactForm.js
```diff
@@ -22,12 +21,16 @@ export class ContactForm extends Component {
...
-    api.post('contact', this.state).then(() => {
+    try {
+      await api.post('contact', new RequestParams(this.state));
       this.props.notify('Message sent', 'success');
       this.setState({ name: '', email: '', message: '' });
-    });
+    } catch (err) {
+      return err;
+    }
+    return true;
```
RafaPolit, line 33: Not sure what returning the err does on a submit, or the convention to return `true`.  Is this a usual HTML convention?

At any rate, neither the success return nor the error cases are tested.

app/api/contact/specs/routes.spec.js
txau: Fixed

app/api/contact/specs/routes.spec.js
txau: Actually is being tested as you suggest. body is set on beforeEach with correct params. I'm only overwriting e-mail with a wrong param.

app/react/Markdown/components/specs/ContactForm.spec.js
txau: I opted for removing the done

app/react/utils/api.js
txau: done

app/react/Markdown/components/ContactForm.js
txau: There was no need for these returns. I simplified them and let the middlewares handle the errors.

txau:
@RafaPolit I just detected that I screwed up some CSS. I'm switching to draft mode while I fix them.

app/api/contact/specs/routes.spec.js
daneryl: @txau but the test should expect that the error is an email error instead of a generic ajv Error i think that is what @RafaPolit is pointing out.

we already have this scenarios when testing entity/template validators and the tests were actually green when the error had nothing to do with the specific test but for another unexpected scenario, its true that is a much simpler scenario...

txau: @daneryl still it will be coupled to how AJV reports the errors.

daneryl: @txau im ok in this case but at some point i think it is more important to test what is actually being validated and be coupled to whatever you are using as a validator, we are coupled already anyways.

app/react/Notifications/components/Notification.js
```diff
@@ -21,7 +21,14 @@ export class Notification extends Component {
     return (
       <div className={cssClass} onClick={this.close.bind(this)}>
         <Icon icon={icon} />
-        <span>{message}</span>
+        <span className="alert-text">
+          {message.split('\n').map((item, i) => (
+            <React.Fragment key={i}>
```
daneryl, line 26: @txau supress this error or use uniqueID form shared/uniqueID.js

txau: Supressed

PR #2983, opened 2020-06-16
Relationship parser now takes into account content

### LINKED ISSUE #2981: CSV Import of relationship fields is not taking into account the property.content
CSV Import of relationship fields is fetching entities with the passed Title form within ALL the entities in the DB, not just those of the particular template you are requesting for.

If there are matches on entities belonging to other templates, it will STILL assign that value to a relationship property that, in theory, could not hold that particular ID.  This causes all sort of weird behaviors.

On the other hand, missing values are created in the property.content template.  If a property has been configured as "any", this is REALLY prone to errors.  So we need to avoid the creation of entities if a relationship field is pointing to ANY, as there is no way of knowing in which template that entity should go into.

RafaPolit:
fixes #2981

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/csv/typeParsers/relationship.ts
+++ b/app/api/csv/typeParsers/relationship.ts
@@ -33,7 +33,7 @@ const relationship = async (entityToImport: RawEntity, property: PropertySchema)
           template: property.content,
         },
         {
-          language: 'en',
+          language: entityToImport.language,
           user: {},
         }
       );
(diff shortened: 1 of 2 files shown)


app/api/csv/typeParsers/relationship.ts
```diff
@@ -4,30 +4,43 @@ import { RawEntity } from 'api/csv/entityRow';
 import { ensure } from 'shared/tsUtils';
 import { PropertySchema } from 'shared/types/commonTypes';
 
+interface Query {
+  title: { $in: string[] };
+  template?: string;
+}
+
```
daneryl, line 11: i do not think you really need an interface here? can we just use an object to not redefine what exists? if it is really needed you can use FilterQuery from @types/mongoose

PR #2984, opened 2020-06-16
HOTFIX CSV Export: error when entity without metadata

fnocetti:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/csv/csvExporter.ts
+++ b/app/api/csv/csvExporter.ts
@@ -7,6 +7,7 @@ import { TemplateSchema } from 'shared/types/templateType';
 import translate, { getLocaleTranslation, getContext } from 'shared/translate';
 import translations from 'api/i18n/translations';
 import formatters from './typeFormatters';
+import { EntitySchema } from '../../shared/types/entityType';
 
 export type SearchResults = {
   rows: any[];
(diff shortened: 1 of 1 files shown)


app/api/csv/csvExporter.ts
```diff
@@ -176,7 +176,7 @@ export const processEntity = (
       return processCommonField(header.name, row, rowTemplate, options);
     }
 
-    if (!row.metadata[header.name]) {
+    if (!row.metadata || !row.metadata[header.name]) {
```
daneryl, line 179: i think ```if (!row?.metadata[header.name])``` should work here instead of the double check.

this should have been detected during development if row was not typed as "any", correct me if im wrong, should row be an entity? or is there a reason for this to be "any" ? if there is none i think we should include that fix also on this PR.

fnocetti: @daneryl you're right. I'll use the `?.` operator.

I agree with you in that this would have been detected during development if I used proper typings. Row is an entity from the search results. The reason why I used any is because I couldn't find typings for our search results.
Now I'm revisiting it probably `EntitySchema` is the type I was looking for.
I'll take a look.

daneryl: @fnocetti create an issue to type the row and lets merge this with the if changed as it is kind of urgent.

PR #2985, opened 2020-06-17
2967 sanitize values on template changes

### LINKED ISSUE #2967: Changing a select property's thesauri target is not sanitizing the data
If you change the thesauri that a select property is pointing to in templates, those entities that already have values are not being sanitized.

daneryl:
fixes #2967 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti:
LGTM.

As for the scenarios described in the issue:

> If you change the thesauri that a select property is pointing to in templates, those entities that already have values are not being sanitized.

When I tried to reproduce this previous to this patch the application was throwing an error failing to index the changes on entities which is what @daneryl stated on his comment. This is fixed now.

> Deleting a property is also not deleting the data, either on the database or on the ES index, not sure.

I could not reproduce this scenario previous to this patch and it is working correctly after the patch. @RafaPolit if you want we can take a look together and see if we can make it happen. If not I'll merge it.

fnocetti:
Did the QA again and LGTM, I'm merging this.

I like the fact that this small change makes the code cleaner :)

PR #2991, opened 2020-06-18
testing CI job separation

txau:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


txau:
@daneryl please take a look to the strategy for caching the production build. The idea is to save up time in subsequent runs of the build if:
- No uwazi source files have changed
- No dependency modules have changed
- No database scripts have changed

Arguably, this should also include other checks for the cache, so ie. if the build script itself has changed. So maybe we should monitor webpack and babel configurations as part of the caching check?

txau:
@RafaPolit regarding the vast amount of issues from code climate, it is coming from switching thresholds for code similarity back and forth. So the number is unreal. I suggest that we pass it as it is?

RafaPolit:
> @RafaPolit regarding the vast amount of issues from code climate, it is coming from switching thresholds for code similarity back and forth. So the number is unreal. I suggest that we pass it as it is?

LOL! How did you manage to create 1119 issues? I cannot even see them :)  I will check, but that is not a good sign.

txau:
@RafaPolit all these issues already existed. Because the code similarity threshold was changed from 70 to 128 (so I removed hundreds of issues), then I changed in back to 90, so I "recreated" 1119 issues. But actually they are the same issues that were already there.

daneryl:
this looks great, we have gained like 8 minutes ? 😀

> Arguably, this should also include other checks for the cache, so ie. if the build script itself has changed. So maybe we should monitor webpack and babel configurations as part of the caching check?

i think it makes sense to include those files as part of the cache key, yes.

konzz:
So this will help when we re-run a branch without modifying it, or are other scenarios I'm not thinking of?

daneryl:
> So this will help when we re-run a branch without modifying it, or are other scenarios I'm not thinking of?

apart from the cache there is the e2e split into to runs, that should help to improve execution times, but now that you mention i see that these times are from a cached production version, and normally this will run uncached, @txau what are the times for the uncached version ? it is still worth it ?

txau:
> apart from the cache, there is the e2e split into two runs, that should help to improve execution times, but now that you mention i see that the current times (8 min per run) are from a cached production version, and normally this will run uncached, @txau what are the times for the uncached version ? it is still worth it ?

Yeah, still worth it. This whole strategy can be improved, so ie. Leaving the test files out of the build key so once fixing/developing tests you still get this benefit. So this is an initial step.

txau:
> > @RafaPolit regarding the vast amount of issues from code climate, it is coming from switching thresholds for code similarity back and forth. So the number is unreal. I suggest that we pass it as it is?
> 
> LOL! How did you manage to create 1119 issues? I cannot even see them :) I will check, but that is not a good sign.

Actually there was a issue, I had missconfigured the threshold for codeclimate's similarity check. Now is ok and we are down to the 27 original issues.

txau:
@daneryl in this branch:
- Split e2e tests into 2 suites to gain some time when running the tests
- Fix a failing e2e test for customHomePage
- Fixed several linting errors
- Upgraded cookie modules
- Added cache to build in the CI process

daneryl:
LGTM

PR #2995, opened 2020-06-19
2993 no stop on errors

### LINKED ISSUE #2993: EntitiesIndex now does not continue upon error, making reindex impossible
https://github.com/huridocs/uwazi/blob/026038e9577732868d3a0d4d5362652823b91c1f/app/api/search/entitiesIndex.js#L68-L77

With the introduction of this new code in #2931 we solved a problem that failing to reindex a single entity will not report back to client.

The side effect is that now, a full collection reindex will trigger the error and stop the process.  This causes library to show only X amount of documents until the reindex process stopped and nothing else.

This is also preventing our deploy procedure to actually deploy even past the first instance.

@daneryl @konzz @fnocetti @mfacar Let's discuss what is the expected behavior for all scenarios and implement that.  If we need to fix our deployment process, that is OK, but let's make sure these are good enough reasons.

RafaPolit:
fixes #2993

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/search/entitiesIndex.js
+++ b/app/api/search/entitiesIndex.js
@@ -73,11 +73,10 @@ const bulkIndex = async (docs, _action = 'index', elasticIndex) => {
 };
 
 const newIndexErrorsOrThrow = err => {
-  if (!(err instanceof IndexError)) {
-    throw err;
-  } else {
+  if (err instanceof IndexError) {
     return err.errors;
   }
+  throw err;
 };
 
 const appendRelationships = async entity => {
--- a/app/api/search/entitiesIndex.js
+++ b/app/api/search/entitiesIndex.js
@@ -80,46 +80,65 @@ const newIndexErrorsOrThrow = err => {
   }
 };
 
-const indexEntities = async (
+const appendRelationships = async entity => {
+  const relations = await relationships.get({ entity: entity.sharedId });
+  return { ...entity, relationships: relations || [] };
+};
+
+const bulkIndexAndCallback = async assets => {
+  const { searchInstance, entitiesToIndex, elasticIndex, batchCallback, totalRows } = assets;
+  await searchInstance.bulkIndex(entitiesToIndex, 'index', elasticIndex);
+  return batchCallback(entitiesToIndex.length, totalRows);
+};
+
+const indexBatch = async (offset, totalRows, options, errors = []) => {
+  const { query, select, limit, batchCallback, elasticIndex, searchInstance } = options;
+  if (offset >= totalRows) {
+    return errors.length ? handleErrors(errors) : Promise.resolve();
+  }
+
+  const entitiesToIndex = await entities.get(query, '', {
+    skip: offset,
+    limit,
+    documentsFullText: select && select.includes('+fullText'),
+  });
+
+  const entitiesToIndexWithRels = await Promise.all(entitiesToIndex.map(appendRelationships));
+
+  let newIndexErrors = [];
+
+  try {
+    await bulkIndexAndCallback({
+      searchInstance,
+      entitiesToIndex: entitiesToIndexWithRels,
+      elasticIndex,
+      batchCallback,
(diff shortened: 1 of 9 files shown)


RafaPolit:
## Important information

- handleErrors in entitiesIndex.js does not create the errorLog, this logic needs to be implemented upstream if required
- bulkIndex in entitiesIndex.js, which is only ever called from within this very file (and implemented in search.js just for testing purposes! :( but no one other than tests use this function) will always throw on error, but it will not transform a potentially non-index error into one as was the case with previous implementation
- indexEntities in entitiesIndex.js can now take a continueOnIndexError option.  This function now  will return an object of the type `{ errors: []}`.  If continueOnIndexError is true, it will populate that array with all the errors found during the batch processes.  If false, it will just throw the error and stop.
- I have implemented a test scenario of using the continueOnIndexError in the reindex_elastic procedure, and it works as expected.  It continues with the process and, after finishing, it reports all the combined index errors that this process may have found.  If it finds non-indexing errors, it will stop as expected.

## ToDo

- We should look into which processes that use this function require to actually stop on error.  Since, at the very least, everything is passing one entity per language, I think that it is probable that we need this on EVERY call, in which case we should change the default from false to true.
- We need to implement on the upstream services that use indexEntities how to deal when errors.length > 0.

@daneryl I have flagged this as draft, but please take a look so we can further discuss the following steps.  Thanks.

daneryl:
> indexEntities in entitiesIndex.js can now take a continueOnIndexError option. This function now will return an object of the type { errors: []}. If continueOnIndexError is true, it will populate that array with all the errors found during the batch processes. If false, it will just throw the error and stop.

> We should look into which processes that use this function require to actually stop on error. Since, at the very least, everything is passing one entity per language, I think that it is probable that we need this on EVERY call, in which case we should change the default from false to true.

@RafaPolit im not understanding why we need this, why should we need to stop, is not this fix to actually do the opposite ? can you give an example of a process that actually needs to stop on an indexation error ?

i think the following flow works:

- always let it finish on indexation errors. (capture all errors and throw them at the end)
- throw any other error that is not an indexation error.


> We need to implement on the upstream services that use indexEntities how to deal when errors.length > 0.

in case there is indexation error i think we want the normal error handling flow, this will end up showing an error/s to the user, this should be already covered by our error handling middleware, maybe we need to adjust it a little to handle this new type of IndexError ? pls check.

RafaPolit:
This assumption came from the previous development.  I was myself questioning if this is still needed.  Since we are always at least indexing one per language, the only scenario we would like to stop is on single language instances on entity creation / editing. 

I agree that this should be a default true or even remove that flag altogether.

RafaPolit:
Logic has been simplified:
- BulkIndex always throws on error
- indexEntities now catches BulkIndex errors and stores the different errors and continues till the end batch processing
- once finished, if there were errors, it throws the error
- it is up to the upstream process to handle the error, most of them should be automatically handled.

Notable exceptions:
- reindex was not prepared to catch this error. It now goes through the entire process and will display errors on failure.  This fix is implemented in this PR
- CSV import: csv import does entities one at a time.  While it relies on indexEntities, it does this one at a time.  Currently, a single point failure will stop the process.  I have opened a new issue for discussing this: #3000

app/api/search/entitiesIndex.js
```diff
@@ -38,6 +35,7 @@ function setFullTextSettings(defaultDocument, id, body, doc) {
 
 const bulkIndex = async (docs, _action = 'index', elasticIndex) => {
   const body = [];
+  // eslint-disable-next-line max-statements
```
RafaPolit, line 38: I have not changed the following .forEach nor appended any code to it.  We should probably refactor this.

daneryl:
LGTM

PR #3015, opened 2020-06-25
Exit migration on uncaught error with 1

RafaPolit:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


daneryl:
@RafaPolit 

this was failing due to calling closeGrayLog function without a callback, i have changed the closeGrayLog function to have a default empty callback.

regarding error handling i have change this to use 
```
process.on('unhandledRejection', error => {
  throw error;
});
```

this ends the script with code 1 and prints the error as node does it by default, error message + stacktrace, this works well and its simple, if you think this works well i change it also on the reindex script and i think we can close #3020 ?

PR #3023, opened 2020-07-01
2968 can assign same property different thesauri

### LINKED ISSUE #2968: You can assign the same property name pointing to different thesauri in different templates
## To reproduce

- Have template A with property "country" pointing to thesauri "countries"
- Have template B with property "country" pointing to thesauri "cases"

This is a weird scenario: the property has a consistent type, but by pointing to different thesauri, the condition opens a whole can of worms.  Specially if the thesauri are nested and one of the properties is selected as "default filter".

This should not be allowed.  If the same name is used with different targets, this should prevent saving.

### LINKED ISSUE #2971: You should not be able to assign the same name and different types to template properties
Until we have typing of properties implemented, we need to block users from assigning different types to same-name properties, as this prevents ES from indexing correctly one type of templates.

mfacar:
fixes #2968 
fixes #2971 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/shared/types/templateSchema.ts
```diff
@@ -155,12 +155,70 @@ ajv.addKeyword('cantDeleteInheritedProperties', {
...
+            { content: { $ne: property.content } },
+            { type: { $ne: property.type } },
+            { relationtype: { $ne: property.relationType } },
+          ],
+        },
+      ],
+    }));
+    const query = { properties: { $elemMatch: { $or: [...condition] } } };
+    const matchedTemplates = await model.get(query);
+    const otherTemplates = matchedTemplates.filter(t => !t._id.equals(template._id));
```
daneryl, line 181: @mfacar excluding the current template can probably be done on the query.

app/shared/types/templateSchema.ts
```diff
@@ -155,12 +155,70 @@ ajv.addKeyword('cantDeleteInheritedProperties', {
...
+      }, []);
+
+      throw new Ajv.ValidationError(
+        errorProperties.map(property => ({
+          keyword: 'cantReuseNameWithDifferentType',
+          schemaPath: '',
+          params: { keyword: 'cantReuseNameWithDifferentType' },
+          message:
+            'Entered label is already in use on another property with a different type or thesaurus',
+          dataPath: `.properties.${property}`,
```
daneryl, line 206: @mfacar,  Ajv error can also have data property, i suggest you include the data that is wrong, this way there will be better information on why the error happens. you can include the properties that do not pass the validation from the current template that is being validated.

app/shared/types/templateSchema.ts
```diff
@@ -155,12 +155,70 @@ ajv.addKeyword('cantDeleteInheritedProperties', {
...
+              (p.content !== property.content ||
+                p.type !== property.type ||
+                p.relationtype !== property.relationType)
+          );
+          if (matches && !propertyNames.includes(property?.name || '')) {
+            propertyNames.push(property?.name || '');
+          }
+        });
+        return propertyNames;
+      }, []);
```
daneryl, line 197: @mfacar have you considered solving this by concat every template property into one array ? or even on a map indexed by type+content,  this way there will be 2 nested iterations instead of 3 and i think it will result in a more readable implementation by default.

fnocetti:
> @mfacar @RafaPolit apart from the code review.
> probably current instances already have the cases that we are not allowing on this validation ? did you guys discussed how to solve this or it just self solves by the users going to the wrongly formed templates and re saving them solving the errors ?
> another option i guess is a migration, but we can not decide arbitrarily the naming...

@daneryl that's a good point you have. I agree with you that migrating is not an option.
Now you bring it, we will take closer look (we overlooked it, sadly) but the flow we are expecting is that whenever an admin tries to modify a template that, after this, doesn't comply with the validations, they will need to adjust their namings before saving.
Anyways, we will review it. Thank you!

RafaPolit:
> @mfacar @RafaPolit apart from the code review.
> probably current instances already have the cases that we are not allowing on this validation ? did you guys discussed how to solve this or it just self solves by the users going to the wrongly formed templates and re saving them solving the errors ?
> another option i guess is a migration, but we can not decide arbitrarily the naming...

@daneryl As you point out, there is no way to decide which of the two is correct and what naming to adopt on the 'wrong one'.  So yes, I believe this would need to be done manually: go in, save and get the error reported back to you so you can fix. 

We *could* create a 'migration' to detect if the problem exists on an instance, and populate some sort of 'warning' in the database (this would require some planning!).  This could be interesting for when we develop breaking changes that cannot be migrated to let the admins know of the issue.

Should we invest on this now? cc @txau

app/shared/types/templateSchema.ts
fnocetti: @daneryl can you elaborate on this please? We are already sending the failing property in the dataPath (which is also shown in the client) and we are generating one error message per failed field.

app/shared/types/templateSchema.ts
```diff
@@ -155,12 +155,70 @@ ajv.addKeyword('cantDeleteInheritedProperties', {
...
 });
 
+ajv.addKeyword('cantReuseNameWithDifferentType', {
+  async: true,
+  errors: true,
+  type: 'object',
+  async validate(_schema: any, template: TemplateSchema) {
+    if (!template.properties || template.properties.length === 0) {
+      return true;
+    }
```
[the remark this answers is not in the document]
fnocetti: Actually not :) if template.properties was undefined it won't return true as expected.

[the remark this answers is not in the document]
daneryl: true ! 👍

app/api/templates/specs/templateSchema.spec.ts
```diff
@@ -189,4 +189,119 @@ describe('template schema', () => {
...
+              }),
+              expect.objectContaining({
+                dataPath: '.properties.sharedproperty2',
+                message: errorMessage,
+              }),
+            ])
+          );
+        }
+      });
+    });
```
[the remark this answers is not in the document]
daneryl: looks good !

app/shared/types/templateSchema.ts
daneryl: @fnocetti including the data that actually fails validation will provide better feeback, the actual property, maybe even the properties that have the same name and why this one is not valid, but i guess that is easily debuggable with the current error, this should be enough as it is. LGTM

app/shared/types/templateSchema.ts
```diff
@@ -155,12 +155,79 @@ ajv.addKeyword('cantDeleteInheritedProperties', {
...
+          p =>
+            p.name === property.name &&
+            (p.content !== property.content ||
+              p.type !== property.type ||
+              p.relationtype !== property.relationType)
+        );
+
+        if (matches && !propertyNames.includes(property?.name || '')) {
+          return propertyNames.concat([property?.name || '']);
+        }
```
daneryl, line 203: property name in this scenario should be always there, defaulting to blank could hide errors, i recomend in this scenarios to use the ensure utility function (from app/shared/tsUtils) until we figure out or work more on our base TS typings.
should be something like this.
```
ensure(property.name)

Comment (mfacar) [reply app/shared/types/templateSchema.ts]: Great! thanks for this tip @daneryl, it's very usefull

**Comment by daneryl** [comment]
@mfacar @RafaPolit @txau LGTM, should i merge this or we still need to decide how to approach a possible solution for the current instances with repeated properties ?

**Comment by txau** [comment]
> @mfacar @RafaPolit @txau LGTM, should i merge this or we still need to decide how to approach a possible solution for the current instances with repeated properties ?

What is going to happen with the current instances if we merge it? Will they continue working as they are an simply give an error if they try to save the faulty template?

**Comment by mfacar** [comment]
> > @mfacar @RafaPolit @txau LGTM, should i merge this or we still need to decide how to approach a possible solution for the current instances with repeated properties ?
> 
> What is going to happen with the current instances if we merge it? Will they continue working as they are an simply give an error if they try to save the faulty template?

That's right, the next time that an inconsistent template is saved, the application will throw the validation error. Perhaps in this case it might be interesting to give them the name of the other templates that contain the same property name, otherwise they coudn't know which templates are involved. What do you think @txau, @RafaPolit, @fnocetti

**Comment by fnocetti** [comment]
@mfacar It sounds like a good idea that also matches what @daneryl was suggesting. Also, I guess we haven't tried this in depth.

@txau do you thing it's worth to invest a little more time in providing the user with the names of the templates/properties in the other side of the conflict?

**Comment by txau** [comment]
> @txau do you thing it's worth to invest a little more time in providing the user with the names of the templates/properties in the other side of the conflict?

Yes. Take a look to what we already have, maybe we can articulate your suggestion from there. There is a list of matching properties when you are configuring a property. Maybe we could display a red text/icon so it catches user's attention.

**Comment by fnocetti** [comment]
So @daneryl, do not merge this yet. We will use some time to try to improve the feedback.

**Comment by fnocetti** [comment]
This is ALMOST ready for QA. Mercy will change it to ready when it is.
I'm leaving an update for the time being:
- The basis for the implementation of the feedback was to recycle the FilterSuggestions component, that showed what filter would be aggregated together to reflect the new validations.
- Now that component is always enabled (not matter wether the field is being configured as a filter).
- It lists all the properties in the system that share the same label and indicates if there is any type of conflict (differing values) for the type, content or relationtype between the listed fields.
- The conflicts are highlighted in red and a warning sign is prefixed to the text in the corresponding cell of the table.
- The component was refactored and splitted to two components to improve readability, and migrated to typescript.
- The unit tests were also refactored to test the redux-connected components and migrated to typescript.
- Added translations for the main for everything in the system context. Translations for other contexts are not yet included. As this component didn't include translations before we are debating if it is a must to close this or we can consider it tech debt.
- The validation work flow is still the same: the validation is still performed in the backend, but now there is a visual feedback (while you are editing the property) on how other related properties are configured in the system.
- Once done this refactor it should be fairly simple to implement a frontend validation. We are not including it in this PR because it's still an extra small refactor that is kinda out of the scope of this issue.

**Comment by mfacar** [comment]
> This is ALMOST ready for QA. Mercy will change it to ready when it is.

After  some development test, we have made a couple of fixes for Relationship and Multimedia properties. It is ready for review now.

File: app/react/Templates/components/specs/SimilarProperty.spec.tsx [hunk 7]
```diff
@@ -0,0 +1,95 @@
+/**
+ * @jest-environment jsdom
+ */
```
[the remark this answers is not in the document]
fnocetti: I guess I added it because I thought it was needed... good catch, thanks!

PR #3027, opened 2020-07-06
2979 Supporting number type in search term

### LINKED ISSUE #2979: Inputing a number in the search box yields a validation error
If you just input a number in the search box, you get a validation error.  Entering "quotes" fixes the problem.  We need to ensure the search is a string or change a bit the validation.

mfacar:
fixes #2979

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


konzz:
@mfacar I think it would be great to add some other scenarios to the tests even though the issue is only about the numbers in searchTerm.

mfacar:
> @mfacar I think it would be great to add some other scenarios to the tests even though the issue is only about the numbers in searchTerm.

@konzz I've added some tests for the query parameters that search could have, all in the scope of the schema only, what other tests do you think we could include for the search validation?  Maybe we should include an api call with invalid query?

konzz:
The search api has its own tests, I think this is good now.

PR #3034, opened 2020-07-09
2972 name sanitizacion revamp

### LINKED ISSUE #2972: Field name sanitization
At the moment we are sanitizing in a very strict manner field names for mongodb storage and elastic mappings. We are replacing potential illegal characters by "_" chars, in some cases leading to names composed just by underscores which in turn generates naming conflicts.

Part of the solution is having a less strict sanitation, since actually both mongo, elastic and JS itself support a wider ranger of characters.

> If you try to create an index with a name whose length exceeds 255 characters (or ~100 UTF-8 encoded bytes) you'll get an error like this one

> InvalidIndexNameException[Invalid index name [...], index name is too long, (266 > 255)]

> As for the valid characters to use in an index, the best place to look for is in their test suite, but basically an index name

    must not contain the characters #, \, /, *, ?, ", <, >, |, ,
        Since ES 7.0 onwards, : is not allowed as well
    must not start with _, - or +
    must not be . or ..
    must be lowercase

From https://stackoverflow.com/questions/41585392/what-are-the-rules-for-index-names-in-elastic-search

daneryl:
fixes #2972 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/settings/settings.ts
```diff
@@ -0,0 +1,200 @@
...
+  const updatedValues: { [k: string]: any } = {};
+  const deletedValues: string[] = [];
+
+  currentValues.forEach((value: SettingsFilterSchema & SettingsLinkSchema) => {
+    const matchValue = newValues.find(
+      v => v[matchProperty] && v[matchProperty]?.toString() === value[matchProperty]?.toString()
+    );
+
+    if (matchValue && matchValue[propertyName] !== value[propertyName]) {
+      if (value[propertyName]) {
```
konzz, line 29: I think this second "if" condition can be combined with the previous one right?

app/api/templates/templates.ts
```diff
@@ -58,11 +66,11 @@ const updateTranslation = (currentTemplate, template) => {
...
 
 export default {
-  async save(template, language) {
+  async save(template: TemplateSchema, language: string) {
     /* eslint-disable no-param-reassign */
     template.properties = template.properties || [];
-    /* eslint-disable no-param-reassign */
-    template.properties = generateNamesAndIds(template.properties);
+    template.properties = await generateNamesAndIds(template.properties);
+    /* eslint-enable no-param-reassign */
```
konzz, line 73: isn't this rule enabled by default?

daneryl: @konz yes it is, in this chunk you can see on line 64 that it is disabled, thats why then is enabled again

app/api/files/exportRoutes.ts
```diff
@@ -56,7 +56,7 @@ export default (app: Application) => {
 
         const results = await search.search(req.query, req.language, req.user);
         // eslint-disable-next-line camelcase
-        const { dateFormat, site_name } = await settings.get();
+        const { dateFormat = '', site_name } = await settings.get();
```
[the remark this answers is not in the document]
daneryl: userUtils.ts site_name its used on the very next line as a string, without defaulting to string the ts compiler warns that it can be undefined.

in this case the site_name is passed to another function that accepts a string and is defaulted there, but the variable its not used in the context of this function.

https://github.com/huridocs/uwazi/blob/900f11a1b317eeb1fded4a6a640ed1eb027ce197/app/api/files/exportRoutes.ts#L67-L72

app/api/settings/settings.ts
daneryl: you are right, done 👍

PR #3053, opened 2020-07-14
3042 range pagination

### LINKED ISSUE #3020: Standardize error handling and exit on standalone scripts
There are, at least, two standalone scripts: migrate.js and reindex_elastic.js.

These need to handle uncaught errors and do two important turn-off procedures: disconnect() and errorLog.closeGraylog() (with a callback that executes on close).

### LINKED ISSUE #3040: Reindex is not restoring Elastic's refresh interval and other settings when there is a reindex exception
As per this snippet:
```
try {
    await prepareIndex(indexUrl);
    await tweakSettingsForPerformmance(indexUrl);
    await reindex(indexUrl);
    await restoreSettings(indexUrl);
  } catch (err) {
    logErrors(err);
  }
```

When an exception is throw it will not reach restoreSettings.

### LINKED ISSUE #3042: Skip/limit pagination of mongo queries is not performant for large collections
This is a known issue of mongodb with known solutions:
https://github.com/huridocs/uwazi/blob/c6ac0749b2bb8390bd1899b8eaa67514e220ba3b/app/api/search/entitiesIndex.js#L83-L88

Skip needs to actually move the cursor through all the previous records. While the impact in small collections is minimal, for large ones the decrease in performance is quite noticeable. According to the community, the most straight forward solution is to paginate via ranged queries.

txau:
fixes #3042
fixes #3020
fixes #3040

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


txau:
The errors in e2e were due to the reindex script being called without any contents in the body. I added a couple of conditions to prevent this, but it feels like the code should not send these reindex request on the first place.

PR #3056, opened 2020-07-17
Remove snyk protect

txau:
This PR prevents snyk protect to be run on every yarn install, but adds snyk as a new CI job.

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


RafaPolit:
I am not sure I understand.  If `snyk protect` is important, and the patches are only applied to already-local dependencies, what is the use of running it as a GitWorkflow? That will patch the dependencies on the temp virtual machine doing the patching and then those patches will get lost.

If this indeed patches up our build, that is great, but then what happens to people self-hosting Uwazi? Should we include instructions to run this on their end? Include it on the `yarn production-build` flow?

Or does this actually modify the yarn lock file? (doesn't look like it).

daneryl:
> I am not sure I understand. If `snyk protect` is important, and the patches are only applied to already-local dependencies, what is the use of running it as a GitWorkflow? That will patch the dependencies on the temp virtual machine doing the workflow and then those patches will get lost.
> 
> If this indeed patches up our build, that is great, but then what happens to people self-hosting Uwazi? Should we include instructions to run this on their end? Include it on the `yarn production-build` flow?
> 
> Or does this actually modify the yarn lock file? (doesn't look like it).

@RafaPolit i think the git workflow does not patch the code, it just throws a report on possible patches and security issues, the hook on the package json actually adds a lot of time to `yarn install`

this forces a not required red check that "forces" us to update posible security issues. is this correct @txau ?

txau:
> this forces a not required red check that "forces" us to update posible security issues. is this correct @txau ?

Yes, that is correct. This is just reporting the vulnerable modules for us to fix. The idea is to have security checks as part of our CI. I've been researching other tools and we could do things like automatically pen-test the app etc, but it takes way longer time and it doesn't add much value.

Snyk will change the package.json when a vulnerability is fixable via an upgrade. It can also patch some modules that haven't been patched by the maintainer yet by directly modifying the code in node_modules, which is what @RafaPolit is mentioning, but I don't think we want this option.

PR #3057, opened 2020-07-19
Stringify the error since not always contains a message

txau:
I found some situations where the exceptions do not contain a message but a reason. Since this particular error logging is very well bounded to the scope "uncaught elastic reindexing errors", I think is ok to treat it as a special case and just stringify the whole error object. Otherwise, I was getting an "undefined" message on index mapping misconfigurations.

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/database/reindex_elastic.js
+++ b/database/reindex_elastic.js
@@ -97,7 +97,9 @@ const processErrors = async err => {
   if (err instanceof IndexError) {
     process.stdout.write('\r\nWarning! Errors found during reindex.\r\n');
   } else {
-    errorLog.error(`Uncaught Reindex error.\r\n${err.message}\r\nWill exit with (1)\r\n`);
+    errorLog.error(
+      `Uncaught Reindex error.\r\n${JSON.stringify(err, '', '\r')}\r\nWill exit with (1)\r\n`
+    );
     await endScriptProcedures();
     throw err;
   }


txau:
> @txau Stringify of an Error actually results in "{}" string, what is the scenario here ? all errors should have a message.

Well, I'm seeing the stringfyed errors, not just '{}'. Some errors are coming without a message. These are errors coming from Elastic that are not indexing errors, but ie. configuration errors. This is what I'm getting from stringify:
```
{
"json": {
"error": {
"root_cause": [
{
"type": "mapper_parsing_exception",
"reason": "no handler for type [icu_collation_keyword] declared on field [sort]"
}
],
"type": "mapper_parsing_exception",
"reason": "Failed to parse mapping [_doc]: no handler for type [icu_collation_keyword] declared on field [sort]",
"caused_by": {
"type": "mapper_parsing_exception",
"reason": "no handler for type [icu_collation_keyword] declared on field [sort]"
}
},
"status": 400
},
"status": 400
}
```

Previously I was getting 'inderfined'.

daneryl:
@txau if a new Error() or any unhandled JS errors is thrown, that results in "{}".
I do not know where that error comes from but it looks like an Object instead of an Error instance, it should be treated specifically or track where this comes from to actually throw a proper Error.

how can this be replicated ?

txau:
@daneryl I guess I screwed up with the branches and the commit for this PR was also in the PR @mfacar was QAing, so it got automatically merged? Sorry, my bad.

mfacar:
Sorry, I thought it was an on-flight fix, I should have asked about it.

daneryl:
@txau @mfacar i will do a PR fix tomorrow NP

PR #3069, opened 2020-07-21
Added support for Elastic 7.8 + updated readme for ICU support

### LINKED ISSUE #3065: Add USE_ELASTIC_ICU env variable to README
After #3058, please add the option USE_ELASTIC_ICU to the readme ENV options.

### LINKED ISSUE #3068: Support Elastic 7.8
For some reason Elastic 7.8 is triggering the HPE_HEADER_OVERFLOW error in node 10.21. Node changed the default limitation of max-header-size from 80kb to 8kb since "headers should not contain metadata and this information should be in the body". This change was introduced to avoid certain DDoS attacks.

For whatever reason Elastic 7.8 is sending bigger headers than Elastic 7.6. While this is probably something that needs to be fixed on Elastic's end, by manually increasing the supported header size to > 16kb this issue is fixed thus we can upgrade Elastic.

The easiest way to fix this header size issue is by passing the CLI flag --max-http-header-size 20000. This flag is compatible with Elastic 7.6 so it is not introducing a breaking change.

txau:
fixes #3068
fixes #3065

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


txau:
I tested this PR on:
- yarn migrate
- yarn hot
- yarn run-production

to make sure that reindexing problems won't trigger when users are changing template properties. I cannot think of other scenarios were this may be a problem.

txau:
This is being reimplemented in #3206

PR #3080, opened 2020-07-23
3077 allow specific types share names

### LINKED ISSUE #3077: Allow specific namings on templates to belong to different types
The scenarios of single vs. multi should be allowed to share names in the templates.

That is: select and multiselect, date and multidate, date-range and multi date-range.  

Also, text and rich text should be allowed to share the same name.

- [ ] This requires testing that filtering for those scenarios is working as expected

fnocetti:
fixes #3077 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/templates/specs/templateSchema.spec.ts
+++ b/app/api/templates/specs/templateSchema.spec.ts
@@ -236,6 +236,22 @@ describe('template schema', () => {
           type: 'daterange',
         },
         {
+          name: 'validpropertymultiselect',
+          label: 'validPropertyMultiSelect',
+          type: 'multiselect',
+          content: 'thesauriId4',
+        },
+        {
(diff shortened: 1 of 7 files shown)


mfacar:
@txau @RafaPolit, the PR already allows us to have compatible properties with the same name. 
As @fnocetti told us, filters with compatible types and the same name don't behave as totally expected. If you agree, we can close this PR while we raise the solution for total compatibility.

fnocetti:
@mfacar can you check again? Thanks!

mfacar:
LGTM! I'll merge when the synchronization with development finishes the checks

PR #3084, opened 2020-07-27
update the README to include Uwazi submodules download instructions

### LINKED ISSUE #3083: README has outdated install instructions that does not include cloning the included git submodules
Issue:
The current git clone command does not clone any git sub-modules included in the Uwazi repo. i.e `$ git clone https://github.com/huridocs/uwazi.git`

So in case one needs to run the fixtures maybe for end-to-end testing there's no immediate clear instructions on how to.

Fix:
Cloning the main Uwazi repository for the first time with `git clone --recurse-submodules https://github.com/huridocs/uwazi` will download the repository including its sub-modules such as the `uwazi-fixtures`.

If the main Uwazi repository had already been cloned/downloaded and now one wants to load its sub-modules, you can run
`git submodule update --init` while inside the uwazi repo directory.

elric-wamugu:
fixes #3083

PR checklist:
- [x] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


install.sh
```diff
@@ -0,0 +1,45 @@
...
+
+echo "Building production"
+yarn production-build
+
+echo installing into $install_path
+mkdir -p $install_path
+mkdir -p $install_path/custom_uploads
+mkdir -p $install_path/log
+mkdir -p $install_path/temporal_files
+mkdir -p $install_path/uploaded_documents
```
RafaPolit, line 14: This is not respecting the env variable UPLOADS_FOLDER which is described as configurable on the custom configuration description.  https://github.com/huridocs/uwazi/wiki/Install-Uwazi-on-your-server

install.sh
```diff
@@ -0,0 +1,45 @@
...
+
+echo "Copy app files..."
```
RafaPolit, line 16: Previous statements used verbs in the form of "building", but now it's "Copy"? Shoud this be "Copying..."?

install.sh
```diff
@@ -0,0 +1,45 @@
...
+cp ./prod/package.json $install_path/package.json
+cp ./prod/run.js $install_path/run.js
+cp ./prod/server.js $install_path/server.js
+cp ./prod/message.js $install_path/message.js
+
+cd $install_path
+echo "Updating database schema..."
+node ./run.js ./app/api/migrations/migrate.js
+
+echo "Reindexing ElasticSearch..."
```
RafaPolit, line 44: Do we really want to force reindex each time? We talked about creating a dedicated migration that would run reindex if the new development requires it.

This approach of always running the reindex is certainly a "safer" option right now, but it could be very time consuming.

install.sh
txau: I guess is the same for temporal_files and uploaded_documents. Why are you pointing out this one particularly? In the configuration description is using the env var UPLOADS_FOLDER, but this is used for uploaded_documents, which is in this script. Can you explain what is the problem?

install.sh
txau: I guess that would be for a different development. At the moment we are instructing users to always run reindex after migrate.

install.sh
txau: @RafaPolit awaiting your clarification to merge this.

RafaPolit: Yes, perhaps there is no problem here.  It will work as expected, yes.

PR #3085, opened 2020-07-28
2511 graphs e2e tests puppeteer

### LINKED ISSUE #2511: Create more extensive Page and Graphs-in-Page e2e Tests

elric-wamugu:
fixes #2511

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


e2e/suites/graphs.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    await page.type(localSelectors.pageContentsInput, graphs.listChart);
+    await expect(page).toMatchElement('button', { text: 'Save' });
+    await expect(page).toClick('button', { text: 'Save' });
+    await expect(page).toMatch('Saved successfully.');
+    await expect(page).toMatch('(view page)');
+  });
+
+  it('should display List chart graph in page with no more than a 3% difference', async () => {
+    const createdPageLink = await page.$eval(localSelectors.createdPageLink, el => (<HTMLLinkElement>el).href);
+    await page.goto(createdPageLink);
```
daneryl, line 110: @elric-wamugu  i think we already discussed this on the nightmarejs version but can't we just do a click on the 'a' instead of an eval to get the url ? something like, same for the other cases where this technique is used.
```
toClick('a', { text: '(view page)'}) 
```

e2e/suites/graphs.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  });
+
+  it('should display Pie chart graph in page with no more than a 3% difference', async () => {
+    const createdPageLink = await page.$eval(localSelectors.createdPageLink, el => (<HTMLLinkElement>el).href);
+    await page.goto(createdPageLink);
+    await page.waitFor(2000);
+
+    const chartContainer = await page.$('.recharts-responsive-container');
+    // @ts-ignore
+    const chartScreenshot = await chartContainer.screenshot();
```
daneryl, line 87: @elric-wamugu i think we should no use ts-ignore unless we abosolutely need it, this case can be solved with a condition or using the ensure utility. same for the rest of @ts-ignore on the same case.

```
const chartContainer = ensure<ElementHandle>(await page.$('.ListChart '))
```

e2e/suites/graphs.test.ts
```diff
@@ -0,0 +1,127 @@
+/*global page*/
+
+import { adminLogin, logout } from '../helpers/login';
+import proxyMock from '../helpers/proxyMock';
+import insertFixtures from '../helpers/insertFixtures';
+// @ts-ignore
```
daneryl, line 6: @elric-wamugu this ignore not needed, you also installed types for this.

package.json
```diff
@@ -248,13 +249,15 @@
     "eslint-plugin-prettier": "^3.1.2",
     "eslint-plugin-react": "v7.12.4",
     "fetch-mock": "v7.3.3",
+    "i": "^0.3.6",
```
daneryl, line 252: @elric-wamugu this package is not being used.

e2e/suites/graphs.test.ts
elric-wamugu: @daneryl the main reason for using evaluate on html element here is to try and persist the page object instance (needed later for page element screenshot). The view page anchor element has the `target="_blank"` which opens the page on a new tab and thereby switching the browser target context. I've reworked the code to switch back the browser page context to the newly opened tab. 
This link provides more context into the issue https://github.com/puppeteer/puppeteer/issues/386

e2e/helpers/graphs.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+  
+  const chartScreenshot = await chartContainer.screenshot();
+  expect(chartScreenshot).toMatchImageSnapshot({
+    failureThreshold: 0.03,
+    failureThresholdType: 'percent',
+    allowSizeMismatch: true,
+  });
+
+  await graphsPage.close();
```
daneryl, line 33: @elric-wamugu i like the idea of abstracting the graph rendering, but can you move pls the testing part to each of the test ? this way you can see the expect on the test itself and you can avoid the conditional for the type.

PR #3099, opened 2020-08-07
3050 tableview

### LINKED ISSUE #3050: tableview
[PRD link](https://docs.google.com/document/d/11FQk2z9ypfavW4Pk-ASxcCSMMPH0P5Cd4LDWDc9fbRc/)

[Design link](https://www.figma.com/file/xjsz7r65epN4B0b96zilac/UWAZI-redesigns?node-id=346%3A3734)

Let me know if something isn't clear in the doc comments

fnocetti:
fixes #3050 

Things that were not included in this MVP
- Implementation of the re-designs (layout and new toggle buttons)
    - New layout
    - New toggle buttons
    - Filters side panel hidden by default
- Specific behavior on rows selection and multi-selection
    - Multiple selected document list should not appear on side panel
    - Selecting entities with the checkbox should not open the side panel
    - Clicking on the row should not change selection status but open the side panel metadata view
- Column resizing (and re-ordering?)
- Filters and orders on table headers. Filters matching highlights on table cells values
- Complex types rendering: geolocation, image, media, nested (to be reviewed with Simon)
- “Hide unselected” and “select all” (title header) checkboxes
- Export CSV button on main view instead of side panel

They're all individually implementable, should we create an issue for each?

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Layout/TableViewer.tsx
```diff
@@ -0,0 +1,64 @@
...
+  documents: any;
+  storeKey: 'library' | 'uploads';
+  clickOnDocument: (...args: any[]) => any;
+  onSnippetClick: (...args: any[]) => any;
+  deleteConnection: (...args: any[]) => any;
+  search: any;
+  templates: any;
+  thesauris: any;
+  columns: any[];
+}
```
RafaPolit, line 17: The amount of "any"s here, looks like we are just using TS for the sake of using it, I don't think that is the correct approach.

But there is a lot of problems here: templates and thesauris are brought directly from the state (store), they are not changed, and they are passed on to TableRow.  That is completely unneeded.  You can just mapStateToProps in TableRow to those two values and remove that from here.

I understand that is how refactors start, but we need to take them to the next step.

"Search" is not being used in this component, at all!

app/react/Layout/TableViewer.tsx
```diff
@@ -0,0 +1,64 @@
...
+class TableViewerComponent extends Component<DocumentViewerProps> {
+  render() {
+    const columns = this.props.columns.filter(c => !c.get('hidden'));
+    return (
+      <div className="tableview-wrapper">
+        <table>
+          <thead>
+            <tr>
+              {columns.map((column: any, index: number) => (
+                <th className={!index ? 'sticky-col' : ''} key={index}>
```
RafaPolit, line 28: This is the precise reason why our rules prevent using 'index' as key.  Here, columns can change, but the index will remain the same.  Please, find a way to use a different key.  Perhaps column name is disambiguation enough?

app/react/Layout/TableViewer.tsx
```diff
@@ -0,0 +1,64 @@
...
+              ))}
+            </tr>
+          </thead>
+          <tbody>
+            {this.props.documents.get('rows').map((document: any, index: number) => (
+              <TableRow
+                {...{
+                  document,
+                  columns,
+                  key: index,
```
RafaPolit, line 40: again, here, index could be document.get("_id") or something better than index?

app/react/Layout/TableViewer.tsx
```diff
@@ -0,0 +1,64 @@
...
+                  onClick: this.props.clickOnDocument,
+                  storeKey: this.props.storeKey,
+                  templates: this.props.templates,
+                  thesauris: this.props.thesauris,
```
RafaPolit, line 44: As mentioned above, there is no need to pass this over. They can be fetched in TableRow in map state to props.

app/react/Layout/TableViewer.tsx
```diff
@@ -0,0 +1,64 @@
...
+        </table>
+      </div>
+    );
+  }
+}
+
+const mapStateToProps = (state: any, props: DocumentViewerProps) => ({
+  templates: state.templates,
+  thesauris: state.thesauris,
+  authorized: !!state.user.get('_id'),
```
RafaPolit, line 59: Authorized is not used at all in the component either.

app/react/Library/components/TableRow.tsx
```diff
@@ -0,0 +1,170 @@
...
+        .get('selectedDocuments')
+        .find((doc: any) => doc.get('_id') === ownProps.document.get('_id'))
+    : false;
+  return {
+    selected,
+    document: ownProps.document,
+    columns: ownProps.columns,
+    templates: state.templates,
+    thesauris: state.thesauris,
+    onClick: ownProps.onClick,
```
RafaPolit, line 166: I am confused here: mapStateToProps extends the passed props (ownProps) with the return props.

If the Component already has document, columns, and onClick props, you don't need to return them again here from ownProps.  They are already passed.  The only reason to do this would be to map to different named props, which is not the case.

app/react/Library/components/TableRow.tsx
```diff
@@ -0,0 +1,170 @@
...
+            this.props.zoomLevel
+          )
+        )}
+      </tr>
+    );
+  }
+}
+
+function mapStateToProps(state: any, ownProps: TableRowProps) {
+  const selected = ownProps.storeKey
```
RafaPolit, line 155: storeKey is defined as either 'library' or 'uploads' in the Props definition.  Yet here, there is a tertiary contemplating the option of not having storeKey?  If this is a possibility, please include it as conditional in the props definition.  If not, then why this tertiary?

app/react/Library/components/TableRow.tsx
```diff
@@ -0,0 +1,170 @@
...
+  });
+  return formattedDoc;
+}
+
+function displayCell(
+  document: any,
+  column: any,
+  index: number,
+  firstColumnCheckbox: any,
+  zoomLevel: number
```
RafaPolit, line 88: So many "any"s again.  Should we review this policy @daneryl @konzz ?

If we are going to type everything as "any", what is the point of TS?

app/react/Library/components/TableRow.tsx
```diff
@@ -0,0 +1,170 @@
...
+  formattedDoc = formatter.prepareMetadata(formattedDoc, templates, thesauris, null, {
+    sortedProperty: 'creationDate',
+  });
+  formattedDoc.metadata.forEach((prop: any) => {
+    formattedDoc[prop.name] = prop;
+  });
+  return formattedDoc;
+}
+
+function displayCell(
```
RafaPolit, line 83: This should really be a different component name simply Cell.

Don't be afraid to have many smaller files.  It makes for much easier reusability, testing, maintenability.

app/react/Library/components/TableRow.tsx
```diff
@@ -0,0 +1,170 @@
...
+}
+
+class TableRowComponent extends Component<TableRowProps> {
+  constructor(props: TableRowProps) {
+    super(props);
+    this.onClick = this.onClick.bind(this);
+    this.firstColumnCheckbox = this.firstColumnCheckbox.bind(this);
+  }
+
+  onClick(e: { preventDefault: () => void }) {
```
RafaPolit, line 117: Please, use semantic names for functions.  This is a particular onClick for the checkbox of the TableRow. Should it be named "select", or "checkboxClick"?

app/react/Library/components/TableRow.tsx
```diff
@@ -0,0 +1,170 @@
...
+    if (this.props.onClick) {
+      this.props.onClick(e, this.props.document, this.props.selected);
+    }
+  }
+
+  firstColumnCheckbox(index: number) {
+    return (
+      !index && (
+        <input type="checkbox" checked={this.props.selected} onClick={this.onClick.bind(this)} />
```
RafaPolit, line 126: the onClick function is already bound to "this" on the constructor.  Is this binding here really necessary?

app/react/Library/components/TableRow.tsx
```diff
@@ -0,0 +1,170 @@
...
+      result = undefined;
+      break;
+  }
+  return result;
+}
+
+function formatDocument(document: any, templates: TemplateSchema[], thesauris: ThesaurusSchema[]) {
+  let formattedDoc = document.toJS();
+  const template = templates.find((t: TemplateSchema) => t.get('_id') === formattedDoc.template);
+  formattedDoc.templateName = template?.get('name');
```
RafaPolit, line 73: We already have a component that renders template names based on template ID.  Is this really necessary here or could we use that component instead?

app/react/Library/components/TableRow.tsx
```diff
@@ -0,0 +1,170 @@
...
+  formattedDoc = formatter.prepareMetadata(formattedDoc, templates, thesauris, null, {
+    sortedProperty: 'creationDate',
+  });
+  formattedDoc.metadata.forEach((prop: any) => {
+    formattedDoc[prop.name] = prop;
```
RafaPolit, line 78: This is a very dangerous procedure.  Why move the metadata props to the root of the document?  What happens if a property is named with the same name as other root components?  Title is the most obvious example that comes to mind.  Let's assume a template overrides "title" to be "name" of person, and they create a metadata property named "title" which is "Mr." or "Mrs", for example.

This procedure will override the 'name' (title) of the entity with Mr. or Mrs..  I don't think this is correct at all.

app/react/Library/components/TableRow.tsx
```diff
@@ -0,0 +1,170 @@
...
+      result = <GeolocationViewer points={prop.value} onlyForCards />;
+      break;
+    default:
+      result = undefined;
+      break;
+  }
+  return result;
+}
+
+function formatDocument(document: any, templates: TemplateSchema[], thesauris: ThesaurusSchema[]) {
```
RafaPolit, line 70: Document vs. Entities is an old naming problem we have from the time when there was such a distinction.  Now, everything is really an entity.  Documents are, now, particular cases of entities that have a PDF attached to them.

Please, rename everything that mentions Document to Entity and only keep Document for those cases it really is differentiating between "regular entities" and "documents".

app/react/Layout/TableViewer.tsx
fnocetti: Totally agree! That's exactly what happened: this refactor took many many shapes on the process, and we failed to revisit some initial steps. I'm glad you caught this things!

app/react/Library/components/TableRow.tsx
fnocetti: Yep, agreed. We'll revisit this

app/react/Library/components/TableRow.tsx
fnocetti: I think the policy is OK. We should have correctly typed what needs to be typed.

app/react/Library/components/TableRow.tsx
fnocetti: No, its not 👍

app/react/Library/components/TableRow.tsx
fnocetti: Yeah, agree

app/react/Library/components/HiddenColumnsDropdown.js
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { DropdownList } from 'app/Forms';
```
RafaPolit, line 3: To avoid a bubble up of not being able to use TS because of this component, please, just migrate this file to TS and ts_ignore (or however it's used) this subcomponent.

app/react/Library/components/TableRow.tsx
mfacar: Given that the function that we are using to format the entity already has contemplated putting the name of the template, we fixed the assignation and use that property instead. https://github.com/huridocs/uwazi/blob/5bed5193603aaab021a925248248c7e10f814399/app/react/Metadata/helpers/formater.js#L322

fnocetti:
This is ready for QA. Nonetheless, we are having an intermittent issue with timeouts in the e2e that we're still debuging so please **DO NOT MERGE** this PR yet

RafaPolit:
There is a conflict with the recently merged PR, can you please solve the conflict so we can test the functionality?  

Thanks.

app/react/Layout/DocumentsList.js
```diff
@@ -78,9 +84,16 @@ class DocumentsList extends Component {
...
-        <b>{documents.get('totalRows')}</b> <Translate>documents</Translate>
+        {this.props.selectedDocuments.size > 0 && (
+          <>
+            <b> {this.props.selectedDocuments.size} </b> <Translate>selected of</Translate>
+          </>
+        )}
+        <b> {this.props.documents.get('rows').size} </b> <Translate>shown of</Translate>
+        <b> {documents.get('totalRows')} </b> <Translate>documents</Translate>
       </span>
     );
```
RafaPolit, line 98: This is the exact type of piece of code that should become a separate pure React component that receives a very simple couple of props.

It makes for much simpler components that have simple code and you don't get all this extra "inner components" that clutter the code.

I'm not saying we need to change this now, but it something to keep in mind.

app/react/Layout/DocumentsList.js
```diff
@@ -137,40 +149,29 @@ class DocumentsList extends Component {
           </div>
           {(() => {
             if (view !== 'graph') {
+              const { CollectionViewer } = this.props;
```
RafaPolit, line 152: The React community usually does this const assignment as the very first line of the render() function of a Component.

In this case, on line 76 there is already this assignment.  So, please just move the CollectionViewer assignment over to that group of assignments.

app/react/Layout/DocumentsList.js
```diff
@@ -137,40 +149,29 @@ class DocumentsList extends Component {
...
-                  ))}
-                </RowList>
+                <CollectionViewer
+                  {...{
+                    rowListZoomLevel,
+                    storeKey: this.props.storeKey,
+                    clickOnDocument: this.clickOnDocument,
+                    onSnippetClick: this.props.onSnippetClick,
+                    deleteConnection: this.props.deleteConnection,
+                    onEndScroll: this.onEndScroll,
```
RafaPolit, line 161: As described above, consider if we really need to pass this as a full onEndScroll or if we want to just pass the loadMore function and reimplement this on the actual place where the infinite scroll is happening.

app/react/Layout/SidePanel.js
```diff
@@ -16,6 +20,11 @@ SidePanel.propTypes = {
   children: PropTypes.oneOfType([PropTypes.object, PropTypes.array]),
   className: PropTypes.string,
   open: PropTypes.bool,
+  fixed: PropTypes.bool,
+};
+
+SidePanel.defaultProps = {
+  fixed: false,
```
RafaPolit, line 27: I'm not sure I understand the semantics of this prop?  Is this only passed when the sidepanel is not allowed to be closed?  What does fixed mean?

Because on the entity view, we also have a 'fixed' side panel, yet we are probably not going to flag that with table-vew-mode.  Should this have a different name?

app/react/Layout/TableRows.tsx
```diff
@@ -0,0 +1,54 @@
...
+const defaultProps = {
+  documents: Immutable.fromJS({ rows: [] }),
+  rowListZoomLevel: 2,
+};
+
+class TableRowsComponent extends Component<TableRowsProps> {
+  static defaultProps = defaultProps;
+
+  render() {
+    console.log('render rows');
```
RafaPolit, line 28: Please, remove this!!!!

app/react/Layout/TableRows.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { IStore, TableViewColumn } from 'app/istore';
+import { IImmutable } from 'shared/types/Immutable';
+import { EntitySchema } from 'shared/types/entityType';
+import Immutable from 'immutable';
+
+interface TableRowsProps {
+  columns: TableViewColumn[];
+  storeKey: 'library' | 'uploads';
+  clickOnDocument: (e: React.SyntheticEvent, doc: EntitySchema, active: boolean) => any;
+  onEndScroll: (amount: number, from: number) => void;
```
RafaPolit, line 14: This prop is not being used in the component.

RafaPolit (changes requested):

app/react/Layout/DocumentsList.js
```diff
@@ -44,6 +42,14 @@ class DocumentsList extends Component {
...
   }
 
+  onEndScroll() {
+    const from = this.props.documents.get('rows').size;
+    const DEFAULT_PAGE_SIZE = 30;
+    if (from) {
+      this.loadMoreDocuments(DEFAULT_PAGE_SIZE, from);
+    }
+  }
+
```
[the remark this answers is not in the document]
mfacar: We had to relocate this function from TableViewer because it depends on the number of loaded entities, and since we introduce a new component for the rows in order to prevent the rerendering of the table at receiving the entities, this new component couldn't fire the scroll event on the table.
Actually, this method can be used for ListView too instead of the load more buttons.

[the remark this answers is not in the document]
fnocetti: Well, you're right and I totally agree with you. We did this trying to balance the use of states within the components. While the ideal solution is what you propose, that means that the table view component calling this should know the amount of documents loadad (which doesn't currently, because the scroll is handled at a parent component and the rows are rendered at a child component), meaning that the table itself will re render when the rows change, which we were trying to avoid. Anyways, I think we can work it in another way. This handler is not necessarily a scroll handler but is about loading more. So we could rename it to "loadNextPage" (or simillar) which could be more descriptive, and could be easily used in future display strategies. In fact, it's a "loadmore" that doesn't require you to care about the current state (on the client side of the method.

app/react/Layout/DocumentsList.js
fnocetti: Totally true. In fact I think we should do it. But also (and we are proposing that in our research) adding so many small components in the current folder structure is kind of confusing. Our current modules are already stuffed of files, we might agree it's hard to look for files within one single folder. At this point I think we should balance between modularity and folder stuffing? But yeah, I think what you say is the way to go!

app/react/Layout/DocumentsList.js
fnocetti: Already answered on another comment

app/react/Layout/SidePanel.js
fnocetti: That's the semantic, yes. Again, we tried to balance :D We can agree there's a mix between abstraction and ad-hoc design here. We can go with a deeper refactor if you think it makes sense.

app/react/Layout/DocumentsList.js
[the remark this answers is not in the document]
RafaPolit: Ok.  Yeah, I can see the argument here.  Still, the value of documents loaded could be stored in the redux store and consumed down the road as well.  I'm OK with this.
I agree it *could* be used in DocumentsList, but not as implemented.

I agree it's a balanance: where should it be, what is store values and which are state, what is a subcomponent what gets a new file.  Yes.  That is what Konz meant by: "we don't have a unified way of doing things".  We should aim to have all these things somehow standardized. 

Still, I'm OK with this.

app/react/Layout/DocumentsList.js
RafaPolit: As long as the folder structure is 'correct' and that the components are where they should (reusable components on shared directories and specific ones in their own module), I am OK with lots of files of components and cleaner component code.

I'm not really OK with the reverse scenario: fewer files but thousand-long files. 

Maybe this is a good topic for the tech call.

app/react/Layout/SidePanel.js
RafaPolit: Ok.

app/react/Layout/TableRows.tsx
RafaPolit: I believe that our lint is not picking up TSX files: the previous PR that had index used as keys should have been picked up as well.  It would be good if you, that already did some work on the linters, could double check that, please.

app/react/Layout/DocumentsList.js
```diff
@@ -158,7 +158,7 @@ class DocumentsList extends Component {
                     clickOnDocument: this.clickOnDocument,
                     onSnippetClick: this.props.onSnippetClick,
                     deleteConnection: this.props.deleteConnection,
-                    onEndScroll: this.onEndScroll,
+                    onEndScroll: this.loadNextGroupOfEntities,
```
RafaPolit, line 161: I like this semantics much better!

fnocetti: Still, the CollectionViewer "interface" needs its onEndScroll prop renamed to loadNextGroupOfEntities. Sorry it's not pushed, we discussed this just really late yesterday.

app/react/Layout/DocumentsList.js
fnocetti: OK, agreed. We will work that.
Also, we are pitching this topic on the next tech call as part as our research :)

app/react/Layout/TableRows.tsx
```diff
@@ -0,0 +1,51 @@
...
+        {this.props.documents.get('rows').map((entity: any) => (
+          <TableRow
+            {...{
+              entity,
+              columns: this.props.columns,
+              key: entity.get('_id'),
+              clickOnDocument: this.props.clickOnDocument,
+              storeKey: this.props.storeKey,
+              zoomLevel: this.props.rowListZoomLevel,
+            }}
```
RafaPolit, line 38: This is a weird use of the ellipsis syntax to use the spread operator.  Why not using the native HTML / React way of assigning keys?
```
<TableRow
   entity={entity}
   columns={this.props.columns}
   ...
/>
```

I agree this also works, and if the namings all coincide (like entity) it would work great, but since we are mostly reassigning, it doesn't make much difference.

On that note, still, doing the trick of assigning variables to props on the first line of the render method:
```
const { columns, clickOnDocument, storeKey, zoomLevel } = this.props;
```
and then just:
```
<TableRow {...{entity, clickOnDocument, storeKey, zoomLevel, key: entity.get('_id')}} />
```
could be simpler?

Just small things to repeat in every component.

app/react/Layout/TableViewer.tsx
```diff
@@ -0,0 +1,63 @@
...
+          </thead>
+          <tbody>
+            <TableRows
+              {...{
+                columns,
+                clickOnDocument: this.props.clickOnDocument,
+                storeKey: this.props.storeKey,
+                onEndScroll: this.props.onEndScroll,
+              }}
+            />
```
RafaPolit, line 46: Same as above, use the first line to assign variables to props, simplifies this greatly.

app/react/Layout/TableViewer.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+const getTableViewColumnsSelector = (state: EntityDisplayState) => state.ui.get('tableViewColumns');
+export const selectTableViewColumns = createSelector(getTableViewColumnsSelector, columns =>
+  columns?.toJS()
+);
+
+const mapStateToProps = (state: IStore, props: TableViewerProps) => ({
+  columns: selectTableViewColumns(state[props.storeKey]),
+});
```
RafaPolit, line 61: I'd like @daneryl to comment on this.  What do you think of using toJS() combined with selectors on mapStateToProps?

app/react/Layout/specs/TilesViewer.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+    });
+
+    describe('Clicking on a document', () => {
+      it('should call on props.clickOnDocument if present', () => {
+        component
+          .find(Doc)
+          .at(0)
+          .simulate('click', 'e', 'other args');
+        expect(props.clickOnDocument.apply.calls.mostRecent().args[1][0]).toBe('e');
+        expect(props.clickOnDocument.apply.calls.mostRecent().args[1][1]).toBe('other args');
```
RafaPolit, line 72: Why is the "apply" needed?

app/react/Library/Library.js
```diff
@@ -57,9 +57,15 @@ export default class Library extends RouteHandler {
...
     return (
-      <LibraryLayout>
-        <LibraryModeToggleButtons storeKey="library" zoomIn={this.zoomIn} zoomOut={this.zoomOut} />
-        <DocumentsList storeKey="library" />
+      <LibraryLayout fixedSidePanels={this.props.fixedSidePanels}>
+        <LibraryModeToggleButtons
+          storeKey="library"
+          zoomIn={this.zoomIn}
+          zoomOut={this.zoomOut}
+          showColumnSelector={this.props.viewer !== undefined}
```
RafaPolit, line 65: I don't like this approach for the entire adding of the TableViewer. I commented on the side panel "fixed" thing, and here we find the same criteria: "If I passed the viewer prop, show the column selector".  What if we later on add a new viewer that is named "mapViewer", or "relationshipsTreeViewer", all of  a sudden we get a column selector?

Please, do this very explicitly! Something like "if props.viewer is an instance of TableView" or, for example, explicitly passing a "showColumnSelector" prop that is controlled where the viewer is defined.  Something really explicit that we want this shown, instead of a secondary effect of passing a viewer.

app/react/Library/LibraryTable.js
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  render() {
+    return (
+      <Library
+        viewer={TableViewer}
+        location={this.props.location}
+        fixedSidePanels
+        showFiltersButton
+      />
```
RafaPolit, line 19: This is the place to say: "show the column selector", "pass the side panel a table-view-mode class".

Not the assumption of: "if I have viewer, put the column selector", "if I passed fixedSidePanels, put a table-view-mode".   

It could be as explicit as: "sidepanelClassName="table-view-mode", and you pass that to the side panel on libraryLayout, for example.

RafaPolit (changes requested):

app/react/Layout/specs/TilesViewer.spec.tsx
fnocetti: We just moved this code from other file, but we will check!

app/react/Library/Library.js
fnocetti: Yeah, we overlooked this. Will correct it.

app/react/Library/LibraryTable.js
fnocetti: Yes, is part of the same thing. :+1:

app/react/Layout/DocumentsList.js
```diff
@@ -242,6 +243,7 @@ DocumentsList.propTypes = {
     pathname: PropTypes.string,
     query: PropTypes.object,
   }),
+  CollectionViewer: PropTypes.instanceOf(Object),
```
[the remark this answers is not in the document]
fnocetti: We changed this to func. Any of the other alternatives won't work because connected components are HOFs and for some reason the checker don't consider them as renderable.

app/react/Layout/TableViewer.tsx
daneryl: toJS is always slower than just passing the object through, in this case though this avoids the major bottleneck, that is unnecessary re-renders. if this is a big component rendered only once then this should be fine in terms of performance.

app/react/Layout/DocumentsList.js
```diff
@@ -77,25 +84,26 @@ class DocumentsList extends Component {
...
       LoadMoreButton,
       rowListZoomLevel,
+      CollectionViewer,
     } = this.props;
-    let counter = (
-      <span>
-        <b>{documents.get('totalRows')}</b> <Translate>documents</Translate>
-      </span>
+
+    const totalConnections = connections
```
fnocetti, line 90: @mfacar should we include this piece of code into the DocumentCounter component?

mfacar: > @mfacar should we include this piece of code into the DocumentCounter component?
Yes, we could do it, in that case we have to pass the connections and connectionsGroups objects to the DocumentCounter, I didn't include before because I tried to pass only numeric values, but since there isn't any other use of totalConnections in DocumentList it makes total sense to move into the specific component

fnocetti: @mfacar well, you're right. It makes sense to make the calculation on the DocumentList also, since the DocumentCounter is merely presentational.

mfacar:
@RafaPolit this PR is already updated with the fixing of multiples request  #3156, as we expected, this solves the problem of infinite scroll.  It is ready to be merged.

RafaPolit:
I am having several problems regarding the infinite scroll and the central pane scrolling:
- The `tableview-wrapper` element has a 100vh max height.  This means that I can achieve a combination of scrolls where I cannot see the last rows, yet I am led to believe that that is the last row.  Usual practice is to leave the infinite scroll area with a `calc(100vh - XYZpx)`, where XYZ is the fixed sizes of the UI.  Mostly, in this case we should discount the 50px of the header (this will in turn complicate custom designs that have different header sizes, but that is expected).  
- Personally, I would also discount the footer and buttons size so that the entire view is always WITHOUT vertical scroll and then I would add the scroll to infinite pain.
- This last point is also crucial to my next bullet point because:
- The only place I can scroll horizontally is if I reach the very end of the table and drag, very unintuitively, the view.  This works, of course, with touch gestures as expected, but mouse scrolling on the main view is not possible by dragging.  You can middel-wheel scroll (not available to all mice) and you CANNOT HORIZONTAL SCROLL, at all!
- Mobile view is completely broken with selectors flowing undeneath the search box and buttons wrapping up over text

RafaPolit:
I believe this is a good MVP.

PR #3115, opened 2020-08-18
promisify multer, abstract upload middleware

### LINKED ISSUE #3110: Error when uploading attachment to all languages in multi-tenant
```
2020-08-14T18:04:03.842Z [cejil-staging] 
url: /api/attachments/upload
body: {
 "entityId": "5dea667114b2ff790131309d",
 "allLanguages": "true"
}
Error: There is no tenant on the current async context
```

### LINKED ISSUE #3112: Uploading this file crashes the instance in multi-tenant mode
[2003.02320v1.pdf](https://github.com/huridocs/uwazi/files/5077709/2003.02320v1.pdf)

image

It doesn't crash in single tenant code.

daneryl:
fixes #3112
fixes #3110

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


RafaPolit:
We found the "no tenant" issue is present on the Remote Public Form.

To reproduce:
- create two synced instances
- configure a remote public form on the slave
- submit an entity without file... all goes well
- attach a file or attachment and it throws the error

daneryl:
> We found the that both Public Form and Remote Public Form, if we turn on throttling and send large-ish files (16Gb on Fast 3G), the request never responds. The entities are not created but the server is not thrown or suspended. It just becomes a phantom call to the API.
> 
> Not sure if this is due to the multi-tenant approach / fix?

this was due to timeouts on nodejs http server and request, this is an issue we have uploading files and performing any other long request.

PR #3135, opened 2020-08-27
updateMany for updateLogs now updates in batches

### LINKED ISSUE #3064: updateMany hook for updateLogs produces an error when query updates a lot of entries.
how to replicate:
1. have a collection with a lot of entities belonging to a template (do not know exactly the number in my case was around 900k but probably can happen with less than that)

2. go to settings -> templates
3. edit the template, change any name and save.
4. this tries to modify all entities that belong to that template
5. fails due to error during the process of updating the updatelog collection:

```
MongoError: distinct too big, 16mb cap
    at Connection.<anonymous> (/Users/daneryl/workspace/uwazi/node_modules/mongodb/lib/core/connection/pool.js:466:61)
    at Connection.emit (events.js:198:13)
    at processMessage (/Users/daneryl/workspace/uwazi/node_modules/mongodb/lib/core/connection/connection.js:364:10)
    at Socket.<anonymous> (/Users/daneryl/workspace/uwazi/node_modules/mongodb/lib/core/connection/connection.js:533:15)
    at Socket.emit (events.js:198:13)
    at addChunk (_stream_readable.js:288:12)
    at readableAddChunk (_stream_readable.js:269:11)
    at Socket.Readable.push (_stream_readable.js:224:10)
    at TCP.onStreamRead [as onread] (internal/stream_base_commons.js:94:17)
```

daneryl:
fixes #3064 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/odm/model.ts
+++ b/app/api/odm/model.ts
@@ -1,15 +1,33 @@
+//@ts-ignore
+import PromisePool from '@supercharge/promise-pool';
 import mongoose, { Schema, UpdateQuery, ModelUpdateOptions } from 'mongoose';
-
 import { model as updatelogsModel } from 'api/updatelogs';
 
-import { OdmModel, WithId, models, UwaziFilterQuery } from './models';
+import { WithId, models, UwaziFilterQuery } from './models';
 import { MultiTenantMongooseModel } from './MultiTenantMongooseModel';
 
 const generateID = mongoose.Types.ObjectId;
 export { generateID };
 
-class UpdateLogHelper {
+const getBatchSteps = async <T>(
+  model: OdmModel<T>,
+  query: UwaziFilterQuery<T>,
+  batchSize: number
+): Promise<T[]> => {
+  const allIds = await model.get(query, '_id', { sort: { _id: 1 } });
+
+  const steps = [];
+  for (let i = 0; i < allIds.length; i += batchSize) {
+    steps.push(allIds[i]);
+  }
+
+  return steps;
+};
+
+export class UpdateLogHelper<T> {
   collectionName: string;
+
+  static batchSizeUpsertMany = 50000;
 
   constructor(collectionName: string) {
     this.collectionName = collectionName;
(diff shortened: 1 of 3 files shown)


app/api/odm/model.ts
```diff
@@ -28,33 +46,48 @@ class UpdateLogHelper {
...
+    batchSize = UpdateLogHelper.batchSizeUpsertMany
+  ) {
+    await new PromisePool()
+      .for(await getBatchSteps(models[this.collectionName], query, batchSize))
+      .withConcurrency(5)
+      .process(async (stepIndex: T) => {
+        const batch = await models[this.collectionName].get(
+          { ...query, $and: [{ _id: { $gte: stepIndex } }] },
+          { _id: 1 },
+          { limit: batchSize }
```
RafaPolit, line 61: I have a couple of questions here:
- since we have all the ids already, is it more performant to use the limit than it would be to do a `@gte` AND `@lt` ?
- why is this $and required here?  is it not just "id greater than stepIndex"?  Is it because there could be also an _id key in the query already? (probably the case, just confirming)

app/api/odm/model.ts
```diff
@@ -28,33 +46,48 @@ class UpdateLogHelper {
...
+        );
+
+        await updatelogsModel._updateMany(
+          { mongoId: { $in: batch }, namespace: this.collectionName },
+          { $set: { timestamp: Date.now(), deleted } },
```
RafaPolit, line 66: Is this true that an entire batch is either deleted true or deleted false?  Seems like a very weird approach.  I understand that this is also on the original code, that is why I have doubt.

app/api/odm/model.ts
daneryl: > * since we have all the ids already, is it more performant to use the limit than it would be to do a `@gte` AND `@lt` ?


I dont think i understand the question, i do not think limit has a performance overhead, normally this problem appears using offset and limit.


> * why is this $and required here?  is it not just "id greater than stepIndex"?  Is it because there could be also an _id key in the query already? (probably the case, just confirming)


yes this is the case, actually it was the case in the test itself, that is why i had to add the $and 🙈

app/api/odm/model.ts
daneryl: yes this is true, we have no operations that use this hooks with a mixed update/create/delete, actually i do not know if this is possible using mongo either but definitely not our case.

PR #3145, opened 2020-09-03
remove metadata ghost properties from entities

### LINKED ISSUE #3141: Migration required for "ghost" metadata
Previously, we did not sanitize the metadata.  Upon deleting a property, the metadata value for existing entities would remain as data of the entity.

Now, if you delete a property, entity's metadata gets sanitized and that property is removed (as should be).

Still, we never migrated the metadata to delete those ghost values.  The problem is that, for old collections that already have this ghost metadata, if you try to update the template, resaving the entities causes failed validations, because you are sending properties that are no longer supported by the template.

There is a "workaround" which is to recreate the template property, save the template, and now delete it once more.  Now, our sanitizing procedure works and deletes the entity's metadata key.

Still, this is very obscure and convoluted, not to mention really undecipherable for a non-tech user.

We need to migrate the entities to be sanitized and free of these ghost values.

daneryl:
fixes #3141

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/migrations/migrations/28-remove_not_allowed_metadata_properties/index.js
+++ b/app/api/migrations/migrations/28-remove_not_allowed_metadata_properties/index.js
@@ -1,0 +1,41 @@
+/* eslint-disable no-await-in-loop */
+export default {
+  delta: 28,
+
+  name: 'remove_not_allowed_metadata_properties',
+
+  description:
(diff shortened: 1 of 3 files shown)


mfacar (approved):
LGTM

PR #3156, opened 2020-09-07
fixed bug of multiple request state in Library

### LINKED ISSUE #3144: Library's "Load More" is duplicating requests with wrong limits
After #3108 , there are 4 calls to the API for each "load more" hit:
- The first two correctly request 'from 30 to 60' for both populated rows and geolocation values... but then
- Makes 2 requests with "from 0 to 60" with rows and geolocation

This second request should only happen on server side rendering.

This can be a side effect of the changes to the componentWillMount processes?

cc @konzz

konzz:
fixes #3144 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti:
@konzz It looks like the issue is also happening in the Uploads. It is not blocking Tableview but might it still be nice to fix?
As for the rest, the code looks good to us (@mfacar and me) and on the library it is working like charm :)

RafaPolit:
@konzz please extend the fix to the uploads as part of this same PR.

mfacar (approved):
LGTM

PR #3180, opened 2020-09-14
Css fix for copy-from panel

### LINKED ISSUE #3160: "Copy From" css not working correctly
Currently, the css of the side panels assigns widths depending on original contents.  This can cause the right pane to be as small as 80px if the options for selects and other properties occupies wider widths.

The width of the each individual side panel needs to be fixed as well.

Here's an example of the problem:
copyfromissue

konzz:
fixes #3160 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/App/scss/modules/_document.scss
+++ b/app/react/App/scss/modules/_document.scss
@@ -295,6 +295,7 @@ mark.searchTerm {
 
 .copy-from {
   flex: 1;
+  min-width: 50%;
 
   .copy-from-buttons {
     position: absolute;


app/react/App/scss/modules/_document.scss
```diff
@@ -295,6 +295,7 @@ mark.searchTerm {
 
 .copy-from {
   flex: 1;
+  min-width: 50%;
```
RafaPolit, line 298: I'm not sure I understand the reasoning behind this combo.  Could the copy-from pane be bigger than 50%?  If not, this should not be a `flex: 1` with `min-width`, it should be a simpler `width: 50%`.  Maybe this is for responsive behavior?

konzz: it can be in mobile

PR #3182, opened 2020-09-14
[HOTFIX] formatMessage takes into account current tenant

daneryl:
Error formatter now uses tenant name to identify the error, when tenant name is "default" then the current DB name is used for backwards compatibility.

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/log/formatMessage.js
+++ b/app/api/log/formatMessage.js
@@ -1,7 +1,14 @@
+import { tenants } from 'api/tenants';
+import { config } from 'api/config';
+
 export default function formatMessage(info, instanceName) {
   const message = info.message && info.message.join ? info.message.join('\n') : info.message;
 
-  const result = `${info.timestamp} [${instanceName}] ${message}`;
+  let tenantName = instanceName;
+  if (info.shouldBeMultiTenantContext) {
+    const tenant = tenants.current();
+    tenantName = tenant.name === config.defaultTenant.name ? instanceName : tenant.name;
+  }
 
-  return result;
+  return `${info.timestamp} [${tenantName}] ${message}`;
 }
--- a/app/api/tenants/tenantContext.ts
+++ b/app/api/tenants/tenantContext.ts
@@ -16,13 +16,11 @@ export type Tenant = {
 class Tenants {
   storage = new AsyncLocalStorage<string>();
 
-  defaultTenantName = 'default';
-
   tenants: { [k: string]: Tenant };
 
   constructor() {
     this.tenants = {
-      [this.defaultTenantName]: { name: this.defaultTenantName, ...config.defaultTenant },
+      [config.defaultTenant.name]: config.defaultTenant,
     };
   }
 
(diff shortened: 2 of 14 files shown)


app/api/log/formatMessage.js
```diff
@@ -1,7 +1,17 @@
+import { tenants } from 'api/tenants';
+import { config } from 'api/config';
+
 export default function formatMessage(info, instanceName) {
   const message = info.message && info.message.join ? info.message.join('\n') : info.message;
 
-  const result = `${info.timestamp} [${instanceName}] ${message}`;
+  const tenant = tenants.current();
+  let tenantName = tenant.name;
```
txau, line 8: Maybe refactor to a ternary conditional?

txau (approved):

app/api/log/formatMessage.js
daneryl: done 👍

app/api/tenants/tenantContext.ts
```diff
@@ -44,7 +42,7 @@ class Tenants {
 
   async run(cb: () => Promise<void>, tenantName?: string): Promise<void> {
     return new Promise((resolve, reject) => {
-      this.storage.run(tenantName || this.defaultTenantName, () => {
+      this.storage.run(tenantName || config.defaultTenant.name, () => {
```
RafaPolit, line 45: Wondering if this should allow tenantName to be empty and use the conditional in the run, or if it just makes sense to default the function argument to the config value if undefined.  Perhaps not a lot of difference?

app/api/tenants/tenantContext.ts
```diff
@@ -16,13 +16,11 @@ export type Tenant = {
...
   storage = new AsyncLocalStorage<string>();
 
-  defaultTenantName = 'default';
-
   tenants: { [k: string]: Tenant };
 
   constructor() {
     this.tenants = {
-      [this.defaultTenantName]: { name: this.defaultTenantName, ...config.defaultTenant },
+      [config.defaultTenant.name]: config.defaultTenant,
```
RafaPolit, line 23: Why do we force to always have a "default" tenant?  If I configure two named tenants, would this keep ALSO a "default" tenant that no one asked for?

daneryl: probably yes, can you create a separate issue to discuss this if you think it merits as it has nothing to do with the hotfix.
as a possible solution we can set the tenants array to blank if there is some tenant coming from the tenants change event.

app/api/tenants/tenantContext.ts
daneryl: done 👍

daneryl:
moved to draft, the usage of errorHandler on server initialization prevents this patch to work properly, errors are not properly handled because there is not tenant context here.
https://github.com/huridocs/uwazi/blob/3feb3b92588fabbfe87155654ef2c009b430ef59/app/server.js#L41-L48

daneryl:
To solve the issue there is a new internal flag for errorLog that indicates the current execution should be a inside a tenant async context, this is easy to know from handleError, if there is a req parameter, then we should be inside a tenant async context, in any other case it will behave like before.

RafaPolit:
LGTM

PR #3193, opened 2020-09-22
Added the with-header flag where needed

### LINKED ISSUE #3147: Document (PDF) view pane is not configured correctly, bottom part never visible
The Document Viewer, when displaying a PDF, is calculating it's height incorrectly.  The lower scroll arrow is below the screen end, and you cannot see the bottom part of the content even after you have scrolled all the way down.

This is also true for the uploads section.

It appears that the inclusion of the bar underneath the navigation is affecting this issue.

txau:
fixes #3147 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Viewer/components/Viewer.js
```diff
@@ -96,6 +96,8 @@ export class Viewer extends Component {
       className += ' connections';
     }
 
+    className += ' with-header';
```
RafaPolit, line 99: There is no need for this extra line, line 87 should read:
`let className = 'document-viewer with-header';`

RafaPolit (changes requested):

PR #3194, opened 2020-09-22
Now the color is coming directly from db, since it was being already …

### LINKED ISSUE #3185: Template Colors are not getting saved

txau:
…stored

fixes #3185 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Templates/components/MetadataTemplate.tsx
```diff
@@ -35,12 +35,12 @@ interface MetadataTemplateProps {
...
-  if (!templateId) {
-    return COLORS[allTemplates.size % COLORS.length];
+const getTemplateDefaultColor = (allTemplates: List<TemplateSchema>, template: any) => {
+  let color = COLORS[allTemplates.size % COLORS.length];
+  if (template.data.color) {
+    color = template.data.color;
   }
-  const index = allTemplates.findIndex((tpl: any) => tpl.get('_id') === templateId);
-  return COLORS[index % COLORS.length];
+  return color;
```
RafaPolit, line 43: This change in approach produces a weird behavior:
- Upon blank-state, Default has no color
- Upon entering Default template it inherits takes Color [1] (allTemplates.size is 1), so Color [0] is never used
- DO NOT SAVE!
- Add a new template, it still gets Color [1] (there is still only one template), so, Default Template and New Template share the same color, in theory
- Go back to Default template: now it has Color [2]

This can be repeated over and over again, because as long as you don't save Default Template, it does not get a color.

The quick solution here is to update the blank state fixtures to actually set the default template to Color[0] (not 1!) and then everything works with this approach.

I believe this is also reflected in the updated spec that shows the same color for both specs, something that wasn't at all the expected behavior.

RafaPolit (changes requested):

app/react/Templates/components/MetadataTemplate.tsx
mfacar: Just as a very small thing, maybe you could invert the order of the condition to avoid calculating the color when it is not needed.

app/react/Templates/components/MetadataTemplate.tsx
```diff
@@ -35,12 +35,12 @@ interface MetadataTemplateProps {
   _id?: string;
 }
 
-const getTemplateDefaultColor = (allTemplates: List<TemplateSchema>, templateId: string) => {
-  if (!templateId) {
-    return COLORS[allTemplates.size % COLORS.length];
+const getTemplateDefaultColor = (allTemplates: List<TemplateSchema>, template: any) => {
+  let color = COLORS[allTemplates.size % COLORS.length];
+  if (template.data.color) {
```
mfacar, line 40: Just as a very small thing, maybe you could invert the order of the condition to avoid calculating the color when it is not needed.

app/react/Templates/components/MetadataTemplate.tsx
txau: I added the color to the default template of the blank state. Also there was an unused copy of the blank state database in the root of the project. I removed it.

app/react/Templates/components/MetadataTemplate.tsx
txau: Refactored as suggested.

mfacar (approved):
LGTM

PR #3195, opened 2020-09-22
add missing accessibility labels to various elements

### LINKED ISSUE #2910: Accessibility improvements
- [ ] Use full language names instead of two-letter abbreviations.  
- [ ] If above not possible, use ARIA tittles to include the full name of the language, as screen readers read: "EN", "ES", "PT", which makes no sense
- [ ] Overhaul review that screen readers are reading primary pages correctly for most of the content.

elric-wamugu:
fixes #2910

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Metadata/components/MetadataFormButtons.js
```diff
@@ -19,13 +19,15 @@ export class MetadataFormButtons extends Component {
     const data = this.props.data.toJS();
 
     const ViewButton = (
-      <I18NLink to={`entity/${data.sharedId}`}>
-        <button type="button" className="edit-metadata btn btn-primary">
```
elric-wamugu, line 23: @RafaPolit @daneryl @konzz Any idea why this button was wrapped with an anchor `a` tag?

daneryl: no idea, this is not needed, the change u made should be good.

app/react/Documents/components/DocumentSidePanel.js
```diff
@@ -196,7 +197,7 @@ export class DocumentSidePanel extends Component {
             })()}
             <li>
               <TabLink to="metadata" default>
-                <Icon icon="info-circle" />
+                <Icon icon="info-circle" aria-label={t('System', 'Info')} />
                 <span className="tab-link-tooltip">{t('System', 'Info')}</span>
```
daneryl, line 201: @elric-wamugu this case is weird, the System text is already on the screen in the span below the Icon, why do we need the aria-label in this case ?
will this be a good usage for [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute#:~:text=Description,in%20a%20space%20delimited%20fashion.) ??

app/react/Entities/components/EntityViewer.js
```diff
@@ -191,13 +192,23 @@ export class EntityViewer extends Component {
             >
               <ul className="nav nav-tabs">
                 <li>
-                  <TabLink to="info">
+                  <TabLink
+                    to="info"
+                    role="button"
+                    tabIndex="0"
+                    aria-label={t('System', 'Info', null, false)}
```
daneryl, line 199: already on screen on the span child.

daneryl (changes requested):
@elric-wamugu Looks good, the only thing i want you to review is the cases where you add aria-label but the text appears to be on the screen already, i have commented in a couple of them.

app/react/Documents/components/DocumentSidePanel.js
elric-wamugu: The reason the span text does not work is because the `tooltip` class gives it a `display: none` style rule which is then omitted by the screen reader. Putting it on the `Icon` component doesn't work as well because the SVG generated has an `aria-hidden="true"` attribute.

app/react/Entities/components/EntityViewer.js
elric-wamugu: The span text does not work is because the `tooltip` class gives it a display: none style rule which is then omitted by the screen reader. Easier solution is to make the parent div focusable using `tabIndex="0"` and registering it as a button then giving it an aria label.

daneryl:
@elric-wamugu fix eslint errors pls and ill merge LGTM

elric-wamugu:
@daneryl for the eslints errors, I've had to make a few changes to the files upload input buttons to make them both:

1. file input label readable by the screen readers,

2. on file input button focus (via keyboard tab key), user can press the keyboard `return/enter` or `spacebar` keys to open the OS file select window which is the native default behavior on interactive elements. The style rule `display: none` that had been applied to the file input buttons would make the screen readers skip those uploads buttons entirely such that the user can even select a file by pressing the `return/enter` or `spacebar` keyboard keys,

while still retaining the intended styling of the buttons.

PR #3202, opened 2020-09-22
3197 tableview stabilization

### LINKED ISSUE #3197: Ensure table view shows all fields correctly
Some fields like Rich Text are not showing correctly in the table view.

Please ensure correct display of all types of fields, creating a dummy template with all possible options.

This should be a release stabilization development, please fork from Release.

### LINKED ISSUE #3204: Table view column selector not refreshing after template filtering
The column selector in table view should show columns only for the templates selected on the filters panel.

mfacar:
fixes #3197
fixes #3204 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


e2e/suites/tableView.test.ts
```diff
@@ -20,6 +20,7 @@ describe('Table view', () => {
   describe('Table actions', () => {
     it('Should show only selected properties', async () => {
       await page.click('.hidden-columns-dropdown');
+      await page.waitFor(200);
```
RafaPolit, line 23: If we are waiting for `#rw_1_listbox li`, why do we need to wait for 200 ms here?  If the wait for the `li` is not good enough, please, input the correct selector that 100% disambiguates the status of the client, and not this arbitrary wait.

RafaPolit (changes requested):

app/react/Library/helpers/requestState.js
```diff
@@ -80,7 +80,11 @@ export default function requestState(request, globalResources, calculateTableCol
         }),
       ];
       if (calculateTableColumns) {
-        const tableViewColumns = getTableColumns(documents, templates);
+        const tableViewColumns = getTableColumns(
+          documents,
+          templates,
+          documentsRequest.data.types || []
```
RafaPolit, line 86: As a comment: these type of defaults are usually best configured as default for the function's arguments to prevent having to create this conditionals every time you call the function.

PR #3206, opened 2020-09-24
Elastic library upgrade

### LINKED ISSUE #3065: Add USE_ELASTIC_ICU env variable to README

### LINKED ISSUE #3068: Support Elastic 7.8

konzz:
fixes #3068
fixes #3065

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/search/search.js
```diff
@@ -650,7 +650,9 @@ const instanceSearch = elasticIndex => ({
     // queryBuilder.query() is the actual call
     return elastic
       .search({ index: elasticIndex || getCurrentTenantIndex(), body: queryBuilder.query() })
-      .then(response => processResponse(response, templates, dictionaries, language, query.filters))
+      .then(response => {
+        return processResponse(response, templates, dictionaries, language, query.filters);
+      })
```
RafaPolit, line 655: This is yielding a warning, and it looks like it was intercepted to do some debugging, but it should be reverted to the arrow function directly returning the processResponse as the warning suggests.

RafaPolit:
This PR should also include an update of the README.  I understand this is NOT backwards compatible with 7.6, right?
Should we include checks on our Server.js to report Node, Mongo and ElasticSearch versions?

txau:
> This PR should also include an update of the README. I understand this is NOT backwards compatible with 7.6, right?
> Should we include checks on our Server.js to report Node, Mongo and ElasticSearch versions?

It needs some README update. But it may be compatible with previous versions. I hope @konzz can confirm.

txau:
For this PR:
- Is it confirmed that we don't need the --max-http-header-size 20000 flag anymore?
- Recover the ICU part for the README
- Confirm if this module is backwards compatible with Elastic 7.6
- Make sure our CI scripts are using ES 7.9.1

konzz:
I didn't upgrade elastic to 7.9
This PR is just "Elastic library upgrade" as the title says, but if you want I can also check this with elastic 7.9 and add that to the PR.

README.md
```diff
@@ -26,16 +25,19 @@ Read the [user guide](https://github.com/huridocs/uwazi/wiki)
 Before anything else you will need to install the application dependencies:
 
 - **NodeJs 14.6.x** For ease of update, use nvm: https://github.com/creationix/nvm.
-- **ElasticSearch 7.6.2** https://www.elastic.co/guide/en/elasticsearch/reference/7.6/install-elasticsearch.html Please note that ElasticSearch requires java.  Follow the instructions to instal the package manually, and install v7.6.2.  v7.7.x and above are known to cause `HPE_HEADER_OVERFLOW` errors.  Please, ensure v7.6.2 only. You also probably need to disable ml module in the ElasticSearch config file:
+- **ElasticSearch 7.9.1** https://www.elastic.co/guide/en/elasticsearch/reference/7.6/install-elasticsearch.html Please note that ElasticSearch requires java. Follow the instructions to instal the package manually, you also probably need to disable ml module in the ElasticSearch config file:
```
RafaPolit, line 28: The link needs to be updated as well, it reads /7.6/

RafaPolit (changes requested):

txau:
LGTM. Too bad the header overflow error is still present. I guess we have to live with the CLI flag for now.

PR #3208, opened 2020-09-24
3109 copyfrom ux glitches

### LINKED ISSUE #3109: Copy-from UX glitches
- When using it in the side-panel, if you have scrolled down already, you get a blank side panel and you don't see the search box.
- It shows the preview which takes a lot of real state. Maybe this is good for easily identifying the source document to copy from but it takes a lot of real state with info that won't be copied.
- The button label says "copy highlighted". Since there are highlighted fields in both sides it can be misleading when a field is highlighted on the left side (target) but not present on the right side (source). If the point is to tell the user that the field will be overwritten with no values maybe we could use some other visual code.
- Some multi-select fields are overwritten with a blank value instead of no value selected:
image

fnocetti:
fixes #3109 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti:
@konzz There is one codeclimate issue that I think should not be tackled as it implies refactoring a pretty core selector. Let me know what you think.

fnocetti:
Comment for reference: This is ready for review.
cc: @konzz @RafaPolit

Issue #2934, opened 2020-05-26
Define how to test AJV Route Validations


RafaPolit:
AJV Definitions are already tested, still there is some need to in-route test that, at least, it is pointing to the right validation.  If it is self-included in the route, this still need some sort of integration test.

RafaPolit:
We are moving away from AJV async validations on the Routes.  Validation for data consistency happen in the application to avoid having to pass through the route for internal data validation.   These are tested by passing a JSON structure and checking for expected error.

The idea is to start moving these validations as the modules get addressed.

Backend will assess if there are routes affected by this that need to be tested (rather than migrated at this point).

cc @fnocetti @LaszloKecskes

PR #3228, opened 2020-09-29
2953 activitylog sanitization

### LINKED ISSUE #2879: Exclude api/documents/download and api/attachments/download from AuditLog
This routes should not be logged as they do not modify the data and pollute the audit log, as well as grow the log considerably.

### LINKED ISSUE #2953: Activity log sanitization
At the moment our logs just grow with no clean up strategy whatsoever.

### LINKED ISSUE #3094: Do not log in updateLogs the activityLog entries

mfacar:
fixes #2953
fixes #2879
fixes #3094

This PR includes:

- [ ] Exclusion of  API endpoints that we don't want logging
- [ ] Refactor of ActivityLogParser and addition of decorators for the raw entries that we want to keep
- [ ] Migration to get rid of the already logged raw data
- [ ] Saving activitylog entries both in mongodb and filesystem as well under `log` folder.
- [ ] TTL index to expire activitylog entries a year after the date of creation
- [ ] Prevent creating UpdateLog for activitylog collection
- [ ] Searching for anonymous entries at Activity Log page

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/activitylog/activitylogMiddleware.js
```diff
@@ -1,19 +1,27 @@
...
+export const IGNORED_ENDPOINTS = [
+  '/api/users',
+  '/api/login',
+  '/api/contact',
+  '/api/unlockaccount',
+  '/api/resetpassword',
+  '/api/recoverpassword',
+  '/api/documents/pdfInfo',
+  '/api/documents/download',
+  'api/attachments/download',
```
RafaPolit, line 14: Please include the forward slash

.gitignore
```diff
@@ -29,6 +29,7 @@ app/api/csv/specs/*.jpg
 app/api/attachments/specs/uploads/toDelete.txt
 uploaded_documents/*
 custom_uploads/*
+activity_log/*
```
txau, line 32: Why adding a new folder when we already have a `log` folder that is already configured in the paths for production? I don't think this is a bad idea, at the same time this forces our production code and self-installed instances to manage this new variable when we release this code.

app/api/activitylog/specs/activitylog.spec.js
```diff
@@ -31,7 +31,7 @@ describe('activitylog', () => {
       expect(entries.length).toBe(5);
```
txau, line 31: From the description of the text ` should return the entries, excluding certain routes`. I think we have just one special case `{ $ne: '/api/semantic-search/notify-updates' }`. IMO we should not log these entries nor filter them.

app/api/attachments/routes.js
```diff
@@ -96,6 +97,7 @@ export default app => {
     '/api/attachments/upload',
     needsAuthorization(['admin', 'editor']),
     uploadMiddleware(attachmentsPath),
+    activitylogMiddleware,
     validation.validateRequest(
       Joi.object()
```
txau, line 102: Maybe take the time to switch from Joi to AJV? In order to get rid of some tech debt with the boy-scout approach.

app/api/activitylog/helpers.js
```diff
@@ -8,66 +12,158 @@ export const methods = {
...
+  ...(builder.name && { name: builder.name }),
+  ...(builder.extra && { extra: builder.extra }),
+});
+
+export class ActivityLogBuilder {
+  constructor(data, entryValue) {
+    this.description = entryValue.desc;
+    this.data = data;
+    this.action = entryValue.method ? entryValue.method : methods.update;
+    this.beautified = true;
```
txau, line 28: Do we actually need this flag? As far as I'm understanding this flag is used for controlling rendering on client side. Can't we beautify all raws, even the raw ones so we don't have the condition and get cleaner code and a more straightforward approach?

app/api/activitylog/specs/activitylog.spec.js
```diff
@@ -47,53 +47,61 @@ describe('activitylog', () => {
...
-    it('should filter by method', async () => {
-      const { rows: entries } = await activitylog.get({ method: ['POST'] });
-      expect(entries.length).toBe(1);
-      expect(entries[0].method).toBe('POST');
-    });
+    describe('filtering', () => {
+      it('should filter by method', async () => {
+        const { rows: entries } = await activitylog.get({ method: ['POST'] });
+        expect(entries.length).toBe(1);
+        expect(entries[0].method).toBe('POST');
```
txau, line 54: Maybe simplify this code to a foreach so that we don't need to know how many there are, all we need to know is that all of them are POST. This way if the fixtures change the condition is better covered and still met.

app/api/activitylog/specs/activitylog.spec.js
```diff
@@ -47,53 +47,61 @@ describe('activitylog', () => {
       });
     });
 
-    it('should filter by method', async () => {
-      const { rows: entries } = await activitylog.get({ method: ['POST'] });
-      expect(entries.length).toBe(1);
-      expect(entries[0].method).toBe('POST');
-    });
+    describe('filtering', () => {
```
txau, line 50: Maybe switch the style to a more semantic approach, ie:
```
describe('when filtering')
  it('should return results by method')
```
So that it reads "when filtering it should return results by method".

.gitignore
mfacar: Given the activity log files include in its name the tenant I'll write them into the log folder.

app/api/migrations/migrations/29-activity_log_sanitization/index.js
```diff
@@ -0,0 +1,37 @@
...
+    );
+
+    const deletedUploadEntriesWithoutBody = await db
+      .collection('activitylogs')
+      .deleteMany({ url: { $in: BODY_REQUIRED_ENDPOINTS }, body: '{}' });
+    process.stdout.write(
+      `${deletedUploadEntriesWithoutBody.result.n} activity log POST entries deleted with empty bodies\r\n`
+    );
+
+    await db.collection('activitylogs').createIndex({ expireAt: 1 }, { expireAfterSeconds: 0 });
```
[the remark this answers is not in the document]
mfacar: You're right, we only need to have into account that there is not a file that contains this older logs

app/api/activitylog/specs/activitylog.spec.js
mfacar: Sure, I didn't realize that. This test is not needed anymore given we are not logging this special case and we'll delete the old entries.

app/api/attachments/routes.js
mfacar: Although to move the validation to AJV is easy, to update the validation tests is not being straightforward, maybe it could be better to change this when we have a definition for #2934

app/api/migrations/migrations/29-activity_log_sanitization/index.js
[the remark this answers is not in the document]
txau: You are right, I think it is ok to sacrifice that part of the logs. The activity log feature was first deployed on September 2019, so the users will be losing record for just a few days.

app/api/activitylog/helpers.js
mfacar: I've removed the beautified flag adding de RAW mark at the parsing, now all entries have the same structure on the client-side.

txau (approved):
LGTM

app/api/activitylog/activitylogParser.js
```diff
@@ -1,245 +1,180 @@
...
-  'POST/api/semantic-search': semanticSearchPOST,
-  'DELETE/api/semantic-search': generateDeleteBeautifier('semantic search', 'searchId'),
-  'POST/api/semantic-search/stop': generateSemanticSearchUpdateBeautifier(
-    'Stopped semantic search'
-  ),
-  'POST/api/semantic-search/resume': generateSemanticSearchUpdateBeautifier(
-    'Resumed semantic search'
-  ),
-  MIGRATE: migrationLog,
+import * as helpers from 'api/activitylog/helpers';
```
fnocetti, line 1: @mfacar this file was basically rewritten, I think it's a good opportunity to make it TS. You think that's relatively quick to accomplish?

app/api/migrations/migrations/29-activity_log_sanitization/specs/29-activity_log_sanitization.spec.js
```diff
@@ -0,0 +1,78 @@
...
+      expect(ignoredEntries).toBe(0);
+    });
+
+    it('should keep upload activity entries with body', async () => {
+      const ignoredEntries = await testingDB.mongodb
+        .collection('activitylogs')
+        .find({ url: { $in: BODY_REQUIRED_ENDPOINTS }, body: { entityId: 'entity1' } })
+        .count();
+
+      expect(ignoredEntries).toBe(0);
```
fnocetti, line 65: Shouldn't we have at least one result to ensure the migration is keeping those records?

app/api/utils/specs/date.spec.js
```diff
@@ -25,4 +25,11 @@ describe('date helper', () => {
       );
     });
   });
+
+  describe('add years to current date', () => {
+    const today = new Date();
```
fnocetti, line 30: I think hardcoded dates could be "more predictable" and prevent possible flakinesses? What do you think?

app/api/odm/model.ts
```diff
@@ -138,7 +74,7 @@ export class OdmModel<T> {
 }
 
 export function instanceModel<T = any>(collectionName: string, schema: mongoose.Schema) {
-  const logHelper = new UpdateLogHelper<T>(collectionName);
+  const logHelper = createUpdateLogHelper(collectionName);
```
fnocetti, line 77: This is missing the generic type T

fnocetti (changes requested):
@mfacar I still need to review the logHelper and its tests but I'm leaving some comments as for now. Please take a look! Thank you!

app/api/odm/model.ts
mfacar: Great, good catch!

app/api/utils/specs/date.spec.js
mfacar: I'm using this utility method specifically to add to a year to the current date, I'm not sure how to hardcoded it in tests.
I could change the function passing a fixed date and then add it 1 year fixed, but I don't know if it could be useful in other contexts.

app/api/migrations/migrations/29-activity_log_sanitization/specs/29-activity_log_sanitization.spec.js
mfacar: Actually, this is the intention of this test, but there were a couple of errors, thanks I didn't notice it.

app/api/migrations/migrations/29-activity_log_sanitization/specs/29-activity_log_sanitization.spec.js
```diff
@@ -0,0 +1,78 @@
...
+      expect(remainingEntries).toBe(3);
+    });
+
+    it('should remove activity entries with unwanted urls', async () => {
+      const ignoredEntries = await testingDB.mongodb
+        .collection('activitylogs')
+        .find({ url: { $in: IGNORED_ENDPOINTS } })
+        .count();
+
+      expect(ignoredEntries).toBe(0);
```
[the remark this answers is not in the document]
mfacar: Sure, I've added some additional fixtures and move into a different test the assert of remaining entries.

app/api/activitylog/specs/activitylogParser.spec.js
```diff
@@ -35,14 +35,17 @@ describe('Activitylog Parser', () => {
     const semanticData = await getSemanticData(log);
     expect(semanticData).toEqual({
       ...expected,
-      beautified: true,
     });
   }
 
   describe('getSemanticData', () => {
     it('should report as beautified: false if no translation present for the route', async () => {
```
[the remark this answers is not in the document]
mfacar: done, I've changed the description with RAW

app/api/activitylog/helpers.js
```diff
@@ -8,66 +12,156 @@ export const methods = {
...
 };
 
+const buildActivityLogEntry = builder => ({
+  description: builder.description,
+  action: builder.action || methods.update,
+  ...(builder.name && { name: builder.name }),
+  ...(builder.extra && { extra: builder.extra }),
+});
+
+export class ActivityLogBuilder {
```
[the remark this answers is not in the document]
mfacar: done

app/api/activitylog/activitylogParser.js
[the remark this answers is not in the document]
mfacar: I've changed to ts, only that I had to include several `any` in some places because the content of the body/query could be anything

fnocetti:
I finished the Code Review and it LGTM. I will proceed with the testing today!

fnocetti:
@mfacar I tested several cases and everything works well. I have just two comments:
- There is no migration for the updateLogs. For example, it is still showing the activityLogs updates.
- The activitylogs for public forms (anonymous) are not registering any body, is that on purpose?

If the previous points are desired, then this is ready to be merged in my opinion!

fnocetti:
@mfacar Sorry, I just noted one more thing: in the file where the logs are written there's no separator between entries, their written one after the other in a single line. I think we should have a separator in case we actually need to parse it.

mfacar:
> * There is no migration for the updateLogs. For example, it is still showing the activityLogs updates.
I've added a migration to delete these records.
> * The activitylogs for public forms (anonymous) are not registering any body, is that on purpose?
I moved the activitylogmiddleware to the /api/public endpoint and update the decorator in order to have more information, given I'm adding this endpoint to the required body ones, all the previous entries will be deleted in the migration, given these doesn't contain any information, it is part of the sanitization of old entries.

fnocetti (approved):
LGTM! Good job! Merging this now.

PR #3231, opened 2020-10-02
3098 group geolocations

### LINKED ISSUE #3098: Allowing single map view on entity view to compare multiple geolocations
There should be an option (or default to this) to view a single map that shows multiple geolocation properties: both local and inherited.  Keep in mind  that the geolocation field is already prepared to have multiple geolocations itself.  There is no UI for this, but do not break that functionality.

fnocetti:
fixes #3098 

PR checklist
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Metadata/components/GroupedGeolocationViewer.tsx
+++ b/app/react/Metadata/components/GroupedGeolocationViewer.tsx
@@ -1,62 +1,91 @@
 import React from 'react';
 import GeolocationViewer from 'app/Metadata/components/GeolocationViewer';
+import { connect } from 'react-redux';
+import { IStore } from '../../istore';
+import { createSelector } from 'reselect';
+import { ensure } from '../../../shared/tsUtils';
 
 interface GeolocationMarker {
   lat: string;
   lon: string;
   label: string;
+  color: string;
 }
 
 interface GroupMember {
   label: string;
-  translationContext: string;
+  translateContext: string;
   value: GeolocationMarker[];
 }
 
 interface GroupedGeolocationViewerProps {
   members: GroupMember[];
+  colors: {
+    [template:string]: string;
+  };
 }
 
 const notInherited = (member: GroupMember) => member.value.length === 1 && !member.value[0].label;
 
-const renderGroupInfo = (member: GroupMember) => {
+const Pill = (props: {label: string, color: string}) => (<span className="pill" style={{backgroundColor: props.color}}>{props.label}</span>);
+
+const renderGroupInfo = (colors: {
+  [template:string]: string;
+}) => (member: GroupMember) => {
   return notInherited(member) ? (
     <dl>
       <dt></dt>
       <dd>
-          <span className="pill">{member.label}</span>
+          <Pill label={member.label} color={colors[member.translateContext]}/>
       </dd>
     </dl>
   ) : (
     <dl>
       <dt><span>{`${member.label}`}</span></dt>
       <dd>
-        {member.value.map(value => (
-          <span className="pill">{value.label}</span>
-        ))}
+        {member.value.map(value => <Pill label={value.label} color={colors[member.translateContext]} />)}
       </dd>
     </dl>
   );
 };
 
-export const GroupedGeolocationViewer = (props: GroupedGeolocationViewerProps) => {
+const GroupedGeolocationViewerComponent = (props: GroupedGeolocationViewerProps) => {
   const markers = props.members.reduce(
     (flat: GeolocationMarker[], member) => {
       if (notInherited(member)) {
         return flat.concat([{
           ...member.value[0],
-          label: member.label
+          label: member.label,
+          color: props.colors[member.translateContext],
         }]);
       }
 
-      return flat.concat(member.value)
+      return flat.concat(member.value.map(v => ({ ...v, color: props.colors[member.translateContext] })))
     },
     []
   );
+  debugger;
   return (
     <>
       <GeolocationViewer points={markers} onlyForCards={false} />
-      {props.members.map(renderGroupInfo)}
+      {props.members.map(renderGroupInfo(props.colors))}
     </>
   );
 };
+
+const colorsMap = createSelector((s: IStore) => s.templates, templates => {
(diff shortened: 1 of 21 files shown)


app/react/Metadata/components/GroupedGeolocationViewer.tsx
```diff
@@ -0,0 +1,147 @@
...
+      restOfGroups.push(member);
+    }
+  });
+
+  return [getFirstGroupInfo(firstGroup, templatesInfo)].concat(
+    restOfGroups.map(getMultiMemberInfo(templatesInfo))
+  );
+};
+
+// eslint-disable-next-line react/no-multi-comp
```
fnocetti, line 102: I'm getting this eslint error because I splitted the "render function", so now I have more than one function returning JSX and eslint understand I'm defining more than one component in the same file. I technically AM but, do you think it's worth making a brand new component out of something so tightly coupled to the "main one"? Please, let me know your thoughts!!

mfacar:
@fnocetti, please correct me if I'm wrong, I think that in a meeting it was agreed that when there is only a geolocation property, it should show the name of the property only above the map and not as a pill under it.

fnocetti:
> @fnocetti, please correct me if I'm wrong, I think that in a meeting it was agreed that when there is only a geolocation property, it should show the name of the property only above the map and not as a pill under it.

@mfacar my understanding is that when a single field, we do show the property name for the title, but for consistency reasons we still keep the pill. @RafaPolit, could you add here? :D

RafaPolit:
> > @fnocetti, please correct me if I'm wrong, I think that in a meeting it was agreed that when there is only a geolocation property, it should show the name of the property only above the map and not as a pill under it.
> 
> @mfacar my understanding is that when a single field, we do show the property name for the title, but for consistency reasons we still keep the pill. @RafaPolit, could you add here? :D

@fnocetti, indeed, your interpretation is what was agreed upon.  @mfacar 's confusion may come from the fact that I actually voiced that particular point, and I agree with Mercy's comment, I believe that is how this particular case should be solved.  Nonetheless, the agreement was, indeed, to keep both name and pill for consistency.

Let's keep it like we agreed and see what ourselves and other users think of it after we get more used to the pills.

app/react/Metadata/components/GroupedGeolocationViewer.tsx
mfacar: How about moving the three complementary functions within the main function component? This is valid and you wouldn't need to ignore the eslint rule.

app/react/Metadata/components/GroupedGeolocationViewer.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  members.forEach(member => {
+    if (notLabeledOrMultiple(member)) {
+      firstGroup.push(member);
+    } else {
+      restOfGroups.push(member);
+    }
+  });
+
+  return [getFirstGroupInfo(firstGroup, templatesInfo)].concat(
```
mfacar, line 97: If there are no elements in the first group, it is still rendering an empty section `<dl class="pills-container"><dt></dt><dd></dd></dl>`

app/react/Metadata/components/GroupedGeolocationViewer.tsx
```diff
@@ -0,0 +1,147 @@
...
+    <dt>
+      <span>
+        <Translate context={member.translateContext}>{member.label}</Translate>
+        <> (</>
+        <Translate>linked</Translate>
+        <> </>
+        <Translate context={member.translateContext}>
+          {templatesInfo[member.translateContext].name}
+        </Translate>
+        <>)</>
```
mfacar, line 69: Maybe replacing the fragments with something like `<>
{' ('} 
<Translate>` could make this section more readable?

mfacar:
@fnocetti, I've tested some scenarios and it works very well, great job! 
The only thing I'm not sure about is the message while editing a geolocation property: "Grouping geolocation properties together will show them on the same map". It seems like something that the user is doing intentionally and probably he doesn't know how to avoid it.

RafaPolit:
> @fnocetti, I've tested some scenarios and it works very well, great job!
> The only thing I'm not sure about is the message while editing a geolocation property: "Grouping geolocation properties together will show them on the same map". It seems like something that the user is doing intentionally and probably he doesn't know how to avoid it.

I also think the name "Grouping" is misleading, as there is no "grouping" process at all, it's the adjacency that triggers this, not any groupping.

Maybe: "Adjacent geolocation properties will render on the same map."?

Still, I wouldn't go out of the way to 'instruct' the user how to avoid it.  We will get to a better UI further down the road.

app/react/Metadata/components/GroupedGeolocationViewer.tsx
fnocetti: Good catch!! :D

app/react/Metadata/components/GroupedGeolocationViewer.tsx
fnocetti: I had trouble with eslint when using that syntax :/ but I agree it's much preferred. I'll give it a try again!

fnocetti:
@RafaPolit @mfacar I felt wrong about the message too and agree Rafa's one should work better. Thanks!

app/react/Metadata/components/GroupedGeolocationViewer.tsx
fnocetti: Fixed

app/react/Metadata/components/GroupedGeolocationViewer.tsx
fnocetti: I had to add an extra space after the closing parenthesis to work an eslint bug around, but I think it pays the price :)

mfacar:
@txau, I have tested this PR and since everything is correct according to my review, as requested by @fnocetti, I am merging it to have it in the staging environment

PR #3238, opened 2020-10-06
3090 elastic explicit mapping

### LINKED ISSUE #3090: Explicit elastic mapping
@txau @daneryl Please write the process to accomplish this

konzz:
fixes #3090 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


txau:
@konzz I don't understand why most fields are mapped with a `label` property:
image

Can you explain? Or is it that template field labels are being accidentally indexed?

txau:
@konzz is this development covering the use case of changing a property's type? I tried deleting a property in order to change it from text to numeric. It displays an information message saying that the property will be deleted when saving the template. I save the template and try to create a new property with the same name for different type, then it tosses the known exception in elastic's mappings.

konzz:
> @konzz is this development covering the use case of changing a property's type? I tried deleting a property in order to change it from text to numeric. It displays an information message saying that the property will be deleted when saving the template. I save the template and try to create a new property with the same name for different type, then it tosses the known exception in elastic's mappings.

No it is not covering it, the reindex message is needed for that.

txau:
> > @konzz is this development covering the use case of changing a property's type? I tried deleting a property in order to change it from text to numeric. It displays an information message saying that the property will be deleted when saving the template. I save the template and try to create a new property with the same name for different type, then it tosses the known exception in elastic's mappings.
> 
> No it is not covering it, the reindex message is needed for that.

Just the message or the whole workflow?

konzz:
> @konzz I don't understand why most fields are mapped with a `label` property:
> image
> 
> Can you explain? Or is it that template field labels are being accidentally indexed?

Most properties are like:
```
{
   metadata: 
     country: [ { label: 'spain', value: 'a2bdl29d' } ]
}
```
The label is the denormalised value of the id. Thats the label mapping that you see in many properties.

Others like numeric fields shouldn't have that field.

PR #3247, opened 2020-10-13
3213 Accents insensitive search

### LINKED ISSUE #3213: Better searches when using accents
When searching in Spanish (an other languages with accents) it is customary not to type the accents since most search engines ignore them. In our case, searches with the right accents weight more than without them rendering misleading results for users. Ie. a user searching for "fernandez" actually should have typed "fernández", and our result pages are way different when they should be pretty much the same.

Elastic supports several approaches to alleviate this effect. We need to research the option that better balances out cost-effectiveness/accuracy and implement it.

mfacar:
fixes #3213

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/database/elastic_mapping/base_properties.js
+++ b/database/elastic_mapping/base_properties.js
@@ -9,6 +9,7 @@ const properties = {
   title: {
     type: 'text',
     index: true,
+    analyzer: 'other',
     fields: {
       sort: textSorting,
       sayt: { type: 'search_as_you_type' },


txau:
@mfacar while this solution seems to work well, I couldn't find the documentation for the analyzer `other` anywhere. I suppose this is an improvement provided that we were not using any analyzer (or maybe a default analyzer?), so I'm not too concerned about the side effects.

Still, I also tried with the `spanish`analyzer and also handles well accent and case insensitive search. This made me realize that we are using language specific analyzers for the full text properties, but not so much for other properties such as text and rich text fields.

Also, regarding the ICU sorting we are only using it for titles. Maybe it should also be considered for other text fields if they are going to be used for sorting.

@konzz is working on a big overhaul of elasticsearch mappings at the moment, maybe he can absorb these requirements into his development?

txau:
@mfacar LGTM can we merge it?

mfacar:
Sure, I was thinking about adding a test but given it could be reviewed later I think it's fine to merge for now.

PR #3253, opened 2020-10-14
3252 card and row namespacing

### LINKED ISSUE #3252: Namespace cards and rows according to template to provide fine-tuning via CSS
Cards and Rows could be namespaced according to:
- template name
- property name
- etc?

This would allow custom CSSs to do tricks like:
- Paint a Row or a Card with different background depending on their template
- Hide a particular field (Title, for instance!) only on certain types of cards
- Custom style one property based on template + name and not forced to all types of the same source

RafaPolit:
fixes #3252

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Layout/Item.js
+++ b/app/react/Layout/Item.js
@@ -42,7 +42,7 @@ export class Item extends Component {
         <div className="item-snippet">{additionalText}</div>
       </div>
     ) : null;
-    const baseClasName = `item-document template-${doc.template.toString()}`;
+    const baseClasName = `item-document template-${doc.template}`;
     const itemClassName = `${baseClasName} ${this.props.className || ''}`;
     const itemProps = {
       className: itemClassName,
--- a/app/react/Layout/Item.js
+++ b/app/react/Layout/Item.js
@@ -42,7 +42,8 @@ export class Item extends Component {
         <div className="item-snippet">{additionalText}</div>
       </div>
     ) : null;
-    const itemClassName = `item-document ${this.props.className || ''}`;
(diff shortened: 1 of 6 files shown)


txau:
LGTM

RafaPolit:
Some failing e2e tests.  Will check further, changing it to draft for now.

RafaPolit:
Corrected the error.  Ready for QA.

txau:
@RafaPolit for cards we are typing the template id, the field types and the field names. For table we are typing only the template id but not the field types and names? Is this intentional?

RafaPolit:
Yes, it reflects the notion that, probably, you want to access the table view up till the level of the 'row'.  It could be expanded to the level of the cell right now, or further down the road.  I saw less usefulness in reaching the level of cell.  I can see a lot less cases for that.  If you think it's useful, we can go for it now, or leave it for another issue.

txau:
@RafaPolit in that case I guess you added the template class to the TD and not to the TR so according to your explanation it is wrongly placed?

image

RafaPolit:
The row is not aware of the data, so the class is the same for all TDs so you can intervene the "row" level, but at the TD level.

RafaPolit:
> The row is not aware of the data, so the class is the same for all TDs so you can intervene the "row" level, but at the TD level.

Actually, this is not true.  I'll fix this.

PR #3262, opened 2020-10-16
Configured Sync to allow multiple targets

### LINKED ISSUE #3261: Allow sync to multiple targets

RafaPolit:
fixes #3261

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/sync/syncWorker.js
```diff
@@ -29,12 +29,12 @@ export default {
...
 
       if (change.deleted) {
-        return synchronizer.syncData(url, 'delete', change, { _id: change.mongoId });
+        return synchronizer.syncData(url, name, 'delete', change, { _id: change.mongoId });
       }
 
       const data = await config.shouldSync(change);
       if (data) {
-        return synchronizer.syncData(url, 'post', change, data, lastSync);
+        return synchronizer.syncData(url, name, 'post', change, data, lastSync);
```
daneryl, line 37: @RafaPolit i think syncData already had too many arguments before, now five, i think the name can be easily removed if we move the responsibility of updating the lastSync from syncData here after each syncData call, what do you think.

daneryl (changes requested):
a couple of minor comments, apart from that LGTM.

i will also add that when dealing with fixtures this size (the migration ones) found out better for readability to just have them inline on the test.

app/api/sync/specs/syncWorker.spec.js
```diff
@@ -625,13 +635,19 @@ describe('syncWorker', () => {
...
-      const [{ lastSync }] = await syncsModel.find();
-      expect(lastSync).toBe(0);
+      const syncs = await syncsModel.find();
+      expect(syncs.length).toBe(2);
+      expect(syncs).toEqual(
+        expect.arrayContaining([
+          expect.objectContaining({ name: 'slave1', lastSync: 8999 }),
+          expect.objectContaining({ name: 'slave3', lastSync: 0 }),
+        ])
+      );
```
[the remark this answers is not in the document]
RafaPolit: Indeed, it was the unknown resulting order, but yes, good catch, I can sort the query.

app/api/sync/syncWorker.js
RafaPolit: My first hunch was to use the URL as the unique identifying property, but thought name was more semantic.
Still, yes, I can look into who is responsible for what and who needs to know which values.

RafaPolit: I have moved them upwards as you suggested.  Still, the function had 5 args already, so I combined 3 those that 'make sense' to be a data object to an object, and left the other two as options.

I'm not really kin of "masking" the "too many args" into objects, but, in this case, it kind of made sense.  What do you think?

daneryl: looks good 👍, not sure about the object either but definitely better.

daneryl (approved):

PR #3275, opened 2020-10-21
2665 validation of query in searches

### LINKED ISSUE #2665: Improve search query validation and UX
Uwazi currently accepts the Lucene syntax for search queries. A search will fail if Lucene reserved characters are input to a query that doesn't match the syntax. This is particularly inconvenient in searches were every new character triggers a new Search API call (e.g. searching for snippets within a document), which results in a bad user experience since the UI will show an error until the user forms a correct Lucene expression.
We should evaluate:

- If triggering a search for each character is a good approach for those cases and, if it is, evaluate if we need a less disruptive way to inform the error on the UI.
- If we should validate the query and decide how to escape the reserved characters before requesting to the backend. 
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-validate.html might be a good option.

mfacar:
when the query has no valid LUCENE syntax, it does a simple_query_string

fixes #2665

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti (commented):
@mfacar could you please take a look at this old PR's? I think this is the ultimate solution for what we tried to do then. If that the case maybe it would be nice to "revert" those changes:

#2663 #2679

fnocetti (approved):
LGTM!
This is an awesome improvement IMHO!
I'll wait for Mercy to look at the old PRs before merging.

mfacar:
> @mfacar could you please take a look at this old PR's? I think this is the ultimate solution for what we tried to do then. If that the case maybe it would be nice to "revert" those changes:
> 
> #2663 #2679

You're right @fnocetti, the escape method is not needed now. Great catch, I've removed it.

mfacar:
@fnocetti, I'm sorry I introduced an additional change. I was trying some patterns and when I test with * as the search term, search still tossed an error. I assumed that in these cases the search should take the * as an empty search term. Please let me know if this could be a correct approach of if I should revert it.

fnocetti:
@mfacar So weird the simple_string_query fails only for the *. I checked the other few special characters that the simple syntax supports and they do not fail in any of the border scenarios I tried.

That said, the semantic for the * is about prefixing search terms. So I think we should choose between one of the following semantics in this case:
- Search an empty string (the whole set) given we are "searching for terms prefixed with nothing".
- Search a string that literally contains "*". We can achieve that with escaping.

Please, let me know what you think.

cc @RafaPolit

mfacar:
@fnocetti, In fact, simple_string_query isn't failing with *, it is elastic at the moment to return the results, the error says cannot convert undefined or null to object. I think it is ok to keep the first semantic you proposed "searching for terms prefixed with nothing"

RafaPolit:
I am really out of my knowledge here.  I'd like to have @konzz take a look at this and give his feedback.

app/api/search/documentQueryBuilder.js
```diff
@@ -86,18 +86,19 @@ export default function() {
       term,
       fieldsToSearch = ['title', 'fullText'],
       number_of_fragments = 1,
-      type = 'fvh',
-      fragment_size = 300
+      searchTextType = 'query_string'
     ) {
-      if (!term) {
+      if (!term || term === '*') {
```
konzz, line 91: what happens if you put '**' as search term?

mfacar: @konzz, Good catch, I didn't test it, ** is returning the same "cannot convert undefined or null to object", in fact, whatever number of * throws the same error, do you think it is worth to extend the control for these?

mfacar: After debugging to identify the main reason for the error, It turns out that it was a problem with the processing of the response, for some reason when the search is one or more *, returned hits include some without snippets, given we were always expecting them, it was producing an exception, it's fixed now.

PR #3282, opened 2020-10-23
3278 thesaurus rename duplicated in translations

### LINKED ISSUE #3278: Renaming a thesaurus leaves behind that old name in Translations
If a thesaurus has been renamed, it still leaves behind that old text in Translations and it cannot be deleted. Expected behavior would be that when you change the name of a thesaurus, the entry in translation is renamed as well.

In order to delete the old name in Translations, the workaround right now is to temporarily add and save an option with the old name of the thesaurus and then delete it later on. In this way, the old name will be deleted from the translations.

mfacar:
fixes #3278 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/thesauri/specs/thesauri.spec.js
```diff
@@ -277,6 +277,21 @@ describe('thesauri', () => {
...
+          const data = { ...fixtures.dictionaries[1], name: 'new name' };
+          const response = await thesauri.save(data);
+          expect(translations.updateContext).toHaveBeenCalledWith(
+            response._id,
+            'new name',
+            { 'dictionary 2': 'new name' },
+            [],
+            { 'new name': 'new name', 'value 1': 'value 1', 'value 2': 'value 2' },
+            'Dictionary'
+          );
```
daneryl, line 292: @mfacar pls check that the values are actually on the db instead of the toHaveBeenCalled, this kind of mocking has proven worse in the past when changes on the system happens, providing with false positives.

app/api/thesauri/thesauri.js
```diff
@@ -86,7 +86,10 @@ async function updateOptionsInEntities(current, thesauri) {
 
 const update = async thesauri => {
   const currentThesauri = await model.getById(thesauri._id);
-  if (JSON.stringify(thesauri.values) !== JSON.stringify(currentThesauri.values)) {
+  const valuesHaveChanged =
+    JSON.stringify(thesauri.values) !== JSON.stringify(currentThesauri.values);
+  const nameHasChanged = thesauri.name !== currentThesauri.name;
+  if (valuesHaveChanged || nameHasChanged) {
```
daneryl, line 92: @mfacar pleas change this to not use JSON.stringify.

daneryl (changes requested):

app/api/thesauri/thesauri.js
mfacar: Currently, when the order of the thesaurus options has changed, we should update the model because this order of options is taken into account in selectors for example. In this case, it could be a cheap option to keep the comparison with the current approach.

app/api/thesauri/specs/thesauri.spec.js
mfacar: Done! 
I had to move the test because the old describe was using the translations as spy functions.

PR #3284, opened 2020-10-27
3250 delete indexed props

### LINKED ISSUE #3250: Delete indexed properties
When a user deletes a property from a template, it will remain in Elastic's index. If the user then tries to reuse this property with a different data type, it will throw an elastic exception.

The only workaround this at the moment is to recreate the index and reindex the information.

The user should be prompted with a erase index/reindex dialog if:
- The property they are trying to create already exists in the index.
- The property is of another type.
- The property is not actually being use din another template. (Uwazi is already taking care of this one).

konzz:
fixes #3250 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/search/specs/entitiesIndex.spec.js
```diff
@@ -78,4 +86,80 @@ describe('entitiesIndex', () => {
...
+      expect(response).toEqual({ errors: [], valid: true });
+
+      templateB.properties[0].type = 'text';
+      response = await checkMapping(templateB, elasticIndex);
+      expect(response).toEqual({ errors: [{ name: 'Date of birth' }], valid: false });
+    });
+  });
+
+  describe('reindexAll', () => {
+    it('should delete a field from the mapping', async () => {
```
RafaPolit, line 141: This test passes green even if I don't call on `indexEntities` and complete delete line `174` from `entitiesIndex`.

Not the actual return nor the crucial step of indexing all entities is tested at all.

app/api/templates/specs/routes.spec.js
```diff
@@ -196,4 +200,17 @@ describe('templates routes', () => {
         .catch(catchErrors(done));
     });
   });
+
+  describe('/api/templates/check_mapping', () => {
+    it('should check if a template is valid vs the current elasticsearch mapping', done => {
```
RafaPolit, line 205: Please, write new specs with the await - async syntax.

app/react/Templates/components/MetadataTemplate.tsx
```diff
@@ -56,14 +65,29 @@ export class MetadataTemplate extends Component<MetadataTemplateProps> {
...
+    const mappingValidation = await validateMapping(template);
+    if (!mappingValidation.valid) {
+      const fields = mappingValidation.errors.map((e: any) => e.name);
+      this.context.confirm({
+        accept: () => {
+          template.reindex = true;
+          return this.props.saveTemplate(template);
+        },
+        title: 'Template conflict',
+        message: `The field or fields [${fields}] has been previously used for a different type of data,
```
RafaPolit, line 85: When singular and plural are mixed, the plural form must be maintained afterwards: "The field or fields"..."have been"

app/react/Templates/components/specs/MetadataTemplate.spec.js
```diff
@@ -173,14 +177,34 @@ describe('MetadataTemplate', () => {
   });
 
   describe('onSubmit', () => {
-    it('should thrim the properties labels and then call props.saveTemplate', () => {
+    it('should thrim the properties labels and then call props.saveTemplate', async () => {
```
RafaPolit, line 180: `Thrim` is a great word, unfortunately not on the English language :).  I know it was there already, but since you are editing these files, please fix this as well? Thanks.

RafaPolit (changes requested):

PR #3289, opened 2020-10-28
Add favicon config support

### LINKED ISSUE #3089: Request to change/replace the Uwazi favicon
One of our partners has requested if they could change the favicon in their instance and change it to their own logo. 
Would someone from the development team change it for them or would we allow them access to be able to change it?

fnocetti:
fixes #3089 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/App/Root.js
```diff
@@ -34,7 +34,17 @@ const googelFonts = (
...
+  const favicon = reduxData.settings.collection.get('favicon');
+
+  if (!favicon || favicon === '') {
+    return '/public/favicon.ico';
+  }
+
+  return favicon;
+};
+
+export const headTag = (head, CSS, reduxData) => (
```
fnocetti, line 47: Dear reviewer,
I had to export this function to be able to test this feature in the more integrated possible way without spending a huge amount of time. The Root component has no tests at the moment and it requires LOTS of input to be rendered, what would have taken me forever to figure out.
If you think this is not the best approach to follow, please, let me know :)

app/react/Settings/components/CollectionSettings.js
```diff
@@ -145,6 +167,43 @@ export class CollectionSettings extends Component {
               </label>
               <Control.text id="collection_name" model=".site_name" className="form-control" />
             </div>
+
```
fnocetti, line 170: The fix for this duplicated code is not easy at all. I tried several alternatives and all of them (abstracting to a component, abstracting to a render function) introduce at least one race condition with the states (because of how the react-redux-form package manages it's own state). So I agreed with @daneryl that:
- Because this is markup and not algorithms, it's not that bad, to have it duplicated, in opposition to algorithms. It would even obfuscate the code.
- The whole view needs a big refactor to clean it up.

That said, we are assuming this CC issue for the time being and I'll open a Tech Debt issue to tackle the CollectionSettings refactor at some point.

app/react/App/Root.js
mfacar: I agree although the Root component seems small there is some logic that should be tested in its own context.

mfacar (approved):
LGTM!Nice Job!

Issue #3142, opened 2020-09-01
Allow "upgrade" of certain template properties: select -> multiselect, text -> rich text (markdown)


RafaPolit:
Given our new metadata structure, selects can, transparently and without any issue, become multiselects (not the other way around!!!).  Text fields can also become Rich Text fields (and, in this case, loosing the formatting, perhaps also the other way around).

We should provide a way to "upgrade" this fields at any given point.  It's a good quality of life improvement and costs nothing on our side.

txau:
@RafaPolit what about rather simplifying and having only multi-select and rich text?

RafaPolit:
Well, I think it's not enough.  You really want some selects to allow only a single value.  So it could be a 'select' that has option to allow single or multiple values.

With text, it's a little less clear.  We could make all text rich text, but then all values would be wrapped in, at least, a <p>, and that is not necessarily something we want for really simple text fields, so not sure.

RafaPolit:
This can be worked as part of the template creation redesign.

PR #3290, opened 2020-10-28
made markdowns use same mapping as text fields

### LINKED ISSUE #3276: Markdown and Text fields mapping problem
Markdown and Text fields are considered as the same type by the template creator, so a Markdown field and a text field both with the same name can coexists, but the mappings are different, Text has mapping for sorting, but makrdown doesn't.

konzz:
fixes #3276

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/database/elastic_mapping/mappings.ts
+++ b/database/elastic_mapping/mappings.ts
@@ -112,12 +112,6 @@ const linkType = () => {
   };
 };
 
-const markdownType = () => {
-  return {
-    value: noSorttext,
-  };
-};
-
 const selectType = () => {
   return {
     label: text,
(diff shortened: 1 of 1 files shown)


mfacar:
LGTM, very clean solution. 
Only for the record, while I was testing, I found that the property named same, one of text type and another of markdown type, in the filtering they are shown in two different inputs, however, when I input a value in either of them, this value is replicated in the other one. I think it should be addressed in #3142. What do you think @konzz ?

mfacar (approved):

konzz:
Good to know, I think this fix is too critical to hold it to include that, so I would merge it.

PR #3293, opened 2020-10-29
CORS middleware in the public submit endpoint

### LINKED ISSUE #3292: Allow cors for public submit endpoint
When submiting entities to the public submit endpoint from a different domain the request is bloqued due to CORS policy. Allow cors for that specific endpoint.

konzz:
fixes #3292 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


daneryl (changes requested):
Usage of sessions for this feature has some disadvantages, maybe it was good enough for the first iteration, i feel like now it merits that we rethink it.
if instead of using a cookie/session the endpoint returns the SVG and a matching ID we can easily check this on the form endpoint. (the form should send the captcha solution and the id)

- No cors and no new flag to uwazi.
- Endpoint its better encapsulated, its easier to understand.
- We will prevent current edge cases that i think exists, like same user opening multiple forms, im assuming only one of those captchas will be valid? (i have not tested this)
- We also will improve the /remotecaptcha endpoint to only pass down the response instead of the target cookie.

I understand its tradeoffs, and maybe we do not want to extend this development more. @txau @RafaPolit @konzz

app/api/auth/captchaMiddleware.js
```diff
@@ -1,11 +1,17 @@
...
-    const captchaInTheBody = req.body && req.body.captcha;
-    if (req.session && captchaInTheBody && req.session.captcha === req.body.captcha) {
+import { CaptchaModel } from './CaptchaModel';
+
+export default () => async (req, res, next) => {
+  if (req.body && req.body.captcha) {
+    const submitedCaptcha = JSON.parse(req.body.captcha);
+    const captcha = (await CaptchaModel.get({ captcha: submitedCaptcha.captcha }))[0];
+
+    if (captcha && captcha._id.toString() === submitedCaptcha.id) {
```
daneryl, line 8: @konzz this query can be exploited with NOSQL injection to by pass the validation sending this on the body as the captcha field:
```
{ "captcha": { "$regex": ".*"},  "id":"id you get on the captcha endpoint"} 
```

this can be easily prevented querying by id to the db and comparing by captcha text i think ?
i think this can only be exploited when there only one captcha in the db or the first one matches, but still the fix is easy enough.

daneryl (changes requested):
one security bug and some minor comments, apart from that works well

daneryl (changes requested):
@konzz captcha field now sends wrong captcha name, ```captcha``` instead of ```text```
also the validation always passes now, i do not know why but it may be related.

daneryl:
@konz still not working, blank page when i try to type on the captcha.

PR #3303, opened 2020-11-02
Added support for ordering files by language and updated tests

### LINKED ISSUE #3229: Order of documents
We are not explicitly managing the order of the main documents associated to an entity. Ideally, the UI language one comes first, ie this is a fora user navigating in Spanish:

image

grafitto:
fixes #3229 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [x] Test for side effects
- [x] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Attachments/components/FileList.tsx
+++ b/app/react/Attachments/components/FileList.tsx
@@ -7,6 +7,7 @@ import { EntitySchema } from 'shared/types/entityType';
 import UploadButton from 'app/Metadata/components/UploadButton';
 import { ConnectedFile as File } from './File';
 import { NeedAuthorization } from 'app/Auth';
+import languageLib from 'shared/languages';
 
 const defaultProps = {
   files: [],
@@ -41,21 +42,9 @@ export class FileList extends Component<FileListProps> {
   render() {
     const { files, entity } = this.props;
     const fileIndex = files.findIndex(file => {
-      let lang = '';
-      switch (file.language) {
-        case 'eng':
-          lang = 'en';
-          break;
-        case 'spa':
-          lang = 'es';
-          break;
-        case 'por':
-          lang = 'pt';
-          break;
-        default:
-          lang = 'en';
-      }
-      return lang === entity.language;
+      const language = languageLib.get(file.language as string, 'ISO639_1');
+      console.log('Language returned: ', language);
+      return language === entity.language;
     });
     if (fileIndex > -1) {
       const temp = files[fileIndex];
--- a/app/react/Attachments/components/FileList.tsx
+++ b/app/react/Attachments/components/FileList.tsx
@@ -40,7 +40,7 @@ export class FileList extends Component<FileListProps> {
   }
 
   orderFilesByLanguage(files: FileType[], systemLanguage: string) {
-    const orderedFiles = files;
+    const orderedFiles = [...files];
     const fileIndex = orderedFiles.findIndex(file => {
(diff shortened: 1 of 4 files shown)


app/react/Attachments/components/FileList.tsx
```diff
@@ -39,8 +40,16 @@ export class FileList extends Component<FileListProps> {
   }
 
   render() {
-    const { files } = this.props;
-
+    const { files, entity } = this.props;
+    const fileIndex = files.findIndex(file => {
```
konzz, line 44: Please abstract this code to a different method

grafitto: On it

grafitto: I am done refactoring

app/react/Attachments/components/FileList.tsx
```diff
@@ -38,15 +39,29 @@ export class FileList extends Component<FileListProps> {
...
+    const fileIndex = orderedFiles.findIndex(file => {
+      const language = languageLib.get(file.language as string, 'ISO639_1');
+      return language === systemLanguage;
+    });
+    if (fileIndex > -1) {
+      const temp = orderedFiles[fileIndex];
+      orderedFiles[fileIndex] = orderedFiles[0];
+      orderedFiles[0] = temp;
+    }
+    return orderedFiles;
```
daneryl, line 53: @grafitto if i comment this code, the tests are still green, the test sets up an scenario where the first file is already the system language. this is something that we want to probably test, but we need to test also when this is not the case.

daneryl (changes requested):
@grafitto i have tested this using the application and it works well 👍, however i think we are missing some important specs, i have commented on the specific code.

app/react/Attachments/components/FileList.tsx
grafitto: Done

grafitto: @daneryl

app/react/Attachments/components/specs/FileList.spec.tsx
```diff
@@ -37,20 +38,37 @@ describe('FileList', () => {
...
+    expect(entity.language).toEqual(language);
+  });
+
+  it('should render the files starting with the one with the system language', () => {
+    const firstFileLanguage = languageLib.get(props.files[0].language as string, 'ISO639_1');
+    if (firstFileLanguage === entity.language) {
+      const temp = props.files[0];
+      props.files[0] = props.files[props.files.length - 1];
+      props.files[props.files.length - 1] = temp;
+    }
```
daneryl, line 62: @grafitto i do not think this is needed, you can just redefine props, its shorter and with better readability.
```
 props.files = [file2, file];
```
 i think it is also a good idea to try to avoid conditionals on tests as much as possible.

daneryl (changes requested):
added new comment on the test, there is also a Eslint error, every check in the merge PR section should be green, except the "Modules security check / yarn-audit" 👍

for the Eslint error its useful to have your IDE or text editor configured to warn you when you are on the files, let me know if you need help.

app/react/Attachments/components/specs/FileList.spec.tsx
grafitto: @daneryl done.

daneryl (approved):

PR #3315, opened 2020-11-12
Fixes a case where user deletion does not appear on activity logs

### LINKED ISSUE #3310: Apparently user deletions are not being logged in the activity log
Research if user deletions is not getting logged in the activity log.

grafitto:
fixes #3310 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [x] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/activitylog/activitylogMiddleware.js
```diff
@@ -27,8 +27,8 @@ export const BODY_REQUIRED_ENDPOINTS = [
 function mustBeLogged(baseurl, method, body) {
   const isLoggedRequest =
     baseurl.includes('/api/') &&
-    !ignoredMethods.includes(method) &&
-    !IGNORED_ENDPOINTS.includes(baseurl);
+    ((!ignoredMethods.includes(method) && !IGNORED_ENDPOINTS.includes(baseurl)) ||
```
konzz, line 30: I think this solution is too specific for the problem, if we found more scenarios like this we dont want to have a long list of "ifs" here.

Probaly making the ignored endpoints a combination of method + endpoint is better.

app/api/activitylog/activitylogMiddleware.js
```diff
@@ -4,7 +4,7 @@ import activitylog from './activitylog';
 
 const ignoredMethods = ['GET', 'OPTIONS', 'HEAD'];
 export const IGNORED_ENDPOINTS = [
-  '/api/users',
+  // '/api/users',
```
grafitto, line 7: I removed this endpoint completely from IGNORED_ENDPOINTS. The reason is, only 3 methods exists on this endpoint, POST, GET and DELETE, POST and DELETE need to be logged and GET should not be logged. Which is taken care of by `ignoredMethods` array... @konzz @mfacar I am not sure why the endpoint was included, kindly shed more light into this. Also thanks @daneryl for the help

app/api/activitylog/activitylogMiddleware.js
```diff
@@ -39,6 +39,7 @@ export default (req, _res, next) => {
   if (mustBeLogged(baseurl, method, body)) {
     const time = Date.now();
     const expireAt = date.addYearsToCurrentDate(1);
+    if (body.password) body.password = '*****';
```
grafitto, line 42: I also removed this coz @daneryl noticed the password is displayed in plain text on the logs

app/api/activitylog/activitylogMiddleware.js
mfacar: @grafitto I think it's fine to remove 'api/users' from the ignored endpoints, I'm not sure when it was added but as you found it is used to update users.

daneryl (changes requested):
the log that appears when POST api/users its marked as RAW, what is the next thing so that this appears properly on the logs ? @RafaPolit @mfacar

app/api/activitylog/activitylogMiddleware.js
daneryl: good, we can remove the commented line then :)

mfacar:
@daneryl, @grafitto, To see the API/users endpoint formatted it is needed to add the entry value in activitylogParser.ts

app/api/activitylog/activitylogMiddleware.js
grafitto: Done

grafitto:
> @daneryl, @grafitto, To see the API/users endpoint formatted it is needed to add the entry value in activitylogParser.ts

Done, I hope that is how it should be done 😄 @mfacar

daneryl (approved):

PR #3322, opened 2020-11-17
3296 augment error log

### LINKED ISSUE #3296: Rewrite createError to show original error and prettyfied error
No matter how sophisticated our prettify error is, there is data that may get lost.  We need to rewrite that so that logs reflect the pretty error, that the client shows the prettified error, but that logs and notifications also include the raw error information for debugging purposes.

fnocetti:
fixes #3296 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/utils/handleError.js
+++ b/app/api/utils/handleError.js
@@ -28,7 +28,7 @@ const joiPrettifier = (error, req) => {
 };
 
 const appendOriginalError = (message, originalError) =>
-  `${message}\nOriginal error: ${JSON.stringify(originalError, null, ' ')}`;
+  `${message}\noriginal error: ${JSON.stringify(originalError, null, ' ')}`;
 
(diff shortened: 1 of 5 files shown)


app/api/utils/handleError.js
```diff
@@ -79,6 +82,31 @@ const prettifyError = (error, { req = {}, uncaught = false } = {}) => {
...
+  const original = data.original || error;
+
+  if (original instanceof Error) {
+    return original;
+  }
+
+  return null;
+};
+
+const sendLog = (data, error, errorOptions) => {
```
fnocetti, line 95: I don't think we should fix this but, if you can think of a way to split this function, please let me know.

PR #3327, opened 2020-11-17
3311 schema fix

### LINKED ISSUE #3311: Check the objectIdSchema type, it is casgting 'unknown | undefined'
The objectIdSchema imported everywhere where _id is defined in types / validations, is casting a:
```
string | { [k: string]: unknown | undefined; }
```
type, which is probably not correct.  We should look into this.

RafaPolit:
fixes #3311

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/attachments/attachments.ts
+++ b/app/api/attachments/attachments.ts
@@ -34,12 +34,16 @@ export default {
       _id: { $ne: entity._id },
     });
 
-    const attachmentToDelete = (entity.attachments || []).find(a => a._id.equals(attachmentId));
+    const attachmentToDelete = (entity.attachments || []).find(
+      a => a._id && a._id.toString() === attachmentId.toString()
+    );
 
     const [savedEntity] = await entities.saveMultiple([
       {
         ...entity,
-        attachments: (entity.attachments || []).filter(a => !a._id.equals(attachmentId)),
+        attachments: (entity.attachments || []).filter(
+          a => !(a._id && a._id.toString() === attachmentId.toString())
+        ),
       },
     ]);
 
--- a/app/api/attachments/attachments.ts
+++ b/app/api/attachments/attachments.ts
@@ -35,14 +35,14 @@ export default {
     });
 
     const attachmentToDelete = (entity.attachments || []).find(
-      a => a._id && a._id.toString() === attachmentId.toString()
+      a => a._id?.toString() === attachmentId.toString()
     );
 
     const [savedEntity] = await entities.saveMultiple([
       {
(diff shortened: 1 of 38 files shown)


app/api/attachments/attachments.ts
```diff
@@ -34,12 +34,16 @@ export default {
       _id: { $ne: entity._id },
     });
 
-    const attachmentToDelete = (entity.attachments || []).find(a => a._id.equals(attachmentId));
+    const attachmentToDelete = (entity.attachments || []).find(
+      a => a._id && a._id.toString() === attachmentId.toString()
```
RafaPolit, line 38: @daneryl I'd like to understand why `a._id?.toString()` is not a better approach. As I understand it, it is, precisely, a shortcand to avoid these extra controls.  Also, it will return `undefined` if any of the checks fail, which is the same (mostly) as doing this booleans which would return false.  

So, using this in filters and other boolean comparators seems like the way to go.

Thanks for any feedback.

app/api/templates/specs/utils.spec.ts
```diff
@@ -13,9 +13,9 @@ describe('templates utils', () => {
       it('should sanitize the labels and append the type', async () => {
         await settings.save({});
         const result = await generateNamesAndIds([
-          { label: ' my prop ', type: 'text' },
-          { label: 'my^foreïgn$próp"', type: 'text' },
-          { label: ' my prop ', type: 'geolocation' },
+          { label: ' my prop ', name: '', type: 'text' },
```
RafaPolit, line 16: Refactor into conforming property function

app/react/Review/components/OneUpEntityViewer.tsx
```diff
@@ -185,7 +184,7 @@ export class OneUpEntityViewerBase extends Component<
                           showType={false}
                           showSubset={this.nonMlProps()}
                         />
-                        <FileList files={entity.documentsdocuments} entity={entity} />
+                        <FileList files={entity.documents} entity={entity} />
```
RafaPolit, line 187: Here, I had to just assume that `documentsdocuments` is a typo.  No idea if that ever worked.  We have NOTHING in the entire system that assigns a `documentsdocuemnts` property. Still, I cannot test this properly, so it is just an assumption.

daneryl: this is 100% a typo

PR #3330, opened 2020-11-18
Avoid JSON.parse-ing the search term on the query

### LINKED ISSUE #3265: Issue with exact term ( "____")  search function
According to our admin guide when one searches for a phrase/term between "___" it should show results that match the exact term within the quotation marks.
Screenshot 2020-10-19 at 6 02 04 PM

However as seen in the screenshot, the search function is picking up individual words as well and showing the results for those too. For example, when searching "Ministry of State Security" I can see results for entities that contain the individual words "of", "State", "Security" and so on instead of seeing the entities that have the entire exact phrase. 

Screenshot 2020-10-19 at 6 02 27 PM

fnocetti:
fixes #3265 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/utils/parseQueryMiddleware.ts
+++ b/app/api/utils/parseQueryMiddleware.ts
@@ -3,7 +3,14 @@ import { Request, Response, NextFunction } from 'express';
 interface query {
   [key: string]: string;
 }
+
+const pureStringProperties = ['searchTerm'];
+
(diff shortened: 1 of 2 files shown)


mfacar (approved):
LGTM, nice and simple solution

PR #3332, opened 2020-11-20
Change numeric field input type to numeric, add support for value 0

### LINKED ISSUE #3331: Numeric field does not allow to type numbers properly.
if you try to type something with decimal numbers, the field will not allow it, the only way to write it is to write the number without decial places, then go back and put the "." or ","

fnocetti:
fixes #3331 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/csv/specs/typeFormatters.spec.ts
+++ b/app/api/csv/specs/typeFormatters.spec.ts
@@ -161,6 +161,7 @@ describe('csvExporter typeFormatters', () => {
   describe('SIMPLE', () => {
     it('should return the correct NUMERIC value', () => {
       testSimple(1234, typeFormatters.numeric, 1234);
+      testSimple(0, typeFormatters.numeric, 0);
     });
 
     it('should return the correct TEXT value', () => {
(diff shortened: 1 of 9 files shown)


mfacar:
@fnocetti, regarding the support for value 0,  when I tried to save a required numeric property with 0, I'm getting the property is required as validation error, I think that 0 should be a valid value for a required number. Don't you think?

fnocetti:
> @fnocetti, regarding the support for value 0, when I tried to save a required numeric property with 0, I'm getting the property is required as validation error, I think that 0 should be a valid value for a required number. Don't you think?
> Also, when we export to CSV, the numeric properties with 0 as value are empty cells. Given filtering doesn't support filtering by 0 (or negative numbers that are supported as values too) maybe could it be better to create a new issue with this detail?

Wow, awesome QA. Thank you!
For the required fields and CSV export: I'll take a look.
For the filters, I think we should create a separate issue. I'll take a look anyway and let you know if it's an easy solution.

mfacar:
@fnocetti great! all the changes are working perfectly, now zero and negative numbers are supported by requested fields and filters nicely. There is only a little thing, at the cards, a numeric property with 0, is being formatting with the label **No value**. Could you address it too?

mfacar (commented):
LGTM!

PR #3333, opened 2020-11-21
Fix empty limit set to empty string on numeric range

### LINKED ISSUE #2978: Numeric range filters fail when only 'from' digit is input
If I create a numeric field and then add it as filter, apparently the ES search fails upon searching only the 'from' (while I am still writing, before entering the "to").  This only happens after the first time.  That is, you enter a From, it filters, you enter a To, it filters.  You then DELETE the To, and it fails... and then if fails whatever you do.

### LINKED ISSUE #3004: Numeric filter sends an empty value when cleared causing a query shard exception
1. Input a value in the field "From" and wait for automatic submission.
2. Input a value in the field "To" and wait.
3. Delete the value from the field "From".

image

fnocetti:
fixes #3004 
fixes #2978

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Forms/components/NumericRange.js
+++ b/app/react/Forms/components/NumericRange.js
@@ -11,9 +11,19 @@ export default class NumericRange extends Component {
   }
 
   onChange(prop, value) {
-    const state = { ...this.state, [prop]: value };
-    this.setState(state);
-    this.props.onChange(state);
+    const newState = {
+      ...this.state,
+      [prop]: value,
+    };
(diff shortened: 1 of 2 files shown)


mfacar (approved):
LGTM!, a clean and nice solution :ok_hand:

PR #3346, opened 2020-11-27
Added support for date modified field on cards

### LINKED ISSUE #1893: Request/Idea: Add an automated "Date updated" property option
I'm putting together a collection of tools and resources related to human rights documentation (https://doc-resources.uwazi.io). This is meant to be a living collection. It would be really helpful if visitors could see the "Date updated" instead of "Date added" so they know when the information on that card has last been updated.

grafitto:
fixes #1893 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Library/components/specs/SortButtons.spec.js
```diff
@@ -110,7 +110,7 @@ describe('SortButtons', () => {
             .find('span')
             .last()
             .text()
-        ).toBe('Search relevance');
+        ).toBe('Date modified (Recently)');
```
daneryl, line 113: @grafitto the description and string being tested do not match.
i think the default still should be "Search Relevance" when there is an active search term.
what you are testing here its already being tested by the full component snapshot in a previous test, this is something i think we should avoid. refactor the test to not use snapshots would be a good option.

app/react/Metadata/helpers/formater.js
```diff
@@ -45,9 +45,22 @@ const addCreationDate = (result, doc) =>
...
 
+const addModificationDate = (result, doc) =>
+  result.push({
+    value: moment.utc(doc.creationDate).format('ll'),
+    label: 'Date modified',
+    name: 'editDate',
+    translateContext: 'System',
+    sortedBy: true,
+  });
+
```
daneryl, line 56: @grafitto this is not tested, actually the value being formated is wrong, i think "creationDate" should be "editDate" ? it is really usefull to test first to make you think about the implementation and avoid this kind of errors.

app/react/Metadata/helpers/formater.js
```diff
@@ -57,6 +70,10 @@ const conformSortedProperty = (metadata, templates, doc, sortedProperty) => {
     result = addCreationDate(result, doc);
   }
 
+  if (sortedProperty === 'editDate') {
+    result = addModificationDate(result, doc);
+  }
```
daneryl, line 75: @grafitto not tested.

app/react/Metadata/helpers/formater.js
```diff
@@ -45,9 +45,22 @@ const addCreationDate = (result, doc) =>
...
+  });
+
 const conformSortedProperty = (metadata, templates, doc, sortedProperty) => {
   const sortPropertyInMetadata = metadata.find(p => sortedProperty === `metadata.${p.name}`);
-  if (!sortPropertyInMetadata && sortedProperty !== 'creationDate') {
+  if (
+    !sortPropertyInMetadata &&
+    sortedProperty !== 'creationDate' &&
+    sortedProperty !== 'editDate'
+  ) {
```
daneryl, line 63: @grafitto  not tested.

app/react/Markdown/components/specs/__snapshots__/MarkdownMedia.spec.js.snap
```diff
@@ -18,6 +18,8 @@ exports[`MarkdownMedia render should render an iframe with the correct video id
           },
```
daneryl, line 18: @grafitto this snapshot test is actually failing locally, also on ci, make sure you upgrade node packages and launch tests again.

app/server.js
```diff
@@ -105,7 +105,7 @@ DB.connect(config.DBHOST, dbAuth).then(async () => {
     });
   }
 
-  const bindAddress = { true: 'localhost' }[process.env.LOCALHOST_ONLY];
+  const bindAddress = { true: '*.*.*.*' }[process.env.LOCALHOST_ONLY];
```
daneryl, line 108: @grafitto why this change ? the config after this do not match the condition.
now the implementation is -> if there is a flag ```LOCALHOST_ONLY``` then use any ip ?
what do you need to achieve here ?

daneryl (changes requested):
some tests are failing, also some tests are missing, the details are in the code comments.

app/server.js
grafitto: This was not supposed to be commited, I commited it by mistake. I had changed to be able to access uwazi via my mobile phone on the same network. I have reverted.

app/api/entities/entities.js
```diff
@@ -305,6 +305,7 @@ export default {
       templates.get({ default: true }),
     ]);
     let docTemplate = template;
+    doc.editDate = date.currentUTC();
```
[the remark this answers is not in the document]
grafitto: Added tests

app/react/Metadata/helpers/formater.js
grafitto: Added tests and made the update, nice catch.

app/react/Metadata/helpers/formater.js
grafitto: I believe I also have added a test to this.

app/react/Metadata/helpers/formater.js
grafitto: Done

app/react/Library/components/specs/SortButtons.spec.js
RafaPolit: The "Date Modified" indeed is being tested by the snapshot.  The "Search relevance", on the other hand, is not, and is only tested here.  The issue is that this is testing the existence of "Search relevance" by position, and it is now on a different location.

The test has been reverted to look for Search relevance but on the correct position.  Still, it would be interesting to make this test resilient to position changes of the element, since there is nothing important regarding the position of the element.  It just needs to be there.

This is a good exercise in writing the "right" tests: the position was not important, the text and its link are.  The text is being tested, the link isn't and the position is being tested when it is not important.

grafitto:
Ready for review

daneryl (changes requested):
does not work properly on already existing collections, to replicate.

- yarn blank state
- modify an entity
- sorting by date modified (recently or least recently) results on the modified entity showing on the last position.

maybe this needs a migration to add the date modified property ? @grafitto @RafaPolit  ?

grafitto:
> does not work properly on already existing collections, to replicate.
> 
> * yarn blank state
> * modify an entity
> * sorting by date modified (recently or least recently) results on the modified entity showing on the last position.
> 
> maybe this needs a migration to add the date modified property ? @grafitto @RafaPolit ?

I think it needs migration...

daneryl:
> - yarn blank state
> - modify an entity
> - sorting by date modified (recently or least recently) results on the modified entity showing on the last position.

this is wrong, the proper way to replicate is to use existing data, (yarn fixtures, instead of blank-state)

grafitto:
> > * yarn blank state
> > * modify an entity
> > * sorting by date modified (recently or least recently) results on the modified entity showing on the last position.
> 
> this is wrong, the proper way to replicate is to use existing data, (yarn fixtures, instead of blank-state)

Noted. After testing, I think the `editDate` property is not added when an entity is edited

grafitto:
Only the entities with the `editDate` property are being sorted, the rest are not sorted and remain the same. If you edit more than one entities, you can actually see them getting sorted but the rest stay in the same position. @daneryl

daneryl (changes requested):
@grafitto migrations looks good.
i have noticed an issue where sorting by the new "date modified" results in the same order whether it is sorted by "recently" or "least recently"

grafitto:
> @grafitto migrations looks good.
> i have noticed an issue where sorting by the new "date modified" results in the same order whether it is sorted by "recently" or "least recently"

Does this happen even after you run the migration?

daneryl:
> > @grafitto migrations looks good.
> > i have noticed an issue where sorting by the new "date modified" results in the same order whether it is sorted by "recently" or "least recently"
> 
> Does this happen even after you run the migration?

yes, does it work properly for you ?

grafitto:
> > > @grafitto migrations looks good.
> > > i have noticed an issue where sorting by the new "date modified" results in the same order whether it is sorted by "recently" or "least recently"
> > 
> > 
> > Does this happen even after you run the migration?
> 
> yes, does it work properly for you ?

Mine seems to work as expected. I run `yarn migrate` and `yarn reindex`, I am not sure if reindexing helps. You could try and see

daneryl:
> > > > @grafitto migrations looks good.
> > > > i have noticed an issue where sorting by the new "date modified" results in the same order whether it is sorted by "recently" or "least recently"
> > > 
> > > 
> > > Does this happen even after you run the migration?
> > 
> > 
> > yes, does it work properly for you ?
> 
> Mine seems to work as expected. I run `yarn migrate` and `yarn reindex`, I am not sure if reindexing helps. You could try that

@grafitto this was an issue on my side, sorting is working properly now, im not sure what happened may be i forgot a reindex.

sorting apparently works properly but the date shown on the card for "Date modified" its always today not the actual date modified, it can be that the new field is somehow excluded from the results, please review.

grafitto:
> > > > > @grafitto migrations looks good.
> > > > > i have noticed an issue where sorting by the new "date modified" results in the same order whether it is sorted by "recently" or "least recently"
> > > > 
> > > > 
> > > > Does this happen even after you run the migration?
> > > 
> > > 
> > > yes, does it work properly for you ?
> > 
> > 
> > Mine seems to work as expected. I run `yarn migrate` and `yarn reindex`, I am not sure if reindexing helps. You could try that
> 
> @grafitto this was an issue on my side, sorting is working properly now, im not sure what happened may be i forgot a reindex.
> 
> sorting apparently works properly but the date shown on the card for "Date modified" its always today not the actual date modified, it can be that the new field is somehow excluded from the results, please review.

On it

grafitto:
@daneryl I have added `editDate` to documentQueryBuilder, kindly check whether that fixes the issue

daneryl (approved):

PR #3360, opened 2020-12-10
Fixed new user password not being usable

### LINKED ISSUE #3209: Setting password upon user creation doesn't get saved
If you provide an specific password when creating user, the password won't be saved. You can the edit the user and save it afterwards and it will be saved.

grafitto:
fixes #3209 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/activitylog/activitylogMiddleware.js
+++ b/app/api/activitylog/activitylogMiddleware.js
@@ -36,7 +36,6 @@ export default (req, _res, next) => {
   const { url, method, params, query, body, user = {} } = req;
   const baseurl = url.split('?').shift();
   if (mustBeLogged(baseurl, method, body)) {
-    const time = Date.now();
     const expireAt = date.addYearsToCurrentDate(1);
     const bodyLog = { ...body };
     if (bodyLog.password) bodyLog.password = '*****';
@@ -48,7 +47,7 @@ export default (req, _res, next) => {
       body: JSON.stringify(bodyLog),
       user: user._id,
       username: user.username,
-      time,
+      time: Date.now(),
       expireAt,
     };
     activitylog.save(entry);
--- a/app/api/activitylog/activitylogMiddleware.js
+++ b/app/api/activitylog/activitylogMiddleware.js
@@ -38,13 +38,14 @@ export default (req, _res, next) => {
   if (mustBeLogged(baseurl, method, body)) {
     const time = Date.now();
     const expireAt = date.addYearsToCurrentDate(1);
-    if (body.password) body.password = '*****';
+    const bodyLog = { ...body };
+    if (bodyLog.password) bodyLog.password = '*****';
     const entry = {
       url: baseurl,
       method,
       params: JSON.stringify(params),
       query: JSON.stringify(query),
-      body: JSON.stringify(body),
+      body: JSON.stringify(bodyLog),
       user: user._id,
       username: user.username,
       time,
--- a/app/api/users/specs/users.spec.js
+++ b/app/api/users/specs/users.spec.js
@@ -99,7 +99,8 @@ describe('Users', () => {
 
       it('should do the recover password process (as a new user)', done => {
         users
-          .newUser({
+          .newUser(
+            {
               username: 'spidey',
               email: 'peter@parker.com',
               password: 'mypass',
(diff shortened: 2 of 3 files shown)


grafitto:
@daneryl @fnocetti I think this is ready for QA, the failing tests are related to the Markdown thing @konzz was talking about, and codeclimate errors could be ignored

app/api/activitylog/activitylogMiddleware.js
```diff
@@ -38,13 +38,14 @@ export default (req, _res, next) => {
   if (mustBeLogged(baseurl, method, body)) {
     const time = Date.now();
     const expireAt = date.addYearsToCurrentDate(1);
-    if (body.password) body.password = '*****';
+    const bodyLog = { ...body };
```
grafitto, line 41: @RafaPolit if body is undefined then `{ ...body }` resolves to an empty object `{}`, would it be okay to leave it that way instead of introducing a branch? @konzz

RafaPolit: Yes, I think this is good enough.

grafitto: This is ready for QA @RafaPolit

app/api/activitylog/activitylogMiddleware.js
```diff
@@ -38,13 +38,14 @@ export default (req, _res, next) => {
...
+    if (bodyLog.password) bodyLog.password = '*****';
```
mfacar, line 42: @grafitto there is a max-statements issue on codeclimate for this method, since all new lines are required, maybe we could reduce the old code by instance removing the constant for time and using it inline at the entry `time: Date.now()`, what do you think?

app/api/users/users.js
```diff
@@ -178,7 +177,7 @@ export default {
     }
     const _user = await model.save({
       ...user,
-      password: await encryptPassword(random()),
+      password: await encryptPassword(user.password),
```
mfacar, line 180: Good catch @grafitto, please could you add a test to cover this, if I reverted the changes all tests are passing.

app/api/users/specs/users.spec.js
```diff
@@ -148,6 +156,7 @@ describe('Users', () => {
           {
             username: 'without2fa',
             email: 'another@email.com',
+            password: 'mypass',
```
mfacar, line 159: Despite this addition is nice to check the whole object, it is not covering the main reason for the problem with the password, because it was originated at the middleware side. I think we should add a kind of integration test at the route in order to verify if users.save is called with the correct params, but since instrumentRoutes, the helper we are using for tests the api calls, is not executing the middlewares as part of the call, it's fine to omit the test for now.

app/api/activitylog/activitylogMiddleware.js
grafitto: Nice idea, will do that

mfacar:
@grafitto LGTM, I left a couple of comments, please review them to be able to merge.

grafitto:
> @grafitto LGTM, I left a couple of comments, please review them to be able to merge.

Will do that

app/api/activitylog/activitylogMiddleware.js
grafitto: I believe the tests are already testing the password as opposed to `random()`

app/api/users/users.js
grafitto: I believe the tests are already testing the password as opposed to `random()`

mfacar: Actually no, I saw some tests that check the password, but since all of them are passing even without the new changes, there is no one that covers this case.

app/api/activitylog/activitylogMiddleware.js
grafitto: Done

mfacar:
The team agree to close the PR until we could test this appropriately

yuvadm:
I'd be happy to test this on `development` since #3367 is blocking me from deploying a new server - anything I need to know before?

RafaPolit:
Please, merge this also as a hot fix to master with all the required flow.

PR #3376, opened 2020-12-16
Fix infinite redirect when home_page has been set to /

### LINKED ISSUE #3268: Infinite redirection on home page URL
Setting the value of the home page to "/" causes an infinite redirection.

grafitto:
fixes #3268 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [x] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Settings/SettingsAPI.ts
+++ b/app/react/Settings/SettingsAPI.ts
@@ -20,6 +20,10 @@ export default {
       }),
     });
 
+    if (processedSettings.data?.home_page === '/') {
+      processedSettings.data.home_page = '';
+    }
+
     return api.post('settings', processedSettings).then((response: any) => response.json);
   },
 
--- a/app/react/Settings/SettingsAPI.ts
+++ b/app/react/Settings/SettingsAPI.ts
@@ -19,11 +19,6 @@ export default {
         return linkWithoutLocalID;
       }),
     });
-
-    if (processedSettings.data?.home_page === '/') {
-      processedSettings.data.home_page = '';
-    }
-
     return api.post('settings', processedSettings).then((response: any) => response.json);
   },
 
--- a/app/api/settings/settings.ts
+++ b/app/api/settings/settings.ts
@@ -127,6 +127,10 @@ export default {
     await saveLinksTranslations(settings.links, currentSettings.links);
     await saveFiltersTranslations(settings.filters, currentSettings.filters);
 
+    if (settings.home_page && settings.home_page === '/') {
+      settings.home_page = '';
+    }
+
     const result = await settingsModel.save({ ...settings, _id: currentSettings._id });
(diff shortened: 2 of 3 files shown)


app/react/Settings/SettingsAPI.ts
```diff
@@ -20,6 +20,10 @@ export default {
       }),
     });
 
+    if (processedSettings.data?.home_page === '/') {
+      processedSettings.data.home_page = '';
+    }
```
grafitto, line 25: Is this sufficient? @RafaPolit @txau

txau: I think this should go on the server side rather than a client side overwrite. There is a validateSettings routine.

grafitto: I have made the update

daneryl:
@grafitto @txau in my opinion the fix should be done on client side to make it more resilient and this way we avoid having to validate anything on server and potentially need a migration for databases that already have this value.
the problem is that the code that does this is not well tested... but still i think its a best long term option.

https://github.com/huridocs/uwazi/blob/4045dd5f78223ccaf70f5070887b8555ab89d0bf/app/react/Routes.js#L101-L123

this is the potential fix, will need more testing.

```
function getIndexRoute(_nextState, callBack) {
  const state = store.getState();
  const homePageSetting = state.settings.collection.get('home_page');
  const customHomePage = homePageSetting ? homePageSetting.split('/').filter(v => v) : [];
  const isPageRoute = customHomePage.includes('page');

  if (isPageRoute) {
    return callBack(null, getPageIndexRoute(customHomePage));
  }

  if (customHomePage.length) {
    return callBack(null, {
      onEnter: (_nxtState, replace) => {
        replace(customHomePage.join('/'));
      },
    });
  }

  return callBack(
    null,
    getDefaultLibraryComponent(state.settings.collection.get('defaultLibraryView'))
  );
}
```

grafitto:
I have updated to filter / home_pages when displaying. However I have noticed when one sets home_page on settings it is not persisted in the database. Not sure why or am I missing something? @daneryl

daneryl:
@grafitto home page settings are persisted for me, the only time when it is not persisted is when it is a '/'. this is probably because of the previous fix: https://github.com/huridocs/uwazi/pull/3376/commits/72c0420f24fa8a24ee81595a4d155cc38c53774d#diff-3cb3383483d8da067b31d58425a8d0de8b2c83e721c85ad9af5cd97705ee2588
this fix should not be necessary anymore, the "/" is a correct path, the bug was in the local parsing.

daneryl:
LGTM, we should test the index function, lets keep track here #3401

PR #3381, opened 2020-12-17
Fixed bug when comparing mapping of different prop types

### LINKED ISSUE #3377: The ''delete property'' feature is failing
Delete property feature is failing, and throws an error message.
- Select a template 
- Delete any property 
- The property first disappears 
- Click ''Save'' 
- You get an error message
- Refresh the browser, the deleted property reappears
- Seems to be specific to this instance only

Screenshot 2020-12-16 at 10 00 47

konzz:
fixes #3377 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/search/entitiesIndex.js
```diff
@@ -184,10 +184,15 @@ const equalPropMapping = (mapA, mapB) => {
...
+  const sameProps =
+    Object.keys(mapA.properties).length === Object.keys(mapB.properties).length &&
+    Object.keys(mapA.properties).reduce(
+      (result, propKey) =>
+        result &&
+        mapB.properties[propKey] &&
+        mapA.properties[propKey].type === mapB.properties[propKey].type,
+      true
+    );
   return sameAmountOfProps && sameProps;
```
mfacar, line 196: @konzz maybe you forgot to remove the constant sameAmountOfProps in the two places or use this value in the condition to avoid executing the condition and the methods inside twice.

app/api/search/specs/entitiesIndex.spec.js
```diff
@@ -131,9 +131,16 @@ describe('entitiesIndex', () => {
...
+        { name: 'dob', type: 'text', label: 'Date of birth' },
+        { name: 'country', type: 'select', label: 'Country' },
+      ];
+
       response = await checkMapping(templateB, elasticIndex);
-      expect(response).toEqual({ errors: [{ name: 'Date of birth' }], valid: false });
+      expect(response).toEqual({
+        errors: [{ name: 'Date of birth' }, { name: 'Country' }],
+        valid: false,
+      });
```
mfacar, line 143: I'm not sure what test case covers this test. I commented the added condition in equalPropMapping and all the tests still passed. Please could you add a test to cover the specific scenario?

mfacar (commented):
@konzz, I was not able to reproduce the case described in #3377, but based on the changed code I tested the scenario that produces an unexpected error and the PR fixes it successfully. However, after the message about the need to do a reindex, I'm getting another message  "Properties can't swap name", maybe am I doing the wrong test scenario?

The steps I'm doing are:
1. Add in a template a Select property
2. Save the template
3. Remove the Select property and add a Text property with the same name of the removed property.
4. Save the template
Before the PR I'm getting an Unexpected error message.
After the PR I'm getting a descriptive message asking for reindexing.

app/api/search/entitiesIndex.js
konzz: Yeah, good one

konzz:
>I'm getting another message "Properties can't swap name", maybe am I doing the wrong test scenario?
That happens when you delete a property and add a new one with the same name and then save, it is expected.

But if you Delete -> Save -> Add Prop -> Save, it should only show the popup for reindex.

app/api/search/specs/entitiesIndex.spec.js
konzz: So to the "sameProps" condition I added two conditions, the line  186 and the line 190, if you coment both it actually throws an error because of the property select that was added to the test.

konzz:
closing in favor of #3454

Issue #2393, opened 2019-06-20
Multi-tenant uwazi core


konzz:
The idea of this issue is to discuss and track all the info regarding the idea of multiple Uwazi instances running in one node.
Issues detected so far:

- [x] Sockets\
   - [x] multi node process
   - [x] multi tenant
- [x] Isolate requests per project
- [ ] Cron jobs (sync, evidence vault)
- [ ] system keys on startup ? deprecate in favor of migration if we need to add any system key.
- [x] separated databases? Possible issues with scalability
- [x] separated folders for files?
- [x] Super lengthy development 👌🤔😬 
- [x] separate db to add tenants, node process to watch changes.
- [x] Migrations should take into account multiple possible databases
_this is not being addressed, migrations will work like they do now, env variables for the db index etc and script launch._
- [ ] does semantic search and topic classification need to be aware of this ?
- [ ] tests using filesystem are a bit flaky, i am assuming due to changes on the branch to use the same testing folder (can not replicate, maybe not as serious as i thought)

RafaPolit:
- Migrations should take into account multiple possible databases

txau:
Some resources regarding multi-tenancy in node apps:
- With continuation local storage: https://blog.lftechnology.com/designing-a-secure-and-scalable-multi-tenant-application-on-node-js-15ae13dda778 and  https://github.com/othiym23/node-continuation-local-storage
- With virtual hosts and express: https://github.com/expressjs/express/blob/master/examples/vhost/index.js

txau:
Apparently this can also be solved via async hooks: https://nodejs.org/docs/latest-v10.x/api/async_hooks.html

daneryl:
@txau, the solution with continuation-local-storage looks really nice, we should do a quick spike, should be "easy" to integrate into our app.

with virtual hosts it looks like it does instantiate as many express apps as needed ? not so good for memory consumption ? not sure though

did not read the async hooks but it is experimental apparently, maybe not yet.

habbes:
Regarding sockets. One approach is to separate the server operations that respond to user-requests and long-running async tasks. We could use a background worker for the async tasks like processing pdfs and semantic search. The server and worker would be decoupled, and a message queue would be put in place to handle communication between the two.

Here's a simple workflow:
- the user makes a request, creating a document with a pdf
- an uwazi server picks up the request, stores the necessary info in the db, adds a task to the job queue in the message broker, and sends a response to the client
[...]

PR #3406, opened 2021-01-08
Multitenant context refactor

fnocetti:
This refactors separates the multitenant async context management into two different concerns:

- Multitenancy
- Async context management

This is needed in order to use the async context for other functionalities, e.g user permissions.

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/utils/AppContext.ts
+++ b/app/api/utils/AppContext.ts
@@ -7,6 +7,12 @@ interface ContextData {
 class AppContext {
   private storage = new AsyncLocalStorage<ContextData>();
 
+  private getContextObject() {
+    const data = this.storage.getStore();
+    if (!data) throw new Error('Accessing nonexistent async context');
+    return data;
+  }
+
   async run(cb: () => Promise<void>, data: ContextData = {}): Promise<void> {
     return new Promise((resolve, reject) => {
       this.storage.run(data, () => {
@@ -18,13 +24,11 @@ class AppContext {
   }
 
   get(key: string) {
-    const data = this.storage.getStore();
-    return data && data[key];
+    return this.getContextObject()[key];
   }
 
   set(key: string, value: unknown) {
-    const data = this.storage.getStore();
-    if (data) data[key] = value;
+    this.getContextObject()[key] = value;
   }
 }
 
--- a/app/api/utils/AppContext.ts
+++ b/app/api/utils/AppContext.ts
@@ -1,0 +1,33 @@
+import { AsyncLocalStorage } from 'async_hooks';
+
+interface ContextData {
+  [k: string]: unknown;
+}
+
+class AppContext {
+  private storage = new AsyncLocalStorage<ContextData>();
+
+  async run(cb: () => Promise<void>, data: ContextData = {}): Promise<void> {
+    return new Promise((resolve, reject) => {
+      this.storage.run(data, () => {
+        cb()
+          .then(resolve)
+          .catch(reject);
+      });
+    });
+  }
+
+  get(key: string) {
+    const data = this.storage.getStore();
+    return data && data[key];
+  }
+
+  set(key: string, value: unknown) {
+    const data = this.storage.getStore();
+    if (data) data[key] = value;
+  }
+}
+
+const appContext = new AppContext();
+
+export { appContext };
--- a/app/api/utils/specs/appContextMiddleware.spec.ts
+++ b/app/api/utils/specs/appContextMiddleware.spec.ts
@@ -1,0 +1,31 @@
+import request from 'supertest';
+import express, { Application, Request, Response, NextFunction } from 'express';
+import { appContextMiddleware } from '../appContextMiddleware';
+import { appContext } from '../AppContext';
+
+const testingRoutes = (app: Application) => {
+  app.get('/api/testGET', (_req, res, next) => {
+    res.json(appContext.get('someKey'));
+    next();
+  });
+};
+
+const helperMiddleware = (req: Request, _res: Response, next: NextFunction) => {
+  appContext.set('someKey', req.get('someHeader'));
+  next();
+};
+
+describe('multitenant middleware', () => {
+  it('should execute next middlewares inside an async context', async () => {
+    const app: Application = express();
+    app.use(appContextMiddleware);
+    app.use(helperMiddleware);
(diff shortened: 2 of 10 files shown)


app/api/utils/AppContext.ts
```diff
@@ -0,0 +1,33 @@
...
+        cb()
+          .then(resolve)
+          .catch(reject);
+      });
+    });
+  }
+
+  get(key: string) {
+    const data = this.storage.getStore();
+    return data && data[key];
```
daneryl, line 22: @fnocetti i think we should manage this condition ?
in the original implementation we were throwing an error:
https://github.com/huridocs/uwazi/blob/3ed63fcd70dedf59bbbcc11042ed4b20f71dc6e1/app/api/tenants/tenantContext.ts#L59-L61

the reason for this is  because there was a bug where on certain cases the async context was lost, this way we have detected several processes we had more easily, where this was the case([reference](https://github.com/huridocs/uwazi/issues/2393#issuecomment-627775969)). 
I think that implementation belongs here now ? (apparently this bug was fixed on node but unless we know all the details on scenarios where the context can be lost, i think its safe to have it)

it is also a non tested condition.

app/api/utils/AppContext.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  set(key: string, value: unknown) {
+    const data = this.storage.getStore();
+    if (data) data[key] = value;
+  }
```
daneryl, line 28: @fnocetti i think we should manage and test this one too.
i feel like there should never be an undefined data ? what scenario apart from context loss can prouce this ? i see this as a potential silent error.

app/api/utils/AppContext.ts
```diff
@@ -0,0 +1,33 @@
...
+      });
+    });
+  }
+
+  get(key: string) {
+    const data = this.storage.getStore();
+    return data && data[key];
+  }
+
+  set(key: string, value: unknown) {
```
daneryl, line 25: can we work with strings here instead of unknown ? do you think it will be sufficient for permissions, we should  only need some key information, the user id ?

daneryl (changes requested):
@fnocetti i think this a really useful abstraction, make sense, nice job 👍 
i have made some comments regarding some details in the new asyncContext abstraction.

app/api/utils/specs/appContextMiddleware.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+    next();
+  });
+};
+
+const helperMiddleware = (req: Request, _res: Response, next: NextFunction) => {
+  appContext.set('someKey', req.get('someHeader'));
+  next();
+};
+
+describe('multitenant middleware', () => {
```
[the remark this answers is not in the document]
fnocetti: 🤦‍♂️  Thank you!!

app/api/utils/AppContext.ts
fnocetti: So happy you brought this topic! So, the easiest scenario in which you could have your data undefined is when you are running get or set outside of a context. That's basically what (I understood) the async hooks do.
My opinion is that we should not allow the use of the context variables outside of a context scope and throw an exception, but I didn't want to limit the functionality before having this discussion.
That said, I will do both:

- Keep the checking in the tenant side (the context might exist but be missing tenant information because of a tenant bug)
- Throw an exception if illegally accessing the getter/setter outside of a context scope (covering also possible bugs on node's side)

I'll also improve the testing coverage.

Thanks!

app/api/utils/AppContext.ts
fnocetti: see previous. thanks!

app/api/utils/AppContext.ts
fnocetti: It's probably (still not sure) not enough for permissions. We could use strings and serialize/deserialize the values on the consumer side if needed.
What are your thoughts?

fnocetti:
> @fnocetti i think this a really useful abstraction, make sense, nice job 👍
> i have made some comments regarding some details in the new asyncContext abstraction.

Thank you @daneryl ! Really valuable review!

app/api/utils/AppContext.ts
daneryl: sounds good to me 👍

app/api/utils/AppContext.ts
daneryl: 👍

app/api/utils/AppContext.ts
daneryl: then maybe a possible solution is to have a hard typed ContextData and somehow take the specific types belonging to the keys into account on the setter/getter ? im not sure this is possible but having it typed would be preferable.

daneryl: seems, those solutions are not straight forward, maybe leave this as unknown and type guard the accessors of the context with their own type, tenants and permissions know the type and this remains generic with unknown.

fnocetti: Yeah, I probably wouldn't hard type the context object and I say that type guarding, or even duck typing is good enough for a key-value storage like this.
If you think its worth we could do something like this:

```
get<T>(key: string): T {
    return this.getContextObject()[key] as T;
  }

  set<T>(key: string, value: T) {
    this.getContextObject()[key] = value;
  }
```

so internally it is still `unknown` and generic, but you could have typings at consumer's level like:

```
appContext.set<string>('key', 'value'); // second param is of type string
const value = appContext.get<string>('key') // return type is string
```

fnocetti: Also it's worth mentioning that, unlike `any`, `unknown` is type-safe. That means that ts will force you cast the value to assign it to anything that is not `unknown`.

daneryl: > Also it's worth mentioning that, unlike any, unknown is type-safe. That means that ts will force you cast the value to assign it to anything that is not unknown.

did not know this, thanks really usefull, if ts is going to force us to cast the value then i think this should be enough without the generic types on the setter/getter, sounds good! 👍👍

daneryl (approved):

PR #3427, opened 2021-01-15
3426 new user without password error

### LINKED ISSUE #3426: Error when a user is created without a password
There is an error when a new user is created without a password

grafitto:
fixes #3426 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [x] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/users/specs/users.spec.js
+++ b/app/api/users/specs/users.spec.js
@@ -121,26 +121,19 @@ describe('Users', () => {
           .catch(catchErrors(done));
       });
 
-      it('should create a random password when none is provided', done => {
-        users
-          .newUser(
-            {
-              username: 'someone',
-              email: 'someone@mailer.com',
-              role: 'admin',
-            },
-            domain
-          )
-          .then(() => {
-            expect(random.default).toHaveBeenCalled();
-            return users.get({ username: 'someone' });
-          })
-          .then(([user]) => {
-            expect(user.username).toEqual('someone');
-            expect(user.email).toEqual('someone@mailer.com');
-            done();
-          })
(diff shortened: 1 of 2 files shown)


grafitto:
The failing e2e tests seem to be unpredictable, since there is no guarantee for the order of the items in the array being tested, the tests sometimes pass and other times fail because the order of items is not guaranteed to remain the same, weirdly enough the tests seem to pass morning hours my time 😆  @txau @RafaPolit

app/api/users/specs/users.spec.js
```diff
@@ -119,6 +121,23 @@ describe('Users', () => {
...
+          .newUser(
+            {
+              username: 'someone',
+              email: 'someone@mailer.com',
+              role: 'admin',
+            },
+            domain
+          )
+          .then(() => {
+            expect(random.default).toHaveBeenCalled();
```
konzz, line 135: Since you are mocking the random object, it would be great to expect that a User is created, and has the mock value 'mypass'

grafitto: Okay. On it

grafitto: The user returned does not include the password. I will just check whether the user is actually created

konzz: Mongodb does not return the user password by default, you have to 'select' it on your query to mongo.

grafitto: Aah okay

grafitto: Done @konzz

PR #3434, opened 2021-01-15
Hotfix snippets search

fnocetti:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/search/specs/search.spec.js
+++ b/app/api/search/specs/search.spec.js
@@ -65,25 +65,21 @@ describe('search', () => {
       expect(snippets.fullText[0].text).not.toMatch('[[34]]');
     });
 
-    it('perform a search on metadata and fullText and return the snippets', done => {
+    it('perform a search on metadata and fullText and return the snippets', async () => {
       const user = { _id: 'userId' };
-      search
-        .searchSnippets('gargoyles', ids.metadataSnippets, 'en', user)
-        .then(snippets => {
-          const titleSnippet = snippets.metadata.find(snippet => snippet.field === 'title');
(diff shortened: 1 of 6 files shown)


app/api/search/specs/search.spec.js
```diff
@@ -85,16 +86,27 @@ describe('search', () => {
...
+    it('should not include unpublished if not logged in', done => {
+      search
+        .searchSnippets('unpublished', 'unpublishedSharedId', 'en', undefined)
+        .then(snippets => {
+          expect(snippets.fullText.length).toBe(0);
+          done();
+        })
+        .catch(catchErrors(done));
+    });
+
```
RafaPolit, line 109: Two observations here:
- please do not create new tests with the `done` flow, use async (and perhaps refactor the previous test?).  There is no longer need for the catchErrors, etc. It's much cleaner.
- I think it's dangerous to have an all-or-nothing test.  The first test tests that all documents come through, the second test that none come through.  If there were a coding bug that prevented any result to come through, test two would still pass green.  So please, make it so that test one has TWO results and test two yields ONE result so that we know that: A. Some results are coming, B. Partial results are coming through.

RafaPolit (changes requested):
Please, review the comment and implement the changes.

app/api/search/specs/search.spec.js
RafaPolit: Well, for the record, I'm a complete moron: this endpoint is intended for single documents.  Disregard the second part of the comment.

PR #3443, opened 2021-01-20
updateMapping and reindex error now will be a 500

### LINKED ISSUE #3440: Log reindex errors as 500 instead of 400
These errors are critical, they need to be logged properly.

daneryl:
fixes #3440 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/search/entitiesIndex.js
+++ b/app/api/search/entitiesIndex.js
@@ -4,7 +4,6 @@ import entities from 'api/entities';
 import errorLog from 'api/log/errorLog';
 import { entityDefaultDocument } from 'shared/entityDefaultDocument';
 import PromisePool from '@supercharge/promise-pool';
-import { createError } from 'api/utils';
 import elastic from './elastic';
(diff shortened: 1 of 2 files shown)


txau:
@daneryl if I'm understanding the code correctly just removing the try/catch block will bubble up the error to the middleware and it will be reported as 500 instead of 400?

daneryl:
> @daneryl if I'm understanding the code correctly just removing the try/catch block will bubble up the error to the middleware and it will be reported as 500 instead of 400?

correct, that is the default behaviour, maybe we should check if this methods are being used outside the context of express.

daneryl:
> > @daneryl if I'm understanding the code correctly just removing the try/catch block will bubble up the error to the middleware and it will be reported as 500 instead of 400?
> 
> correct, that is the default behaviour, maybe we should check if this methods are being used outside the context of express.

checked, we should be good.

PR #3465, opened 2021-02-09
toc_generation service

### LINKED ISSUE #3447: toc auto generated integration
~~receives tenant and task (toc generation)
calls external toc generation external service (http)
blocks specified task until its finished, the same task for the same tenant can not be run at the same time. (task should be blocked in a shared resource to support uwazi multiprocess architectures)
on toc generated, it does save the toc on the file and sets a "Machine Generated" flag
reindex related file entity with the machine generated flag to be able to aggregate and filter
add number of page on the toc UI
do not change size of the toc sidepanel (will be a separate development)
do not implement the next/prev UI to see a selection of entities (will be a separate development)~~

daneryl:
fixes #3447 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


daneryl:
missing stuff
- [ ] label "auto-generated" on the TOC header sidepanel (css)
- [ ] label "auto-generated" on the TOC per each file on entity sidepanel (css)
- [ ] "review toc" button icon
- [ ] filter (css)
- [ ] number of page on the toc
- [ ] do not crash app when toc service throws an error (unavailable)

daneryl:
fixes #3447 

setup for qa:

- ```yarn blank-state```
- ```mongo localhost/uwazi_development --eval 'db.settings.updateOne({}, {$set: {features: {tocGeneration: {url:"localhost:5000/process"}}}})'```
- start a mock tock generation service (mock nodejs http server on the botom of the comment)
- restart uwazi
- upload some files and  wait for tocs to be auto generated (around 10s per toc)

```
const http = require('http');

const toc = [
  {
    selectionRectangles: [
      {
        top: 335.953125,
        left: 94.453125,
        width: 581.78466796875,
        height: 40,
        page: '1',
      },
      {
        top: 383.953125,
        left: 94.453125,
        width: 176.82577514648438,
        height: 40,
        page: '1',
      },
    ],
    label: 'Title1',
    indentation: 0,
  },
  {
    selectionRectangles: [
      {
        top: 97.1875,
        left: 123.453125,
        width: 323.6376953125,
        height: 24,
        page: '3',
      },
    ],
    label: 'subtitle 1',
    indentation: 1,
  },
  {
    selectionRectangles: [
      {
        top: 97.1875,
        left: 123.453125,
        width: 179.54669189453125,
        height: 24,
        page: '4',
      },
    ],
    label: 'subtitle 2',
    indentation: 1,
  },
  {
    selectionRectangles: [
      {
        top: 91.609375,
        left: 94.453125,
        width: 28.203262329101562,
        height: 23,
        page: '12',
      },
      {
        top: 91.609375,
        left: 132.453125,
        width: 322.78369140625,
        height: 23,
        page: '12',
      },
    ],
    label: 'title2',
    indentation: 0,
  },
];

const requestListener = function(_req, res) {
  console.log('toc generated!');
  res.setHeader('Content-Type', 'application/json');
  res.end(JSON.stringify(toc));
};
const server = http.createServer(requestListener);
server.listen(5000);

```

app/react/ToggledFeatures/tocGeneration/specs/FilterTocGeneration.spec.tsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import { shallow, ShallowWrapper } from 'enzyme';
+import SelectFilter from 'app/Library/components/SelectFilter';
+import { FilterTocGeneration } from '../FilterTocGeneration';
+import { Aggregations } from 'shared/types/Aggregations';
+
+describe('ReviewTocButton', () => {
```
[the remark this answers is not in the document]
daneryl: renamed 👍

app/api/search/elasticTypes.ts
```diff
@@ -1,5 +1,6 @@
 import { RequestParams } from '@elastic/elasticsearch';
 import { RequestBody } from '@elastic/elasticsearch/lib/Transport';
+import { Aggregations } from 'shared/types/Aggregations.d.ts';
```
RafaPolit, line 3: I don't think the extensions are required?

app/api/toc_generation/tocService.ts
```diff
@@ -0,0 +1,46 @@
+import { files, uploadsPath } from 'api/files';
+import { prettifyError } from 'api/utils/handleError';
+import errorLog from 'api/log/errorLog';
+import request from 'shared/JSONRequest';
+import entities from 'api/entities';
+
+const tocService = (serviceUrl: string) => ({
```
RafaPolit, line 7: Is the parenthesis here required? We use this as the shorthand way to return directly an object instead of opening a function, but here it does not appear to be the case?  I am confused.

app/api/toc_generation/tocService.ts
```diff
@@ -0,0 +1,46 @@
...
+    );
+
+    try {
+      if (nextFile) {
+        const toc = await request.uploadFile(
+          serviceUrl,
+          nextFile.filename,
+          uploadsPath(nextFile.filename)
+        );
+        await files.save({ ...nextFile, toc, generatedToc: true });
```
RafaPolit, line 29: Should this line fail, the error log will log the error, but subsequent calls will fetch the same file over and over again.  Do we want that to be the case? If this is only logging errors and not failing, how is the caller supposed to know something went wrong?

app/react/App/scss/modules/_search.scss
```diff
@@ -233,12 +233,23 @@
...
+  }
+}
+
 .search__filter {
   list-style-type: none;
   margin: 0;
   padding: 15px;
   color: $c-grey-dark;
 
+
```
RafaPolit, line 252: Probably remove this line?

app/react/Documents/components/DocumentSidePanel.js
```diff
@@ -1,9 +1,11 @@
+/* eslint-disable max-lines */
```
RafaPolit, line 1: Do we want to commit this on non-spec files?  It's indispensable while working, but having a "reminder" that this is an overly long file may be useful?

app/react/Documents/components/ShowToc.js
```diff
@@ -7,6 +7,8 @@ import ShowIf from 'app/App/ShowIf';
 import { t } from 'app/I18N';
 import { Icon } from 'UI';
 
+import './scss/showToc.scss';
```
RafaPolit, line 10: Have we moved away from creating scss files in the App folder namespaced by either component type or section type?  Also, usually scss files are imported in a single place instead on a per-component basis.  Is this actually working?  We don't really have styled components, which would be ideal.  But not sure this is the right approach?

app/react/Documents/components/scss/showToc.scss
```diff
@@ -0,0 +1,52 @@
...
+          } @else if $i == 1 {
+            font-weight: 400;
+            font-size: 14px;
+            line-height: 16.41px;
+            color: #333;
+          } @else if $i == 2 {
+            font-weight: 400;
+            font-size: 14px;
+            line-height: 16.41px;
+            color: #555;
```
RafaPolit, line 39: If we are going to have `if`s here, was it really necessary the `for`?  This is a very complex CSS to read. If we had hundreds of indents, maybe it could be worth it.  But for 5, maybe just putting:
```
li.toc-indent-1, 
li.toc-indent-2,
li.toc-indent-3,
li.toc-indent-4,
li.toc-indent-5 {
...
```

was enough?   Not sure here.

app/react/Documents/components/scss/showToc.scss
```diff
@@ -0,0 +1,52 @@
+@import "../../../App/scss/config/_colors.scss";
+
+.toc {
+  padding-left: 5px !important;
+  padding-right: 5px !important;
+
+  ul.toc-view {
+    $indents: 5;
+
+    @for $i from 0 through $indents {
```
RafaPolit, line 10: Not really sure the usefulness of a variable with fixed value used in the very next line?  `from 0 through 5` should be enough?

app/react/Documents/components/scss/showToc.scss
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+}
+
+.page-number {
+  position: relative;
+  float: right;
+  font-weight: normal;
+  padding-right: 4px;
+}
```
RafaPolit, line 52: Missing enter

app/react/Documents/components/scss/showToc.scss
```diff
@@ -0,0 +1,52 @@
+@import "../../../App/scss/config/_colors.scss";
```
RafaPolit, line 1: Why is this being imported? None of the color constants are being used in the file.

app/react/Documents/components/scss/toc.scss
```diff
@@ -0,0 +1,22 @@
+@import '../../../App/scss/config/_colors.scss';
```
RafaPolit, line 1: Again, not sure why is this imported?

app/react/Library/actions/specs/libraryActions.spec.js
```diff
@@ -157,6 +160,7 @@ describe('libraryActions', () => {
 
     afterEach(() => backend.restore());
 
+    //eslint-disable-next-line max-statements
```
RafaPolit, line 163: :(

RafaPolit (changes requested):
Half way through the files, this is what I have so far.  Some questions, some suggestions.

app/react/ToggledFeatures/tocGeneration/TocGeneratedLabel.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { FileType } from 'shared/types/fileType';
+
+export interface TocGeneratedLabelProps {
+  file: FileType;
+  children: JSX.Element | string;
+}
+
+export const TocGeneratedLabel = ({ file, children }: TocGeneratedLabelProps) => (
+  <FeatureToggle feature="tocGeneration">
+    {file.generatedToc && <span className="badge">{children}</span>}
```
RafaPolit, line 12: If children can be a JSX.Element, `<span>` only allows a very few children. Should this may be an inline div? Should it just accept strings?

app/react/ToggledFeatures/tocGeneration/specs/FilterTocGeneration.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+      expect(options).toEqual([
+        expect.objectContaining({ value: true, results: 0 }),
+        expect.objectContaining({ value: false, results: 2 }),
+      ]);
+    });
+  });
+
+  describe('when aggregations are not defined/complete', () => {
+    it('should not fail (render blank state)', () => {
+      //@ts-ignore
```
RafaPolit, line 52: Why is this ignore required? Because aggregations do not accept empty objects?

app/react/ToggledFeatures/tocGeneration/specs/ReviewTocButton.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const store: MockStore<object> = mockStoreCreator({});
+    component = shallow(
+      <Provider store={store}>
+        <ReviewTocButton file={{ ...file, _id: 'id' }}>
+          <span>test</span>
+        </ReviewTocButton>
+      </Provider>
+    )
+      .dive()
+      .dive();
```
RafaPolit, line 22: We have been requiring this wrapper with store and dual `dive()` calls on a lot of new components.  Just like an `idea`, we may want to look into abstracting this.

app/react/ToggledFeatures/tocGeneration/specs/actions.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+        _id: fileId,
+        generatedToc: false,
+      },
+      documents: [
+        {
+          _id: fileId,
+          generatedToc: false,
+        },
+      ],
+    };
```
RafaPolit, line 59: Maybe refactor into a quick function that returns either the true or false values and all else being the same instead of this?  Not a really important suggestion, not really needed if time is of the essence.

app/shared/types/settingsSchema.ts
```diff
@@ -133,7 +133,14 @@ export const settingsSchema = {
       type: 'object',
       properties: {
         _id: { type: 'string' },
-        semanticSearch: { type: 'boolean' },
+        tocGeneration: {
+          type: 'object',
+          required: ['active', 'url'],
```
RafaPolit, line 138: Is this right? You have a required property that is not defined in properties?

emitSchemaTypes.js
```diff
@@ -61,30 +61,36 @@ const writeTypeFile = (file, commonImport, snippets) => {
...
+      }
+      return compile(schema, schema.title || firstUp(name), opts);
+    })
+  );
+
+  const contents = fs.readFileSync(file).toString();
 
-    const schemas = require(`./${file}`);
+  writeTypeFile(file, typeImports(contents.match(typeImportFindRegex)), snippets);
+};
```
RafaPolit, line 77: Can you quickly explain what is the purpose of this refactor / change? Thanks.

RafaPolit (changes requested):

app/api/entities/entities.js
```diff
@@ -163,6 +163,10 @@ async function updateEntity(entity, _template) {
       if (typeof entity.template !== 'undefined') {
         d.template = entity.template;
       }
+
+      if (typeof entity.template !== 'undefined') {
+        d.generatedToc = entity.generatedToc;
+      }
```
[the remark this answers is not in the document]
daneryl: because it should not be the same condition.
added test to ensure the param is not copied when its undefined with the right condition. 👍

app/api/toc_generation/tocService.ts
daneryl: this function actually returns an object, its supposed to be an "instance" method.

app/api/search/elasticTypes.ts
daneryl: it was not required, fixed, weird in my initial implementation i was getting an error.

app/react/Documents/components/DocumentSidePanel.js
daneryl: totally agree, removed.

app/react/Documents/components/ShowToc.js
daneryl: im not sure either, this definitely works, and it also make sense, the css go with the component, if you remove this component this css will not be part of the final build instead of having to remember that this have some associated specific css.

app/react/Documents/components/scss/showToc.scss
daneryl: @grafitto can you check this please ?

app/react/Documents/components/scss/showToc.scss
daneryl: @grafitto can you check this please ?

app/react/Documents/components/scss/showToc.scss
daneryl: @grafitto can you check this please ?

app/react/Documents/components/scss/toc.scss
daneryl: was not needed, removed.

app/react/ToggledFeatures/tocGeneration/TocGeneratedLabel.tsx
daneryl: this was to allow <Translate> or a string, since it can not be specified [which element is allowed as a TS type](https://react-typescript-cheatsheet.netlify.app/docs/advanced/patterns_by_usecase#what-you-cannot-do)
i have changed it for a div as you suggested, badge is "inline-block"

app/react/ToggledFeatures/tocGeneration/specs/FilterTocGeneration.spec.tsx
daneryl: correct, i want to still be sure that an empty object does not make this fail.

app/react/ToggledFeatures/tocGeneration/specs/ReviewTocButton.spec.tsx
daneryl: we can think about it, the problem i faced when i have tried this is that the number of dives can vary and sometime you want the context or not in the dive, as an example:

https://github.com/huridocs/uwazi/blob/37d91765babc2e6b685a584d39f10e578777c2e7/app/react/Templates/components/specs/MetadataProperty.spec.js#L80-L92

but maybe this is wrong and the 2 dive scenario with Provider->Component is generic enough ? should i create a tech debt issue for this ?

app/react/ToggledFeatures/tocGeneration/specs/actions.spec.ts
daneryl: done 👍

app/shared/types/settingsSchema.ts
daneryl: this is not right, removed active from required 👍

emitSchemaTypes.js
daneryl: this refactor allows for files without the name convention "*Schema" to be also used to generate types as long as they are inside shared/types, so now the convention is SomethingSchema or whatever filename inside shared/types.
specifically this was made to generate the types for https://github.com/huridocs/uwazi/blob/d2fc5a99c6c6a8525f3f8e76c559d2f445778b0c/app/shared/types/searchParams.ts

app/api/files/files.ts
```diff
@@ -30,4 +31,27 @@ export const files = {
 
     return toDeleteFiles;
   },
+
+  async tocReviewed(_id: string, language: string) {
+    const savedFile = await files.save({ _id, generatedToc: false });
+    const sameEntityFiles = await files.get({ entity: savedFile.entity }, { generatedToc: 1 });
+    const [entity] = await entities.get({
+      sharedId: savedFile.entity,
+    });
```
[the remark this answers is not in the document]
daneryl: i do not see any benefit other than a really small performance impact ?, adding a condition will add complexity. do you foresee any problems saving entities with the same values ?

app/api/toc_generation/tocService.ts
daneryl: it is not supposed to, error logs should be enough for now, we designed a different system to handle this kinds of problems, but it was too complex for this initial testing version. the main reason the error is handled here is to prevent a crash on "service not online" error as this runs outside the normal request/response flow.

do you have any suggestion to quickly make this more resilient ?

app/react/Library/actions/specs/libraryActions.spec.js
daneryl: i do not think this rule has that much value here, and i do not know how to properly fix it, wrapping some "it(" on describes will fix the problem but this is obviously not why we have this rule, anyways, one of the variables was not needed to be declared in this scope, removing it solved the issue for this particular case. :)

app/react/Documents/components/scss/showToc.scss
grafitto: Removed

app/react/Documents/components/scss/showToc.scss
grafitto: True, I will just get rid of the loop all together

grafitto: Removed the loop, actually there are only 3 levels not 5

app/react/Documents/components/scss/showToc.scss
grafitto: Removed the loop completely

app/api/files/files.ts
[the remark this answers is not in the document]
RafaPolit: My concern is that conditionals should happen at some point.  So, if not here, then the entities save, when updating all the connected entities, should know: hey, nothing changed.  But if this save actually triggers an full entity check and reindex on all connected entities, I would rather have the conditional here.  If you are confident that is not going to happen, I agree a single save (or potentially n entities depending on languages) is not a huge deal.

app/react/ToggledFeatures/tocGeneration/specs/ReviewTocButton.spec.tsx
RafaPolit: Let's wait and see if they keep popping up. Let's keep an eye on future PRs and our own work.  For example, on this file you quoted, we could return the two-dive component, and the external file does the extra dives, for example?

RafaPolit (approved):

app/react/ToggledFeatures/tocGeneration/specs/ReviewTocButton.spec.tsx
daneryl: yes i guess we can do that on this file, lets track this on a tech debt issue ?

app/api/toc_generation/tocService.ts
daneryl: @RafaPolit "better" error handling is now implemented, this is basic error handling taking into account this is a beta feature.

1. when service returns empty toc: a fake toc is generated labeled "empty toc was generated"
2. when any error except "service not available": a fake toc is generated labeled with the error message
3. when service not available: does nothing, keeps trying until it is available.

app/api/files/files.ts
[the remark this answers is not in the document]
daneryl: @RafaPolit im not confident its not going to happen, but if saving one entity can be this problematic i believe  this should be handled at a lower level, not anytime we need to save one ? tech debt for this pls ?

app/react/Documents/components/DocumentSidePanel.js
```diff
@@ -340,6 +345,15 @@ export class DocumentSidePanel extends Component {
               />
             </TabContent>
             <TabContent for="toc" className="toc">
+              <div className="tocHeader">
+                <h1>
+                  <Translate>Table of contents </Translate>
+                </h1>
+                &nbsp;
+                <TocGeneratedLabel file={this.props.file}>
+                  <Translate>auto-created ⓘ </Translate>
```
konzz, line 354: Probably keep the ⓘ outside of the translate?

app/react/ToggledFeatures/tocGeneration/FilterTocGeneration.tsx
```diff
@@ -0,0 +1,49 @@
...
+const filteredAggregation = (aggregations: Aggregations, key: string) => {
+  const bucket = (aggregations?.all?.generatedToc?.buckets || []).find(a => a.key === key) || {
+    filtered: { doc_count: 0 },
+  };
+  return bucket.filtered.doc_count;
+};
+
+const options = (aggregations: Aggregations = { all: {} }) => [
+  {
+    label: 'Automatically generated',
```
konzz, line 23: should we use t function to make those labels translatable at least in the translations form? or maybe in another issue update the Multiselect component to wrap labels arround <Translate> ?

app/shared/JSONRequest.js
```diff
@@ -117,7 +117,7 @@ export default {
 
   delete: (url, data, headers) => _fetch(url, data, 'DELETE', headers),
 
-  // TEST!!! Fully untested function
+  // TEST!!!! Fully untested function
```
konzz, line 120: 👍

konzz: Remember that after the 4th exclamation we should use question marks: `?`

app/react/Documents/components/DocumentSidePanel.js
daneryl: @konzz this allows not only to translate but to edit, in case you do not want the ⓘ, i think its good to have it there ?

app/react/ToggledFeatures/tocGeneration/FilterTocGeneration.tsx
daneryl: done, i have translated this labels

PR #3469, opened 2021-02-10
Removed toJS calls on getMarkers and unnecessary find calls

### LINKED ISSUE #3466: Remove toJS and refactor find calls on getMarkers function
Make the getMarkers function faster, probably there is other reasons why is slow but these are the obvious ones.
1. There is abusive and repeated usage of .toJS inside
2. here is repeated calls to .find also

The multiple calls seems to be because of not needed re-renders, ideally the function gets called only once.

A subtask for #3444

grafitto:
fixes #3466 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Library/components/MapView.js
+++ b/app/react/Library/components/MapView.js
@@ -69,7 +69,7 @@ export class MapView extends Component {
             <b>{markers.get('totalRows')}</b> {t('System', 'documents')}
           </span>
         </div>
-        <Markers entities={markers.get('rows')}>
+        <Markers entities={markers.get('rows')} storeKey={storeKey}>
           {processedMarkers => (
             <Map
               ref={ref => {
--- a/app/react/Library/components/MapView.js
+++ b/app/react/Library/components/MapView.js
@@ -69,7 +69,7 @@ export class MapView extends Component {
             <b>{markers.get('totalRows')}</b> {t('System', 'documents')}
           </span>
         </div>
-        <Markers entities={markers.get('rows')} storeKey={storeKey}>
+        <Markers entities={markers.get('rows')}>
           {processedMarkers => (
             <Map
               ref={ref => {
--- a/app/react/Map/helper.js
+++ b/app/react/Map/helper.js
@@ -1,6 +1,3 @@
-/* eslint-disable max-statements */
-/** @format */
-
 const _mergeArrays = (first, second) => {
   let bigArray;
   let smallArray;
--- a/app/react/Map/helper.js
+++ b/app/react/Map/helper.js
@@ -57,51 +57,76 @@ const markersToStyleFormat = markers =>
 
 const noop = () => {};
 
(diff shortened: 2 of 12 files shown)


app/react/Library/components/MapView.js
```diff
@@ -69,7 +69,7 @@ export class MapView extends Component {
             <b>{markers.get('totalRows')}</b> {t('System', 'documents')}
           </span>
         </div>
-        <Markers entities={markers.get('rows')}>
+        <Markers entities={markers.get('rows')} storeKey={storeKey}>
```
konzz, line 72: Why does the Markers component need the storeKey property?

app/react/Map/helper.js
```diff
@@ -41,105 +57,106 @@ const markersToStyleFormat = markers =>
...
-          .forEach(({ value: point }) => {
-            const { lat, lon } = point;
-            markers.push({
-              properties: { entity, color, info: point.label },
-              latitude: lat,
-              longitude: lon,
-              label,
+  const geolocationPropNames = geolocationProps.map(prop => prop.get('name'));
+
+  if (geolocationProps.size && metadata) {
```
konzz, line 64: Please refactor the nested if and else into functions

app/react/Map/specs/__snapshots__/helper.spec.js.snap
```diff
@@ -3,11 +3,12 @@
 exports[`map helper getMarkers() Inheritance should include inherited properties 1`] = `
```
konzz, line 3: Change the spec that is using this snapshot to test that the inherited property is in the array of markers, an snapshot here is confusing and hard to tell when the snapshot is correct

konzz (changes requested):

app/react/Map/helper.js
```diff
@@ -1,5 +1,21 @@
+/* eslint-disable max-statements */
 /** @format */
```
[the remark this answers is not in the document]
grafitto: Okay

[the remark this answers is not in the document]
grafitto: Removed

app/react/Library/components/MapView.js
grafitto: It actually does not need storeKey, I forgot to remove when I realized entities are passed as a property. Removing...

grafitto: Removed

app/react/Map/helper.js
grafitto: Done

app/react/Map/specs/__snapshots__/helper.spec.js.snap
grafitto: Done, kindly check if that is what is needed.

PR #3482, opened 2021-02-23
3480 improved settings collection UI

### LINKED ISSUE #2938: Improve the help and tips provided in Settings -> Collection
The Collection settings is filled with difficult-to-understand fields: "Public Form Destination", "Matomo Configuration", "MapTiler Key".

As a temp solution for this, we created this blue "blobs" that explain a bit what a field does.  Yet, as this grows in size and complexity, perhaps this is not the right route to go.

Instead of trying new approaches on each value we add, perhaps we can come up with a "unified" way of showing in-line help (not in external documents?). @simonfossom

### LINKED ISSUE #3480: Improved Settings Collection UI
17 Points AVG

- [x] Apply the designs for the new collection settings.
- [x] Check Mobile
- [x] Migrate the form  library (react-redux-forms) to react-hook-form
- [x] Split the form in smaller components

https://www.figma.com/file/xjsz7r65epN4B0b96zilac/0-uwazi-redesigns?node-id=1554%3A1

Zasa-san:
fixes #3480
fixes #2938

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Settings/components/SettingsLabel.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+export interface SettingsLabelProps {
+  children: React.ReactNode;
+  className?: string;
+}
+
+export const SettingsLabel = ({ children, className }: SettingsLabelProps) => (
+  <div className={`inline-form-label ${className}`}>{children}</div>
```
daneryl, line 9: i think the label should be a ```<label>``` ?

app/react/Settings/components/ToggleChildren.tsx
```diff
@@ -0,0 +1,39 @@
...
+    }
+    setToggled(!isToggled);
+  };
+
+  return (
+    <>
+      <div className="toggle-children-button">
+        <ToggleButton onClick={handleClick} checked={isToggled} />
+      </div>
+      <div className="toggle-children-children" style={{ display: isToggled ? 'block' : 'none' }}>
```
daneryl, line 34: @RafaPolit @Zasa-san this is still rendering the elements themselves, not a big issue but i think it should be preferabe to not render instead of display: none ?

app/react/Settings/components/specs/ToggleChildren.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { shallow, ShallowWrapper } from 'enzyme';
+import { ToggleChildren } from '../ToggleChildren';
+
+describe('Toggle children', () => {
+  let component: ShallowWrapper<typeof ToggleChildren> = shallow(
+    <ToggleChildren toggled={false} />
+  );
+
+  it('should not display children if not toggled', () => {
+    expect(component.find('.toggle-children-children').text()).toBe('');
```
daneryl, line 11: i believe this test is wrong, the reason why text() returns empty string is because there is no children in the component being tested
```
  let component: ShallowWrapper<typeof ToggleChildren> = shallow(
    <ToggleChildren toggled={false} />
  );

```

app/react/Settings/components/specs/ToggleChildren.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  it('should display children if toggled', () => {
+    component = shallow(
+      <ToggleChildren toggled>
+        <p>Content</p>
+      </ToggleChildren>
+    );
+    expect(component.find('.toggle-children-children').text()).toBe('Content');
+  });
```
daneryl, line 26: same as the previous test, the only reason why this shows content is because there is content, if you set toggled=false the test will still be green. this is because togglechildren does render the child and "hides" them via display:none

app/react/istore.d.ts
```diff
@@ -55,9 +56,9 @@ export interface OneUpState {
   reviewThesaurusValues: string[];
 }
 
-export interface SettingsState {
-  features?: { topicClassification: boolean };
-}
+// export interface SettingsState {
+//   features?: { topicClassification: boolean };
+// }
```
daneryl, line 61: not needed anymore ? remove pls :)

daneryl (changes requested):
@RafaPolit @Zasa-san apart from a couple of comments on the code i found out that the "custom starting location" is not being saved properly.

is the collectionSettings being removed and the V2 being renamed when this is finished ?

everyghing else LGTM !

app/react/Settings/components/ToggleChildren.tsx
RafaPolit: No.  Actually, that removed the reference on the form, so they need to be rendered without display or the bind between the input and the form hook is lost.  What you mention was our first approach and it didn't work, unfortunately. :(

app/react/Settings/components/SettingsLabel.tsx
RafaPolit: Looking into this...

RafaPolit:
> @RafaPolit @Zasa-san apart from a couple of comments on the code i found out that the "custom starting location" is not being saved properly.
> 
> is the collectionSettings being removed and the V2 being renamed when this is finished ?
> 
> everyghing else LGTM !

Indeed, Geolocation was missing the starting value. Thanks.

Yes, the idea is to replace the old collections with the new V2 and rename files.

app/react/Settings/components/SettingsLabel.tsx
RafaPolit: We are changing this to label, but the usefulness of <label> is precisely the `for` argument, which in this case is not being passed, so clicking on the label will not focus on the input at this point (and there are some properties that are not simple inputs or that we would not want the click to toggle on-off the buttons).

app/react/istore.d.ts
RafaPolit: Done

app/react/Settings/components/specs/ToggleChildren.spec.tsx
RafaPolit: Done

app/react/Settings/components/specs/ToggleChildren.spec.tsx
RafaPolit: Done

app/react/Settings/components/SettingsLabel.tsx
RafaPolit: Done

daneryl (approved):

PR #3490, opened 2021-03-02
Adds a migration to remove _id from filter items

### LINKED ISSUE #3485: Settings/Filters update throw a validation error when the instance has grouped filters
This is happening in instances that have grouped filters, created some time ago. Not possible to reproduce with a fresh instance (at least with the tested cases).
The object sent to the API contains an additional item property that's not allowed `_id` that contains the value of the property id.
The error occurs at saving Settings of Filters even without any changes.

To solve it's necessary to do a migration that removes this property.
Screen Shot 2021-02-26 at 10 46 38

grafitto:
fixes #3485 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [x] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/migrations/migrations/34-remove-_id-from-filter-items/specs/34-remove-_id-from-filter-items.spec.js
+++ b/app/api/migrations/migrations/34-remove-_id-from-filter-items/specs/34-remove-_id-from-filter-items.spec.js
@@ -30,25 +30,93 @@ describe('migration remove-_id-from-filter-items', () => {
   });
 
   it('should ignore items without _id property', async () => {
-    const filters = [
+    const settings = [
       {
-        items: [
+        filters: [
           {
-            id: 'someid',
-          },
-          {
-            id: 'someotherid',
+            items: [
+              {
+                id: 'someid',
+              },
+              {
+                id: 'someotherid',
+              },
+            ],
           },
         ],
       },
     ];
-    await testingDB.clearAllAndLoad({ settings: [...filters] });
+    await testingDB.clearAllAndLoad({ settings });
     await migration.up(testingDB.mongodb);
-    const settings = await testingDB.mongodb
+    const savedSettings = await testingDB.mongodb
       .collection('settings')
       .find({})
       .toArray();
 
-    expect(settings).toEqual(filters);
+    expect(savedSettings).toEqual(settings);
+  });
+
+  it('should ignore collections without filters', async () => {
+    const settings = [
+      {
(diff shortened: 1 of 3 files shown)


app/api/migrations/migrations/34-remove-_id-from-filter-items/specs/34-remove-_id-from-filter-items.spec.js
```diff
@@ -0,0 +1,31 @@
+import testingDB from 'api/utils/testing_db';
+import migration from '../index.js';
+import fixtures from './fixtures.js';
+
+describe('migration remove-_id-from-filter-items', () => {
+  beforeEach(async () => {
+    spyOn(process.stdout, 'write');
+    await testingDB.clearAllAndLoad(fixtures);
```
RafaPolit, line 8: The migration looks good, and I believe it works as expected.

Still, I am concerned that the fixtures only show a single scenario: "A collection whose settings have filters, and those filters have nested items with _id present".

There are many other scenarios that need to be tested that they are not affected:
- Collections that don't have filters
- Collections that have filters and don't have nested items
- Collections that have filters with nested items and no _id present in there.

RafaPolit (changes requested):

RafaPolit:
While this migration may be needed, the fact that we have no clue as how to those _id's appeared there may leave us having to re-run an identical migration if this condition happens again. Please, spend some time trying to learn how this condition appeared in the first place, in order to first fix the problem, and only then run the migration.

app/api/migrations/migrations/34-remove-_id-from-filter-items/specs/34-remove-_id-from-filter-items.spec.js
grafitto: I have added a few test cases as suggested, you can check them out when you get time

grafitto:
> While this migration may be needed, the fact that we have no clue as how to those _id's appeared there may leave us having to re-run an identical migration if this condition happens again. Please, spend some time trying to learn how this condition appeared in the first place, in order to first fix the problem, and only then run the migration.

I have tried creating new filters and grouped filters, both do not add `_id` into the database, There is a high chance we are dealing with old data, maybe before the decision was made to remove the field from the database. However I am not sure whether there are other cases where filters are creating other than the settings > filters section. If so you can let me know so that I check it out.

PR #3511, opened 2021-03-09
2708-Entity-counter-limits

### LINKED ISSUE #2708: Counter on the main screen stops at 10'000 (3pts)
Uwazi seems to have a hard-coded imposed limit on its search to determine the counter on the main screen, which stops at 10'000 regardless of the number of records in a collection. 

This needs to be changed, especially for collections where there are 10'000+ records. 

Screenshot 2020-01-17 at 16 52 35

Zasa-san:
fixes #2708

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Layout/DocumentCounter.tsx
+++ b/app/react/Layout/DocumentCounter.tsx
@@ -6,7 +6,7 @@ interface EntityCounterProps {
   entityListCount: number;
   entityTotal: number;
   totalConnectionsCount: number;
-  maxedTotalEntities: string;
+  hitsTotalRelation: string;
 }
 
 export const DocumentCounter = (props: EntityCounterProps) => {
@@ -15,7 +15,7 @@ export const DocumentCounter = (props: EntityCounterProps) => {
     selectedEntitiesCount,
     entityListCount,
     entityTotal,
-    maxedTotalEntities,
+    hitsTotalRelation,
   } = props;
   const counter =
     totalConnectionsCount === undefined ? (
@@ -29,7 +29,7 @@ export const DocumentCounter = (props: EntityCounterProps) => {
         <b>
           {' '}
           {entityTotal}
-          {maxedTotalEntities === 'gte' && '+'}{' '}
+          {hitsTotalRelation === 'gte' && '+'}{' '}
         </b>{' '}
         <Translate>documents</Translate>
       </>
--- a/app/react/Layout/DocumentCounter.tsx
+++ b/app/react/Layout/DocumentCounter.tsx
@@ -1,11 +1,11 @@
 import { Translate } from 'app/I18N';
 import React from 'react';
 
-interface EntityCounterProps {
+export interface EntityCounterProps {
   selectedEntitiesCount: number;
   entityListCount: number;
   entityTotal: number;
-  totalConnectionsCount: number;
+  totalConnectionsCount?: number;
   hitsTotalRelation: string;
 }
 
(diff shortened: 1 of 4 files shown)


app/react/Layout/specs/DocumentCounter.spec.tsx
```diff
@@ -33,5 +34,9 @@ describe('DocumentCounter', () => {
       expect(component.children().get(6).props.children[1]).toBe(100);
       expect(component.children().get(8).props.children).toBe('documents');
     });
+
+    it('should show a + sign next to the total entities when maxedTotalEntities has the gte flag', () => {
+      expect(component.children().get(6).props.children).toContain('+');
+    });
```
RafaPolit, line 40: By passing the 'gte' as the only option in the props, there is no test to prove that the `+` is ever missing.

If you just returned a '+' instead of the logic on the component, all the tests here would be green as well.

That is why following TDD's guidelines is so useful.  In this case, you first create this test.  See it fail, and, what should have happened was to implement the minimum code to make it green: return the '+'.

Then, in order to implement further functionality, you would have required the missing test of: 'when maxedTotalEntities is NOT gte'.

Please, remove the 'gte' from the sent values in props, implement a test proves that only the correct amount is shown, and then implement a test that sends the `gte` with this expect.

But more importantly, please, use TDD on this simpler issues so that it becomes second nature to your programming flow. Otherwise, it is never going to happen.

app/react/Layout/DocumentCounter.tsx
```diff
@@ -19,13 +26,18 @@ export const DocumentCounter = (props: EntityCounterProps) => {
           </>
         )}
         <b> {entityListCount} </b> <Translate>shown of</Translate>
-        <b> {entityTotal} </b> <Translate>documents</Translate>
+        <b>
+          {' '}
+          {entityTotal}
+          {maxedTotalEntities === 'gte' && '+'}{' '}
+        </b>{' '}
```
RafaPolit, line 33: I'm not sure I like the structure of these `{' '}`: there is one inside the <b>, then one in the same line before the </b> and then one again after the </b>.  Can we control this with paddings on the CSS instead of this random spaces here?

app/react/Layout/DocumentCounter.tsx
```diff
@@ -6,10 +6,17 @@ interface EntityCounterProps {
...
 }
 
 export const DocumentCounter = (props: EntityCounterProps) => {
-  const { totalConnectionsCount, selectedEntitiesCount, entityListCount, entityTotal } = props;
+  const {
+    totalConnectionsCount,
+    selectedEntitiesCount,
+    entityListCount,
+    entityTotal,
+    maxedTotalEntities,
```
RafaPolit, line 18: I am unsure about the semantics of this variable:  the values could be 'eq' or 'gte'.  So, `maxedTotalEntities greater than` is not really correct.  Should we keep the original naming and go with something like `hitsTotalRelation`?  Something else?

RafaPolit (changes requested):

app/react/Layout/DocumentCounter.tsx
Zasa-san: Yes, I struggled to come up with a reference that made semantic sense. Lets go with hitsTotalRelation

app/react/Layout/DocumentCounter.tsx
Zasa-san: I think in this case is easier to add a space after _show of_ and another before _documents_ instead of CSS padding, since there are supposed to be a nbsp between this elements of the component.

RafaPolit:
LGTM

PR #3515, opened 2021-03-10
Encapsulated tenant for error and debug logs

### LINKED ISSUE #3430: add tenant name on the debug.log messages like the errorLog have.

RafaPolit:
fixes #3430

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/utils/specs/__snapshots__/handleError.spec.js.snap
```diff
@@ -12,13 +12,14 @@ test error",
...
-    Object {
-      "pretty": "
+    "
 body: {
  \\"username\\": \\"########\\",
  \\"password\\": \\"########\\"
 }
 test error",
+    Object {
+      "shouldBeMultiTenantContext": true,
```
daneryl, line 22: @RafaPolit why do we include this info on the error ? this is supposed to be internal? or do you think its going to be useful for debugging ?

RafaPolit: The handle error is bubbling up this into the transport in order to add the tenant.  That is required, I think (and it was already there for the error log, I just combined it into the debug call.

The final error is neither reporting this nor logging this into the files.  Do you think that is good enough?

fnocetti (approved):
This looks really good. Great improvement on the logs output!!
Everything works smoothly. The ES errors are also showing the complete information.
Tested for multiple tenants and default tenant (which is of course shown as localhost).
Approving and merging.

PR #3516, opened 2021-03-11
Attachments mimetype

konzz:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/shared/specs/JSONRequest.spec.js
```diff
@@ -303,4 +303,21 @@ describe('JSONRequest', () => {
...
+      };
+      backend
+        .restore()
+        .head('http://localhost:3000/hello', JSON.stringify({ response: mockresponse }));
+
+      const response = await request.head('http://localhost:3000/hello');
+      const { method } = backend.calls()[0][1];
+
+      expect(method).toBe('HEAD');
+      expect(response.headers).toBeDefined();
```
fnocetti, line 320: Do you think it would be also useful to assert that there should be not body, which is kind-of the reason behind using head?

fnocetti (commented):
The code LGTM! Left one quick comment.
Proceeding to the QA.

app/api/migrations/migrations/34-move-attachments/index.js
```diff
@@ -15,14 +15,15 @@ export default {
       const entity = await cursor.next();
 
       await Promise.all(
-        entity.attachments.map(async ({ filename, originalname }) =>
+        entity.attachments.map(async ({ filename, originalname, mimetype }) =>
```
fnocetti, line 18: @konzz Bonus comment: this change makes sense as long as we don't deploy the PR in which this migration was included. Did we deploy it already?

app/shared/specs/JSONRequest.spec.js
konzz: Well, it's not JSONrequest responsibility to return or not the body, it's part of the HTTP protocol. What I tested is that the correct method is used, and making sure we return the headers if they are present. 

Anyway we are mocking the backend that responses to this calls, and you can actually return a body in a HEAD request, so in order to make it green I just need to dont send a body in the mock, but thats not testing anything from JSONrequest.

Also I could add an if to check the method is "HEAD" and do not return the body, but the point of using HEAD is to avoid large downloads and at that point you already downloaded the body so there is no point in trying to not return it.

app/api/migrations/migrations/34-move-attachments/index.js
konzz: Yes, when I did it, it was not deployed yet, I removed it since it will need to go in a new migration anyway

PR #3534, opened 2021-03-16
Configured uploadFile to be able to accept external cookie.

### LINKED ISSUE #3532: Improve JSONRequest's uploadFiles with remote cookie instead of singleton value
Currently, our JSONRequest object is a singleton and uploadFiles allows for a single cookie upon request to be configured.  Therefore, different services (like different remote Syncs) are failing.

First approach is to allow for a custom cookie, and later on we will refactor this to be a better reusable service.

### LINKED ISSUE #3533: Sync is not sending individual cookies per service, therefore log-in is not respected
When multiple sync targets are configured per instance, log-in is failing to keep headers upon requests from different services.

This is preventing correct sync of multiple services.

RafaPolit:
fixes #3532
fixes #3533

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/sync/specs/syncWorker.spec.js
+++ b/app/api/sync/specs/syncWorker.spec.js
@@ -79,14 +79,14 @@ describe('syncWorker', () => {
 
   const syncWorkerWithConfig = async config =>
     syncWorker.syncronize({
-      url: 'url',
+      url: 'url-slave1',
       name: 'slave1',
       config,
     });
 
   const syncAllTemplates = async (name = 'slave1', batchSize) =>
     syncWorker.syncronize({
-      url: 'url',
+      url: `url-${name}`,
       name,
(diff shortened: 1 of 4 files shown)


fnocetti (approved):
CR LGTM! Nice job, I like the approach.
I'll try to perform a quick testing.

fnocetti:
Works like charm! Tested:
On a multitenant set-up using the default tenant as a master + 2 extra tenants as slaves.
Used different passwords for the admin users. Sync'ed 2 templates with different properties, attachments, documents, and used filter conditions for one of the slaves. 
Everything looks just good.
Merging now.

PR #3544, opened 2021-03-18
3492 share public in modal

### LINKED ISSUE #3492: Share with public group (14pts)
- Add the public option to the "share with" list of options in the share modal. (Only Admins and Editors will have this option)
- Remove the Publish button from all the context menus.
- This option will use the already existing publish flow, and wont be in the permissions array.

Screenshot 2021-03-02 at 18 19 00

fnocetti:
fixes #3492 

For the QAer: Intentionally left out the notice for when the user changes the publishing status. Refer to the issue for more information.

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


konzz:
Probably is a bit late in the development to say this, but since we already have an "auth" folder, probably most of the permissions feature should have been in the auth folder.

app/api/permissions/specs/entitiesPermissions.spec.ts
```diff
@@ -103,5 +176,24 @@ describe('permissions', () => {
         },
       ]);
     });
+
+    describe('publicly shared', () => {
+      it('should return a permission of type "public" and level "mixed" if the entities are published and unpublished', async () => {
+        const permissions = await entitiesPermissions.get(['shared1', 'shared4']);
+        expect(permissions).toEqual(
+          expect.arrayContaining([{ ...publicPermission, level: MixedAccess.MIXED }])
```
konzz, line 184: Shouldnt `MixedAccess.MIXED` be a constant of `Accesslevels`

app/api/permissions/entitiesPermissions.ts
```diff
@@ -51,51 +54,91 @@ async function setAccessLevelAndPermissionData(
...
   });
 
+  const publishedEntities = publishedData.filter(published => published).length;
+  const totalEntities = publishedData.length;
+
+  if (publishedEntities) {
+    permissionsData.push({
+      refId: 'public',
+      type: 'public',
+      label: 'Public',
```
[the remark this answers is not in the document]
fnocetti: @konzz Yeah, it was in my TODO list and I forgot. Good catch, thanks!

app/api/permissions/specs/entitiesPermissions.spec.ts
fnocetti: @konzz  It should, and should not :D 
The thing is: We should have two enums, one with and one without `MIXED`; because in some contexts mixed is allowed and in some others it isn't. But there's a limitation in TS for which you cannot 'extend' enums, and having duplicated values just because we wan't those two scenarios made no sense. So the solution is to have one enum for real permissions, one enum for the contextual permission (mixed access) and a type that merges both for typing things. Sadly, you cannot access the enum value from that merge. So that's why we use both enums to get values.

fnocetti:
> Probably is a bit late in the development to say this, but since we already have an "auth" folder, probably most of the permissions feature should have been in the auth folder.

@konzz yeah, it's a bit late 😬 
Anyways, almost everything in the permissions implementation is decoupled from the auth code. Even if it were not, I think authenticating a user, and securing entities with an (kind of)ACL implementation, are two different concerns.

app/react/Metadata/components/MetadataFormButtons.js
```diff
@@ -168,6 +170,7 @@ MetadataFormButtons.defaultProps = {
   formName: 'metadataForm',
   delete: () => {},
   copyFrom: () => {},
+  storeKey: undefined,
```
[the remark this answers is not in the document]
fnocetti: I took a look and the thing is that the MetadataFormButtons is used also in context where there is no storeKey (like the EntityViewer). Formerly, there wasn't even a need to pass a storeKey, but know we need it to dispatch actions when there is a storeKey.
These actions are meant to be designed so that they only have the implied effect if there is a storeKey available. Regarding this, I found a bug that I'll fix.

app/api/permissions/specs/entitiesPermissions.spec.ts
konzz: 👍

PR #3555, opened 2021-03-23
Testsuite improvements

daneryl:
fixes: #3561

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/files/filesystem.ts
+++ b/app/api/files/filesystem.ts
@@ -47,12 +47,34 @@ async function deleteFiles(files: FilePath[]) {
 const testingUploadPaths = {
   uploadedDocuments: `${__dirname}/specs/uploads/`,
   attachments: `${__dirname}/specs/uploads/`,
-  customUploads: `${__dirname}/specs/customUploads/`,
+  customUploads: `${__dirname}/specs/uploads/`,
   temporalFiles: `${__dirname}/specs/uploads/`,
 };
 
-const setupTestUploadedPaths = () => {
-  testingTenants.changeCurrentTenant(testingUploadPaths);
+const createDirIfNotExists = async (dirPath: string) => {
+  try {
+    await asyncFS.mkdir(dirPath);
+  } catch (e) {
+    if (!e.message.match(/file already exists/)) {
+      throw e;
+    }
+  }
+};
+
+const generateUploadsPath = async (subPath: string) => {
+  if (subPath) {
+    await createDirIfNotExists(`${__dirname}/specs/uploads/${subPath}`);
+  }
+  return {
+    uploadedDocuments: `${__dirname}/specs/uploads/${subPath}`,
+    attachments: `${__dirname}/specs/uploads/${subPath}`,
+    customUploads: `${__dirname}/specs/uploads/${subPath}`,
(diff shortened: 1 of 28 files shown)


app/api/files/specs/jsRoutes.spec.js
```diff
@@ -26,8 +27,9 @@ describe('upload routes', () => {
   let req;
   let file;
 
-  const deleteAllFiles = cb => {
-    const directory = `${__dirname}/uploads/`;
+  const deleteAllFiles = async cb => {
```
RafaPolit, line 30: There are more calls to `deleteAllFiles` in this spec that are not being awaited.  They have a `done` inside that could cause the same effect, but maybe for consistency add the await on those calls? (currently lines 88 and 186)

app/api/files/filesystem.ts
```diff
@@ -47,12 +47,34 @@ async function deleteFiles(files: FilePath[]) {
...
+
+const generateUploadsPath = async (subPath: string) => {
+  if (subPath) {
+    await createDirIfNotExists(`${__dirname}/specs/uploads/${subPath}`);
+  }
+  return {
+    uploadedDocuments: `${__dirname}/specs/uploads/${subPath}`,
+    attachments: `${__dirname}/specs/uploads/${subPath}`,
+    customUploads: `${__dirname}/specs/uploads/${subPath}`,
+    temporalFiles: `${__dirname}/specs/uploads/${subPath}`,
```
RafaPolit, line 72: Same concern as above,! Maybe this solves it if we are using only one at a time in each test?

RafaPolit (commented):
I think these are great improvements.  Just a couple of comments / questions more than changes request.   I ran the suite several times, I had the `exportRoutes` timeout in one of the runs, but could not replicate it again, so it may have been an intermittent test or something just hanging up on my end.

app/api/files/filesystem.ts
```diff
@@ -47,12 +47,34 @@ async function deleteFiles(files: FilePath[]) {
 const testingUploadPaths = {
   uploadedDocuments: `${__dirname}/specs/uploads/`,
   attachments: `${__dirname}/specs/uploads/`,
-  customUploads: `${__dirname}/specs/customUploads/`,
+  customUploads: `${__dirname}/specs/uploads/`,
   temporalFiles: `${__dirname}/specs/uploads/`,
```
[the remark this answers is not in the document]
daneryl: @RafaPolit i have reverted the customUploads path, its now /specs/customUploads again, for the others i think the issue described still remains, we should create a tech debt.

app/api/files/specs/jsRoutes.spec.js
daneryl: Good catch, the 2 calls are now properly awaited

app/api/files/filesystem.ts
daneryl: this was not being used anymore actually, it has been removed now.

RafaPolit (approved):

RafaPolit:
LGTM

PR #3527, opened 2021-03-16
Added a migration to populate mimetype on attachments

grafitto:
fixes #3525 

PR checklist:
- [ ] Update READ.me ?
- [ ] Update API documentation ?

QA checklist:
- [ ] Smoke test the functionality described in the issue
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

PR #3558, opened 2021-03-23
Added a migration to populate mimetype on attachments (recreated)

### LINKED ISSUE #3525: Populate mimetype on attachments with url field
Add a migration that populates mimetype to attachments that have the url field.

grafitto:
fixes #3525 
similar to #3527

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/migrations/migrations/36-populate-mimetype-on-attachments/index.js
```diff
@@ -0,0 +1,31 @@
...
+    const cursor = await db.collection('files').find({});
+
+    // eslint-disable-next-line no-await-in-loop
+    while (await cursor.hasNext()) {
+      // eslint-disable-next-line no-await-in-loop
+      const file = await cursor.next();
+      if (file.url && !file.mimetype) {
+        // eslint-disable-next-line no-await-in-loop
+        const response = await request.head(file.url);
+        const mimetype = response.headers.get('content-type') || undefined;
```
fnocetti, line 23: I'm not sure what's the app behavior when there's no mime type. @konzz does it make sense to have a default type like application/octet-stream?

app/api/migrations/migrations/36-populate-mimetype-on-attachments/index.js
```diff
@@ -0,0 +1,31 @@
...
+        db.collection('files').updateOne({ _id: file._id }, { $set: { mimetype } });
+      } else if (file.filename && file.type === 'attachment' && !file.mimetype) {
+        const mimetype = mime.lookup(attachmentsPath(file.filename));
```
fnocetti, line 26: As far as I know `lookup` will return `false` if the type cannot be determined. We definitivelly need a default here like `.lookup(...) || undefined` (the default is to be defined, see previous comment)
cc. @konzz

fnocetti (changes requested):
I left one comment that I consider important. Apart from that, the migration works just fine!

app/api/migrations/migrations/36-populate-mimetype-on-attachments/index.js
fnocetti: Effectively, `false` it's been saved to the database in those cases.

app/api/migrations/migrations/36-populate-mimetype-on-attachments/index.js
konzz: Not sure if would be better to leave it empty if we don't know it. We only have a use case now if its image or video, but if in the future we use it for other things, not sure if a default would help or would harm.

fnocetti: Ok, let's leave it undefined :+1: thanks

app/api/migrations/migrations/36-populate-mimetype-on-attachments/index.js
fnocetti: @grafitto, ok, lets use undefined for the default, but still we need to implement it as default value as noted in the previous comment :) thanks!

grafitto: Done

fnocetti (approved):

PR #3559, opened 2021-03-23
3540 type aggregations based on read access to entities

### LINKED ISSUE #3540: Metadata aggregations counts are ignoring the permissions restrictions. (6pts)
When a collaborator user is presented with the library/uploads, the filters side panel is showing the counts as if permissions didn't exist. That is: if 2 entities exists for a particular aggregation and the user only has permissions to see 1, the aggregation count is still showing 2.
The fix for this issue implies, instead of manually pushing the permission filters to the query, setting the permissions filter by using the `addFilter` method in the query builder (because that method also works with the aggregations).

mfacar:
fixes #3540

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/search/documentQueryBuilder.js
+++ b/app/api/search/documentQueryBuilder.js
@@ -21,6 +21,20 @@ const nested = (filters, path) => ({
 });
 
 export default function() {
+  const defaultFilter = [
+    {
+      bool: {
+        should: [
+          {
+            term: {
+              published: true,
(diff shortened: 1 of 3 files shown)


daneryl:
LGTM!

PR #3562, opened 2021-03-24
search does not return permissions by default

### LINKED ISSUE #3541: Optimize permission results for searches. (13pts)
An optimization was already implemented in #3477 to not return the permissions information on mongo queries by default, preventing the application from iterating over the results to remove privileged information every time. After this optimization, calculations only occur when the permissions information is intentionally requested for a collaborator role, throw a field selection (see `select` in mongo/mongoose queries).
The same needs to be implemented for elasticsearch searches.

This task DO NOT includes adapting the client to leverage this functionality. Instead, for this time, the client is supposed to request permissions always. A follow up task might be created for the future since it's a really complex task.

daneryl:
fixes #3541

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Entities/EntitiesAPI.js
+++ b/app/react/Entities/EntitiesAPI.js
@@ -1,10 +1,15 @@
-/** @format */
-
 import api from 'app/utils/api';
+import { RequestParams } from 'app/utils/RequestParams';
 
 export default {
-  get(requestParams) {
-    return api.get('entities', requestParams).then(response => response.json.rows);
+  get(requestParams = new RequestParams()) {
+    const params = requestParams.add({
+      include:
+        requestParams.data && requestParams.data.include
+          ? requestParams.data.include.concat(['permissions'])
+          : ['permissions'],
+    });
+    return api.get('entities', params).then(response => response.json.rows);
   },
 
   countByTemplate(requestParams) {
--- a/app/react/Search/SearchAPI.js
+++ b/app/react/Search/SearchAPI.js
@@ -1,4 +1,5 @@
 import api from 'app/utils/api';
+import { RequestParams } from 'app/utils/RequestParams';
 
 export default {
   countByTemplate(requestParams) {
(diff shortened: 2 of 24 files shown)


app/react/Entities/EntitiesAPI.js
```diff
@@ -1,10 +1,15 @@
...
 
 export default {
-  get(requestParams) {
-    return api.get('entities', requestParams).then(response => response.json.rows);
+  get(requestParams = new RequestParams()) {
+    const params = requestParams.add({
+      include:
+        requestParams.data && requestParams.data.include
+          ? requestParams.data.include.concat(['permissions'])
+          : ['permissions'],
```
mfacar, line 10: @daneryl, more as a question than a change request, is not this the same that keeping burned "+permissions" into the route action? I was thinking the intention of adding the include parameter is to avoid getting the permissions property in several places at the frontend.

app/react/Search/SearchAPI.js
```diff
@@ -11,8 +12,14 @@ export default {
...
 
-  search(requestParams) {
-    return api.get('search', requestParams).then(response => response.json);
+  search(requestParams = new RequestParams()) {
+    const params = requestParams.add({
+      include:
+        requestParams.data && requestParams.data.include
+          ? requestParams.data.include.concat(['permissions'])
+          : ['permissions'],
+    });
```
mfacar, line 21: Here I have the same previous question

mfacar (commented):
@daneryl, LGTM, there are a lot of nice improvements around API use. I only have a question about the approach, as soon as you confirm me I'll be starting with the QA

app/react/Entities/EntitiesAPI.js
daneryl: @mfacar the intention was for the api not to return permissions by default, in our own stack looks the same, but now you have you explicitly ask for the permissions, does this answers the question ?

mfacar: Totally, thank you, I'll proceed with the QA

mfacar (approved):
LGTM!

PR #3572, opened 2021-03-30
getDocument returns relations.

### LINKED ISSUE #3499: Opening newly uploaded documents with relationships from the relationship tree crashes the document viewer (8pts)
1. On a blank Uwazi create an empty default "Entity A".
2. Upload a PDF document
3. Edit "Entity A"'s relationships and connect it to the document.
4. Save.
5. Click on the "View" button of the document in the relationship tree.
6. Uwazi tosses a JS error.

Uwazi is adding a "ref=xxxxx" param to the URL that should not be there.

daneryl:
fixes #3499 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Documents/components/DocumentSidePanel.js
+++ b/app/react/Documents/components/DocumentSidePanel.js
@@ -25,6 +25,7 @@ import { entityDefaultDocument } from 'shared/entityDefaultDocument';
 import SearchText from './SearchText';
 import ShowToc from './ShowToc';
 import SnippetsTab from './SnippetsTab';
+import helpers from '../helpers';
 
 export class DocumentSidePanel extends Component {
   constructor(props) {
@@ -271,7 +272,8 @@ export class DocumentSidePanel extends Component {
 
     const TocForm = this.props.tocFormComponent;
 
-    const { attachments, documents, language, defaultDoc } = doc.toJS();
+    const jsDoc = helpers.performantDocToJSWithoutRelations(doc);
+    const { attachments, documents, language, defaultDoc } = jsDoc;
 
     const isEntity = !documents || !documents.length;
     const defaultDocumentToC =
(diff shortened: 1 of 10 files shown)


RafaPolit:
I am wondering that polluting 6 files with `.delete('relations').toJS()` is the right approach.  I understand it is the solution, but if we later on want to actually use them, or find better solutions, finding this spread out within the code is not great.

Perhaps we can abstract that in a utility like `sanitizeDoc` somewhere that concentrates this `delete` in a single location?

Also, I'm not sure that the removed `omitRelationships` in `app/react/Viewer/actions/documentActions.js` is without impact for queries like a countries that could have thousands of relationships.

RafaPolit (changes requested):
I am wondering that polluting 6 files with `.delete('relations').toJS()` is the right approach.  I understand it is the solution, but if we later on want to actually use them, or find better solutions, finding this spread out within the code is not great.

Perhaps we can abstract that in a utility like `sanitizeDoc` somewhere that concentrates this `delete` in a single location?

Also, I'm not sure that the removed `omitRelationships` in `app/react/Viewer/actions/documentActions.js` is without impact for queries like a countries that could have thousands of relationships.

daneryl:
> I am wondering if polluting 6 files with `.delete('relations').toJS()` is the right approach. I understand it is the solution, but if we later on want to actually use them, or find better solutions, finding this spread out within the code is not great.
> 
> Perhaps we can abstract that in a utility like `sanitizeDoc` somewhere that concentrates this `delete` in a single location?

done :+1: 

> Also, I'm not sure that the removed `omitRelationships` in `app/react/Viewer/actions/documentActions.js` is without impact for queries like a countries that could have thousands of relationships.

i thought that the relationships returned were hard coded to 300 limit, this is not true, and only happens when asking for text relationships, hard coding a limit here can have unexpected ramifications.
I reverted back the `omitRelatonships` by default and instead opted for assigning on the `requestState` the same `textReferences` requested to the `doc.relations`, this way there is an exact match including not exceeding the 300 limit, let me know what you think @RafaPolit 

```diff --git a/app/react/Viewer/actions/routeActions.js b/app/react/Viewer/actions/routeActions.js
index a3ca9a9fe..ae3cc7c89 100644
--- a/app/react/Viewer/actions/routeActions.js
+++ b/app/react/Viewer/actions/routeActions.js
@@ -47,7 +47,10 @@ export async function requestViewerState(requestParams, globalResources) {
   return [
     setViewerState({
       documentViewer: {
-        doc,
+        doc: {
+          ...doc,
+          relations: references,
+        },
         references,
         relationTypes,
         rawText,


```

PR #3583, opened 2021-04-06
3570 remove publish buttons

### LINKED ISSUE #3570: Remove publish buttons and allow mixed access (8pts)
Follow up of: #3492 
Once publishing is allowed through the sharing functionality we need to:

- [x] Remove the publish/unpublish buttons
- [x] Allow mixed access permissions: when multiselecting entities and trying to assign permissions, the user is shown with "mixed access" for every user/group that has different access levels assigned to each multiselected entities. As for now, you need to fix that to be able to save (overwrite those permissions in order to be equal to the whole set). We need to change this so mixed access is valid and it keeps each entities the corresponding access levels for that member on save.
- [ ] Forbid entity publishing through POST/api/entity **(needs review)**
- [x] Include a notice in the bottom of the modal (currently used for validation errors) when the user is changing the publishing status (adding/removing the public permission). The logic for showing the validation errors is going to be removed in this issue. Now, it should be re-cycled to be used for this purpose.

### LINKED ISSUE #3573: Adding warning on sharing with public
We need to add additional warning when sharing publicly.

public

([Link to design](https://www.figma.com/file/2AT2ilGWBTe0pWy4j66yxW/Uwazi-Access-Levels?node-id=835%3A939))

fnocetti:
fixes #3570 
fixes #3573 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/entities/entities.js
+++ b/app/api/entities/entities.js
@@ -134,6 +134,9 @@ async function updateEntity(entity, _template) {
           await this.renameRelatedEntityInMetadata({ ...currentDoc, ...entity });
         }
         const toSave = { ...entity };
+
+        toSave.published = d.published;
+
         if (entity.metadata) {
           toSave.metadata = await denormalizeMetadata(entity.metadata, entity, template);
         }
@@ -161,10 +164,6 @@ async function updateEntity(entity, _template) {
           d.suggestedMetadata[p] = entity.suggestedMetadata[p] || [];
         });
         d.suggestedMetadata = await denormalizeMetadata(d.suggestedMetadata, d, template);
-      }
-
-      if (typeof entity.published !== 'undefined') {
-        d.published = entity.published;
       }
 
       if (typeof entity.template !== 'undefined') {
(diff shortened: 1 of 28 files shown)


daneryl (changes requested):
@fnocetti LGTM, found a couple things only.
- publish button still appears when selecting multiple entities.
- not related to this specific issue, maybe i reported this already on any other access-levels branch ? do not remember: when selecting some value on the permissions modal and deleting it, this value can not be selected again.
out

fnocetti:
> * publish button still appears when selecting multiple entities.
Good catch. I must have forgotten. I'll work on it. Thanks

> * not related to this specific issue, maybe i reported this already on any other access-levels branch ? do not remember: when selecting some value on the permissions modal and deleting it, this value can not be selected again.
As far as I saw in the past, you was able to search by typing and get the member again, so I considered it low priority. Maybe that's not true. I'll report an issue. Thanks!

daneryl (approved):

PR #3601, opened 2021-04-15
3547 search v2

### LINKED ISSUE #3547: Search Slice Initial Refactor
As a team, we want to refactor the search slice, so we can improve our velocity while developing in this area and improve the performance of any search done in Uwazi.

As initial phase we are going to develop a new endpoint for search, this endpoint will be used by the functionality of search entities to create relationships / references. 

- Develop a new endpoint that accepts a search by title
- Develop a new Search / QueryBuilder that will be used by this new endpoint
- Update the react action to search entities by title to use the new endpoint (create relationships, copy from... etc)

note: Minimize impact on client, the main objective is to refactor the backend slice.

daneryl:
fixes #3547 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Connections/actions/actions.js
```diff
@@ -1,22 +1,26 @@
+import qs from 'qs';
```
RafaPolit, line 1: I think we already had querystring as a depencency? Is this the same thing?

mfacar (changes requested):
LGTM! Great!, pretty nice and clean implementation. I left some minor comments. 
I tested the functionality from References, but the Connections screen and Copy from are still using the old search endpoint (they are mentioned in the issue)
Regarding the functionality all works fine, just a case when there's an error in elastic search, I forced an error sending a very long searchString, the alert shows an empty box, it seems the error is not been passing to the client, maybe in this case in particular, it could be useful to add a max length for the param queryString?

app/react/Connections/actions/actions.js
```diff
@@ -1,22 +1,26 @@
...
 import * as types from './actionTypes';
 import * as uiActions from './uiActions';
 
-export function immidiateSearch(dispatch, searchTerm, connectionType) {
+export function immidiateSearch(dispatch, searchString, connectionType) {
   dispatch(uiActions.searching());
 
-  const requestParams = new RequestParams({ searchTerm, fields: ['title'] });
+  const requestParams = new RequestParams(
+    qs.stringify({
```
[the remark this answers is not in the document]
daneryl: rison is still there, on the browser application only, qs is to stringify query objects to the new endpoint.

app/api/search.v2/routes.ts
```diff
@@ -0,0 +1,63 @@
+import { Application, Request, Response, NextFunction } from 'express';
+//@ts-ignore
+import boolParser from 'express-query-boolean';
```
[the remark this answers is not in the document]
daneryl: let me check.

app/react/Connections/actions/actions.js
```diff
@@ -1,22 +1,26 @@
...
 import api from 'app/utils/api';
 import debounce from 'app/utils/debounce';
 import { RequestParams } from 'app/utils/RequestParams';
-import SearchApi from 'app/Search/SearchAPI';
 
 import * as types from './actionTypes';
 import * as uiActions from './uiActions';
 
-export function immidiateSearch(dispatch, searchTerm, connectionType) {
+export function immidiateSearch(dispatch, searchString, connectionType) {
```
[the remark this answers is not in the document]
daneryl: done

app/api/search.v2/routes.ts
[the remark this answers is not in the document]
daneryl: since im removing query-string in favor of qs im going remove boolParser and use the recommended way from qs https://github.com/ljharb/qs#parsing-primitivescalar-values-numbers-booleans-null-etc, it also relies on a library that seems not maintained, the  source code is really small, if we have any problem with this we can take care of this ourselves.

app/react/Connections/actions/actions.js
daneryl: @RafaPolit [not the same thing](https://stackoverflow.com/questions/29175465/body-parser-extended-option-qs-vs-querystring/29177740#:~:text=The%20main%20advantage%20of%20qs,any%20json%2Dlike%20data%20structure.&text=querystring%20library%60%20provides%20basic%20serialization,browsers%20to%20serialize%20form%20data.) i have used qs mainly because of its ability to parse nested objects.
i have also removed query-string as a dependency and changed the previous code to use qs instead.

daneryl:
> I tested the functionality from References, but the Connections screen and Copy from are still using the old search endpoint (they are mentioned in the issue)

this new endpoint should only be used on the sidepanel when creating a connection, not in the connections screen main search.

> Regarding the functionality all works fine, just a case when there's an error in elastic search, I forced an error sending a very long searchString, the alert shows an empty box, it seems the error is not been passing to the client, maybe in this case in particular, it could be useful to add a max length for the param queryString?

aparently the blank error is caused by https://github.com/huridocs/uwazi/blob/7dade2f8bfc3ae88a7d53f01f5d1a8ec1aa33e5e/app/shared/JSONRequest.js#L98
we are returning a blank error in this scenario. the error for the case you describe is 
```
SyntaxError: Unexpected end of JSON input
```
which is caused by a 431 http status Request Header Fields Too Large, should we create an issue to discuss how we should treat this errors ? @RafaPolit @konzz @mfacar

mfacar:
> which is caused by a 431 http status Request Header Fields Too Large, should we create an issue to discuss how we should treat this errors ?

Thank you for the detailed analysis @daneryl, in my opinion, I think it merits adding an issue to handle this scenario maybe adding a custom error message at `handleErrorStatus` and probably at the front end adding a max for input length

mfacar (approved):

app/api/search.v2/routes.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+};
+
+const searchRoutes = (app: Application) => {
+  app.get(
+    '/api/v2/entities',
+    queryTypes.middleware(),
+    validateRequest({
+      query: SearchQuerySchema,
+    }),
```
daneryl, line 42: this validation has no tests, and i think is wrong it should be something like
```
{
  properties: {
      query: SearchQuerySchema,
    }
}

File: app/api/search.v2/routes.ts [hunk 6]
```diff
@@ -0,0 +1,63 @@
...
+    next(e);
+  }
+};
+
+const searchRoutes = (app: Application) => {
+  app.get(
+    '/api/v2/entities',
+    queryTypes.middleware(),
+    validateRequest({
+      query: SearchQuerySchema,
```
konzz, line 41: this validation has no tests, and i think is wrong it should be something like

konzz (changes requested):

app/api/search.v2/routes.ts
daneryl: added test and fixed

PR #3602, opened 2021-04-16
denormalize inherited

### LINKED ISSUE #3497: Denormalize inherited values instead of “UI automagic” (20pts)
Right now the UI uses the relationships of an entity to show inherited values. Because of that, the inherited values are only properly shown in the side panel. We want the data to be denormalised, store in the database, indexed and shown correctly in all the app.

- The [denormalise](https://github.com/huridocs/uwazi/blob/37d91765babc2e6b685a584d39f10e578777c2e7/app/api/entities/entities.js#L22) function should get the inherited values and store them in a `.inherited` key in the property.

When denormalizing you are going to find:
 1. A concrete value (ie. a date, number or text), the value gets denormalized
 2.  A thesauri key, both the key and the value needs to get denormalized (the value for representation, the key for aggregations), although maybe these aggregations can be based on the values, but still you need a way to cascade update when there are changes in the inherited values.
 3.  An inherited property (that is, trying to inherit an already inherited property). We want to avoid this scenario for now. This can lead to circular references or combinations that cycle through massive datasets.

- The [formater](https://github.com/huridocs/uwazi/blob/development/app/react/Metadata/helpers/formater.js) should know how to handle properties with inherited values, and remove any code related to the previous implementation.

- Update the metadata property [schema](https://github.com/huridocs/uwazi/blob/37d91765babc2e6b685a584d39f10e578777c2e7/app/shared/types/commonSchemas.ts#L93)
- Create a migration to denormalize the existing data
- Remove any code related to the previous implementation
- Review that when changing the template, the metadata inherited should be properly updated

konzz:
fixes #3497

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


daneryl:
@konzz there is a couple issues.
- this development needs a migration for the already existing datasets
- selecting an inherited property that is not set on the target entity causes the app to crash. (im not sure this can happen on new data but its happening to me using fixtures and selecting as inherited a Mecanismo -> resumen)
- editing the original property does not updated inherited values on the entities that inherit the property.

daneryl (changes requested):
@konzz there is a couple issues.
- this development needs a migration for the already existing datasets
- selecting an inherited property that is not set on the target entity causes the app to crash. (im not sure this can happen on new data but its happening to me using fixtures and selecting as inherited a Mecanismo -> resumen)
- editing the original property does not updated inherited values on the entities that inherit the property.

daneryl (changes requested):
migration fails at least in one dataset we are hosting.

daneryl (approved):

PR #3617, opened 2021-04-20
3598 - Custom page for entity view

### LINKED ISSUE #3598: Custom page for entity view
As users we want to define a custom view (page) for each template so our data is represented in a meaning-full way.

Tasks:

- Add a markdown editor to the template edition page.
- Add a css box
- Add some text help so the users know how to put different values
- Update backend templates model / validation 
- Hook the entity view to use the template page when defined
- No custom JS (for now)

Tips:
- We already have a [Value](https://github.com/huridocs/uwazi/blob/development/app/react/Markdown/components/Value.js) component for pages:
```
<PagesContext.Provider value={{ title: 'Bruce Wayne' }}>
   <ValueComponent path="title" />
</PagesContext.Provider>
```

- Some values like dates will require a pre-formating, so makes sense to use the formater to prepare the data.

Zasa-san:
fixes #3598

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

For MVP we handle editing/deleting pages used for entity view the following way:
If a user wants to stop using an existing page as entity view, and they try to toggle off the functionality on the page editor or delete the page, they will get an error message explaining that the page is in use with a list of the templates using them. This way they can remove the page from the templates using it, then proceed to toggle entity view off or delete it.


Zasa-san:
For MVP we handle editing/deleting pages used for entity view the following way:
If a user wants to stop using an existing page as entity view, and they try to toggle off the functionality on the page editor or delete the page, they will get an error message explaining that the page is in use with a list of the templates using them. This way they can remove the page from the templates using it, then proceed to toggle entity view off or delete it.

app/react/Pages/components/PageViewer.js
```diff
@@ -12,12 +12,14 @@ import Script from './Script';
...
-export class PageViewer extends Component {
+class PageViewer extends Component {
   render() {
     const { page, itemLists, datasets } = this.props;
     const lists = itemLists.toJS();
     const originalText = page.getIn(['metadata', 'content']) || '';
-    const scriptCode = page.getIn(['metadata', 'script']) || '';
+    let scriptCode = page.getIn(['metadata', 'script']) || '';
+
+    scriptCode = `var datasets=${JSON.stringify(datasets)};\r\n${scriptCode}`;
```
daneryl, line 22: was this really part of the issue ? or absolutely needed ? i think we should discuss this first ? we are now coupling markdown datasets with the script that was thought to be self contained ?

app/react/Pages/utils/specs/setPageAssets.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+          { customDataset: { url: 'someurl' } },
+          { customData: { localData: 'from store' } }
+        );
+
+        expect(getDatasetsActions(stateActions)).toEqual({
+          request1: 'url1',
+          request2: 'url2',
+          customDataset: { url: 'someurl' },
+          customData: { localData: 'from store' },
+        });
```
daneryl, line 135: this is easier to test if the setPagesAssets only returns the values and not the actions.

app/react/Templates/components/MetadataTemplate.tsx
```diff
@@ -131,42 +138,50 @@ export class MetadataTemplate extends Component<MetadataTemplateProps> {
...
-                <div className="no-properties">
-                  <span className="no-properties-wrap">
-                    <Icon icon="clone" />
-                    Drag properties here
-                  </span>
-                </div>
-              </ul>
-            )}
-          </ShowIf>
+          {environment === 'template' && (
```
daneryl, line 141: is this to use this form also for relationType form ?
i think there is better ways to accomplish this reusability, like Component composition.\
Example:
FormWithTitle is the base form, accepts children that MetadataForm can pass to create the extended form
relationType just uses the baseForm or an extension to give semantics.

app/react/Templates/components/ViewTemplateAsPage.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { useEffect } from 'react';
```
daneryl, line 1: ViewTemplateAsPage reads like this is a view of some sort, can we provide a better name that hints that this is a Control of some sort ?

app/shared/types/pageSchema.ts
```diff
@@ -0,0 +1,74 @@
...
+              ', '
+            )}`,
+            dataPath: '.pages',
+          },
+        ]);
+      }
+    }
+    return true;
+  },
+});
```
daneryl, line 40: i know there is already some schema validators that include async data consistency validations, but this one is new, can we leave it as only schema validator ? it makes sense to have this 2 concepts separated ? schema validator is agnostic and can be used in other contexts, like in the browser application.

daneryl (changes requested):
still missing the functional QA, in the meantime there are some comments in the code.

app/api/pages/pages.ts
```diff
@@ -0,0 +1,105 @@
...
+      const templatesTitles = templatesUsingPage.map(template => template.name);
+      return Promise.reject(
+        createError(
+          `This page is in use by the following templates: ${templatesTitles.join(
+            ', '
+          )}. Remove the page from the templates before trying again.`,
+          409
+        )
+      );
+    }
```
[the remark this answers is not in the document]
RafaPolit: Interesting! This is a `delete` operation.  How/when would a validator kick in during the deletion process?

app/api/pages/specs/routes.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    await testingDB.disconnect();
+  });
+
+  describe('/api/pages', () => {
+    it('should validate with minimum required props', async () => {
+      const goodData = { title: 'good structure' };
+
+      const response = await request(app)
+        .post('/api/pages')
+        .set('X-Requested-With', 'XMLHttpRequest')
```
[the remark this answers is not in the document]
RafaPolit: Yes, let's discuss this during a standup to see what would be the right approach?

app/react/Entities/actions/specs/uiActions.spec.js
```diff
@@ -6,4 +6,10 @@ describe('Entities uiActions', () => {
       expect(uiActions.showTab('tab')).toEqual({ type: 'SHOW_TAB', tab: 'tab' });
     });
   });
+
+  describe('resetUserSelectedTab', () => {
+    it('should broadcast RESET_USER_SELECTED_TAB', () => {
+      expect(uiActions.resetUserSelectedTab()).toEqual({ type: 'RESET_USER_SELECTED_TAB' });
+    });
+  });
```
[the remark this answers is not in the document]
RafaPolit: Yes, mostly useless. Either tested by integration or abstracted.  Agree.

app/react/Pages/components/PageViewer.js
RafaPolit: This was requested by @txau precisely as a intermediate step into "dynamic" access to properties.  I actually think it is monstrously useful and really don't know why we didn't do this sooner.  

If you don't think it a good feature, yes, lets please discuss this.

app/react/Pages/utils/specs/setPageAssets.spec.ts
RafaPolit: Maybe, but then we would still need to abstract the setting of the actions so they can be called from two different routes, so the test of this function will be simpler, but the test of the "parent" function will still look like this.  Do you think this merits the change still?

app/react/Templates/components/MetadataTemplate.tsx
RafaPolit: Yes, this is probably true.  This was the least intervention we found (as it was already pseudo implemented with some ShowIfs.... we have already touched a lot of code not belonging to this issue.  Do you think this refactor needs to happen now?

app/react/Templates/components/ViewTemplateAsPage.tsx
RafaPolit: Yes.  Good idea.

app/shared/types/pageSchema.ts
RafaPolit: I am confused.  We have both separated: the file exports the agnostic schema validator, but it also exports the function to validate the page.  Is this not what you refer to?

app/react/Pages/PageView.js
```diff
@@ -1,69 +1,18 @@
...
-      items: listSearchs[index].rows,
-      options: searchOptions[index],
-    }));
-
-    return [
-      actions.set('page/pageView', pageView),
-      actions.set('page/itemLists', itemLists),
-      actions.set('page/datasets', datasets),
-    ];
+    return setPageAssets(requestParams);
```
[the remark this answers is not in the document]
RafaPolit: Yes, this is probably true.  It is called set because it also executes the actions to set the redux store.  Hmmmm... not sure.

app/api/pages/pages.ts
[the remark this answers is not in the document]
daneryl: i did no notice this was a delete, forget about it for now, but still maybe worth extracting to some kind of validator at some point, this and other deletion validations.

app/react/Templates/components/MetadataTemplate.tsx
daneryl: if you did not needed to touch this, i prefer to be like it was than making a refactor that does not make the code better.

app/shared/types/pageSchema.ts
daneryl: ok, understood this works for now, but the validator exported includes async validations, you are just exporting the schema as a separated export, to be used by the "route validator" right ?

app/react/Pages/utils/specs/setPageAssets.spec.ts
daneryl: the initial reason was for better concern separation and readability of other functions, so yes i think it merits.

app/react/Pages/components/PageViewer.js
daneryl: i understand the utility, but some of this datasets are huge, and we already have them loaded like this on some scenarios, for instance when you do SSR they are on ```window.__reduxData__.page.datasets``` we will duplicate what already is really big in this scenario.

maybe an alternative is to give the script access to the store itself, that has all that data loaded in memory already which give the script access not only to datasets, but also to everything else ? i guess its potentially a security issue ? but come on, this free JS script its already that.

app/shared/types/pageSchema.ts
RafaPolit: Yes, and client type definitions.

app/react/Pages/components/PageViewer.js
RafaPolit: Moving this development to a new issue.

RafaPolit:
We found a problem:
- page's `entityView` is not being synced across multiple languages, and you can end up with a page selected as `entityView` in one language and not in another.

We need to fix this before moving forward.  DO NOT MERGE!

Zasa-san:
> We found a problem:
> 
>     * page's `entityView` is not being synced across multiple languages, and you can end up with a page selected as `entityView` in one language and not in another.
> 
> 
> We need to fix this before moving forward. DO NOT MERGE!

This should be addressed now after the last push.

daneryl (changes requested):

daneryl (commented):
https://github.com/huridocs/uwazi/issues/3598#issuecomment-823359958

> prevent those pages from being assigned to menus

- this is not being taken into account, i can even assign one of the existing pages to a template, it works well, but it does not make sense once one starts writing specific entity data on the page.

- when the entity has a document there is no icon to see the page.

- im not able to show entity information using the new page, can you provide an example ?

- we think that the default context/dataset for this pages should be the entity itself for ease of access, instead of the "default" dataset (also properly formatted), thoughts ?
   - ValueComponent can be used to access raw data specifying the dataset you want (this currently does not work because of bad ValueComponent design ?)
   - ValueComponent should have as a default context the formated current entity for ease of access, (this also removes the need of the new CurrentEntity component) we still believe that the orignal issue tip is a better solution.

app/api/pages/specs/pages.spec.ts
```diff
@@ -80,6 +80,20 @@ describe('pages', () => {
...
+        const updatedPages = await pages.get({ sharedId: '3' });
+        expect(updatedPages.length).toBe(2);
+        expect(updatedPages).toEqual(
+          expect.arrayContaining([
+            expect.objectContaining({
+              entityView: true,
+            }),
+          ])
+        );
+      });
```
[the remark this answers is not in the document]
Zasa-san: It's certainly more readable, changed.

RafaPolit:
Proposed wiki:
https://github.com/huridocs/uwazi-documentation/blob/eef2d5010afa77252c02c8bab38946147ec75c74/admin-docs/creating-entity-pages.md

PR #3619, opened 2021-04-20
Share button only renders if theres an entity selected

### LINKED ISSUE #3618: Adding/removing Public permission causes an error.
The error is in the client and doesn't cause an inconsistent state.
It is due to the entity being removed from the store and not being unselected, leading to a selected `null`.
The problem is that the `unselect documents` action is wrongly dispatched.

fnocetti:
fixes #3618 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


mfacar (approved):
LGTM!, I'm taking advance of this PR to merge with the target branch feature/access_levels and verify the integration made with development

PR #3621, opened 2021-04-21
Consider public type in activity log

### LINKED ISSUE #3615: Activity log shows an error when there were permissions modified with public entities (3pts)
When there are entities whose permissions were modified with public as an option, the activity log fails at parsing the request body.

Screen Shot 2021-04-19 at 16 23 35

fnocetti:
fixes #3615

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/activitylog/helpers.js
+++ b/app/api/activitylog/helpers.js
@@ -129,5 +129,5 @@ export const loadAllowedUsersAndGroups = data => {
 
   return ` with permissions for${grantedUsers.length ? ` USERS: ${grantedUsers};` : ''}${
     grantedNames.length ? ` GROUPS: ${grantedNames}` : ''
-  }${data.public ? '; and PUBLIC' : ''}`;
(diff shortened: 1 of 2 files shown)


daneryl (changes requested):
@fnocetti all good, there is a minor issue with the message generated. if you only add the public group to an entity or multiple entities, the activity log message reads as follows:

```with permissions for; and PUBLIC```

fnocetti:
> @fnocetti all good, there is a minor issue with the message generated. if you only add the public group to an entity or multiple entities, the activity log message reads as follows:
> 
> `with permissions for; and PUBLIC`

Good one. I'll fix it.

fnocetti:
@daneryl so I basically removed the "and" since there were none before this change. Please review.

daneryl (approved):

PR #3624, opened 2021-04-23
3606 disallow permissions setting entities api

### LINKED ISSUE #3606: Setting the permissions field through api/entities produces inconsistent state and shouldn't be allowed (3 pts)
The permissions field management is internal to the permissions module.
Setting it using the entities api should be forbidden as it results in an inconsistent state (different permissions per lang) and is a potential security issue.

fnocetti:
fixes #3606 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/entities/entities.js
+++ b/app/api/entities/entities.js
@@ -107,7 +107,7 @@ const FIELD_TYPES_TO_SYNC = [
 ];
 
 async function updateEntity(entity, _template, unrestricted = false) {
-  const docLanguages = await this.getAllLanguages(entity.sharedId, '+permissions');
+  const docLanguages = await this.getAllLanguages(entity.sharedId);
   if (
     docLanguages[0].template &&
     entity.template &&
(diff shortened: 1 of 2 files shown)


app/api/entities/entities.js
```diff
@@ -137,6 +137,7 @@ async function updateEntity(entity, _template, unrestricted = false) {
         const toSave = { ...entity };
 
         toSave.published = d.published;
+        toSave.permissions = d.permissions;
```
daneryl, line 140: @fnocetti i think there may be a better way to ignore the permissions, unsetting them or setting them to undefined should work, without having to request the originals, same for published? or is this done this way in order to be able to set a default ?

daneryl (changes requested):

app/api/entities/entities.js
fnocetti: No, the default was a precondition for this to work. I'm not sure at this point, I'll give it a try. I remember failing to do it with `published`. But I might be wrong.
I'll let you know how it goes 👍

daneryl (approved):

PR #3625, opened 2021-04-26
Added async validation to the templates to check the thesauri exists

### LINKED ISSUE #1003: Selecting a deleted thesauri on a template and saving it, should return a server error saying the thesauri no longer exists (5 pts)
- Right now it can happen that you select a thesauri that has been just deleted and the server saves it no problem, comming back to the template results in an error when trying to edit the property with the unexistent thesauri.
- Make sure the async validation happens on the "model" and not in the route.

konzz:
fixes #1003 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/templates/specs/fixtures.js
+++ b/app/api/templates/specs/fixtures.js
@@ -14,6 +14,7 @@ const swapTemplate = db.id();
 const relatedTo = db.id();
 const templateToBeInherited = db.id();
 const propertyToBeInherited = db.id();
+const thesauriId1 = db.id();
 export default {
   templates: [
     {
@@ -130,6 +131,7 @@ export default {
       languages: [{ key: 'en', label: 'English', default: true }],
     },
   ],
+  dictionaries: [{ _id: thesauriId1, name: 'options' }],
 };
 
 export {
@@ -139,4 +141,5 @@ export {
   swapTemplate,
   templateToBeInherited,
   propertyToBeInherited,
+  thesauriId1,
 };
--- a/app/api/templates/specs/fixtures.js
+++ b/app/api/templates/specs/fixtures.js
@@ -1,9 +1,3 @@
-/**
- * /* eslint-disable
- *
- * @format
- */
-
 import db from 'api/utils/testing_db';
 import { propertyTypes } from 'shared/propertyTypes';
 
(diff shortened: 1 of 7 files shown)


app/api/templates/specs/fixtures.js
```diff
@@ -14,6 +14,7 @@ const swapTemplate = db.id();
 const relatedTo = db.id();
 const templateToBeInherited = db.id();
 const propertyToBeInherited = db.id();
```
mfacar, line 16: Please, can you remove the first six lines in this file given there aren't needed already?

app/shared/types/templateSchema.ts
```diff
@@ -75,14 +76,20 @@ ajv.addKeyword('uniquePropertyFields', {
...
       return true;
     }
     if (['multiselect', 'select'].includes(data.type)) {
-      return !!(data.content && data.content.length);
+      if (!data.content || !data.content.length) {
+        return false;
+      }
+
+      const found = await thesauri.getById(data.content);
+      return !!found;
```
mfacar, line 92: Given the description, in the issue, this validation should throw a specific message "the thesaurus no longer exists", is it possible to change the definition of the keyword to return a ValidationError with a custom message?

mfacar (changes requested):
LGTM!
I just have a doubt respect the message, please review the comments.

app/api/templates/specs/fixtures.js
konzz: done

app/shared/types/templateSchema.ts
konzz: This error is quite rare, I think adding a different error message for this scenario only adds complexity and no real benefit, but thats just my opinion, so I will comment this today in the standup and see if we get an agreement

mfacar: For the record, we agreed to change the keyword to requireOrInvalidContentForSelectFields, having "validation failed: .properties[0] should pass "requireOrInvalidContentForSelectFields" keyword validation" as message

PR #3631, opened 2021-04-27
3630 public form with access levels

### LINKED ISSUE #3630: Creating new entities through a submission form with a logged-out user breaks the app
Testing a public form as a logged-out user threw an error then the instance went down for maintenance, it came back up after forcing a refresh.

mfacar:
fixes #3630

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/entities/specs/entities.spec.js
+++ b/app/api/entities/specs/entities.spec.js
@@ -536,7 +536,7 @@ describe('entities', () => {
       });
     });
 
-    describe('save from public form', () => {
+    describe('save entity without a logged user', () => {
       it('should save the entity with unrestricted access', async () => {
         const user = {};
         userFactory.mock(undefined);
--- a/app/api/entities/specs/entities.spec.js
+++ b/app/api/entities/specs/entities.spec.js
@@ -535,6 +535,18 @@ describe('entities', () => {
           .catch(catchErrors(done));
       });
     });
+
+    describe('save from public form', () => {
+      it('should save the entity with unrestricted access', async () => {
+        const user = {};
+        userFactory.mock(undefined);
+        const entity = { title: 'Batman begins', template: templateId, language: 'es' };
+        const createdEntity = await entities.save(entity, { user, language: 'es' });
+        expect(createdEntity._id).not.toBeUndefined();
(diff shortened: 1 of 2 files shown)


app/api/entities/specs/entities.spec.js
```diff
@@ -535,6 +535,18 @@ describe('entities', () => {
...
+      it('should save the entity with unrestricted access', async () => {
+        const user = {};
+        userFactory.mock(undefined);
+        const entity = { title: 'Batman begins', template: templateId, language: 'es' };
+        const createdEntity = await entities.save(entity, { user, language: 'es' });
+        expect(createdEntity._id).not.toBeUndefined();
+        expect(createdEntity.title).toEqual(entity.title);
+        userFactory.mockEditorUser();
+      });
+    });
```
daneryl, line 549: @mfacar what makes this test a  "save from public form" ? is it the fact that there is not user logged in ?

daneryl (commented):
@mfacar i do not understand what is the cause of loosing the session and what is the fix, can you briefly explain ?

app/api/entities/specs/entities.spec.js
mfacar: yes @daneryl, I'm not sure if there is another place where the save method is called with an empty user, but I think there is no other way to distinguish if it comes from a public form. Maybe the name should be more descriptive in order to avoid any confusion?

mfacar:
> @mfacar i do not understand what is the cause of loosing the session and what is the fix, can you briefly explain ?

There is not a loss of session, basically, the error is when an entity is saved from the public endpoint without a user logged. It should be allowed, however, obviously, it cannot be a trace of any user at permissions, thus the rest of the implementation of saving fails.

daneryl:
> There is not a loss of session, basically, the error is when an entity is saved from the public endpoint without a user logged. It should be allowed, however, obviously, it cannot be a trace of any user at permissions, thus the rest of the implementation of saving fails.

@mfacar understood, i mixed up issues !, looks good !

PR #3635, opened 2021-04-28
Allow hidding label in sidebar and entity view

### LINKED ISSUE #3616: Hide label option should also work in entity view and side panel (3 pts)
As users we want that the hide label option also works in the side panel and entity view, because it better represents our data.

The hide label option in the template property configuration, only hides the label for properties in cards, it should also hide the label when representing the metadata in the entity view and the side panel.

fnocetti:
fixes #3616 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Metadata/components/specs/__snapshots__/ShowMetadata.spec.js.snap
+++ b/app/react/Metadata/components/specs/__snapshots__/ShowMetadata.spec.js.snap
@@ -79,6 +79,7 @@ exports[`Metadata should render without timeline by default 1`] = `
   <Connect(BaseFormatMetadata)
     entity={Object {}}
     groupGeolocations={false}
+    renderLabel={[Function]}
   />
 </div>
 `;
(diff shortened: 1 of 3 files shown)


app/react/Metadata/components/specs/__snapshots__/ShowMetadata.spec.js.snap
```diff
@@ -79,6 +79,7 @@ exports[`Metadata should render without timeline by default 1`] = `
   <Connect(BaseFormatMetadata)
     entity={Object {}}
     groupGeolocations={false}
+    renderLabel={[Function]}
   />
 </div>
 `;
```
daneryl, line 85: can you please replace this full snapshot with a specific test for the timeline not being rendered? i know its not from the issue.

daneryl (changes requested):
@fnocetti all looks good, there is just a small request on the code :)

app/react/Metadata/components/specs/__snapshots__/ShowMetadata.spec.js.snap
fnocetti: Yeah, makes sense 👍

daneryl (approved):

PR #3639, opened 2021-04-30
Public form e2e

konzz:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/auth/routes.js
+++ b/app/api/auth/routes.js
@@ -82,7 +82,8 @@ export default app => {
 
   app.get('/api/captcha', cors(corsOptions), async (_req, res) => {
     const captcha = svgCaptcha.create({ ignoreChars: '0OoiILluvUV' });
-    const storedCaptcha = await CaptchaModel.save({ text: captcha.text });
+    const text = process.env.DATABASE_NAME !== 'uwazi_e2e' ? captcha.text : '42hf';
+    const storedCaptcha = await CaptchaModel.save({ text });
 
     res.json({ svg: captcha.data, id: storedCaptcha._id.toString() });
   });
--- a/e2e/suites/public-form.test.ts
+++ b/e2e/suites/public-form.test.ts
@@ -3,13 +3,6 @@ import proxyMock from '../helpers/proxyMock';
 import insertFixtures from '../helpers/insertFixtures';
 import disableTransitions from '../helpers/disableTransitions';
 
-const selectors = {
-  pageContentsInput:
-    '.page-viewer.document-viewer > div > div.tab-content.tab-content-visible > textarea',
-  useCustomLandingPage:
-    '#collectionSettings > div:nth-child(5) > div > div.toggle-children-button > label',
-};
-
 describe('Custom home page and styles', () => {
   beforeAll(async () => {
     await insertFixtures();
@@ -38,7 +31,7 @@ describe('Custom home page and styles', () => {
     await expect(page).toClick('a', { text: 'Add page' });
     await expect(page).toFill('input[name="page.data.title"]', 'Public Form Page');
     await expect(page).toFill(
-      selectors.pageContentsInput,
+      '.markdownEditor textarea',
       '<PublicForm template="58ada34c299e82674854504b" />'
     );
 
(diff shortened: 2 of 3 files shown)


app/api/auth/routes.js
```diff
@@ -82,7 +82,8 @@ export default app => {
 
   app.get('/api/captcha', cors(corsOptions), async (_req, res) => {
     const captcha = svgCaptcha.create({ ignoreChars: '0OoiILluvUV' });
-    const storedCaptcha = await CaptchaModel.save({ text: captcha.text });
+    const text = process.env.DATABASE_NAME !== 'uwazi_e2e' ? captcha.text : '42hf';
```
daneryl, line 85: we sure we want to depend on the database_name ? i know the chances are small, but it does not feel right, are we using the same condition for the mail "mock" ?

e2e/suites/public-form.test.ts
```diff
@@ -0,0 +1,81 @@
+import { adminLogin, logout } from '../helpers/login';
+import proxyMock from '../helpers/proxyMock';
+import insertFixtures from '../helpers/insertFixtures';
+import disableTransitions from '../helpers/disableTransitions';
+
+const selectors = {
+  pageContentsInput:
+    '.page-viewer.document-viewer > div > div.tab-content.tab-content-visible > textarea',
```
daneryl, line 8: i think we are abandoning this way of using selectors ? any specific reason to use the full path?, i believe there is only 1 or 2 textarea on the page ?

e2e/suites/public-form.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  useCustomLandingPage:
```
daneryl, line 9: this selector seems not used

e2e/suites/public-form.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    );
+
+    await expect(page).toClick('button', { text: 'Save' });
+  });
+
+  it('should visit the page and do a submit', async () => {
+    const [newTab] = await Promise.all([
+      new Promise(resolve => browser.once('targetcreated', target => resolve(target.page()))),
+      expect(page).toClick('a', { text: 'view page' }),
+    ]);
```
daneryl, line 52: this is difficult to understand, what does this do ? abstract into a function to give it meaning.

e2e/suites/public-form.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    );
+    await expect(newTab).toClick('span', { text: 'Bahamas' });
+    await expect(newTab).toFill('.captcha input', '42hf');
+    await expect(newTab).toClick('input[type="submit"]');
+    await expect(newTab).toClick('.alert.alert-success');
+  });
+
+  it('should check the newly created entity', async () => {
+    await page.bringToFront();
+    await page.waitFor(500); // index delay
```
daneryl, line 67: @fnocetti did you finally implement a function to force the elasticsearch refresh ?

daneryl (changes requested):

app/api/auth/routes.js
konzz: yes, its the same condition than in the mail mock, do not like it either

e2e/suites/public-form.test.ts
konzz: Updated it to '.markdownEditor textarea'  thats the best I can do without modifying the Markdown component.
I think all components we write should accept className.

e2e/suites/public-form.test.ts
konzz: removed

e2e/suites/public-form.test.ts
konzz: extracted to a different method

e2e/suites/public-form.test.ts
konzz: added @fnocetti  elastic refresh

app/api/auth/routes.js
daneryl: lets go with it and create an tech debt to change both ?

daneryl (approved):

Issue #3600, opened 2021-04-15
Diagnose performance problems on entity save (4 days) (5pts)


konzz:
Some users have experienced that saving an entity takes exponential time based on the amount of relationships it has.
Some processes that save multiple entities (like updating a template) result in an out of memory error for some instances.

We dont know the problem nor the solution so we want to spend some time investigating the issue.

RafaPolit:
Probably related:
#2276
#3378

daneryl:
https://github.com/huridocs/uwazi/blob/32f33ec5e9f167df88bedf664e285c4ca14f0a55/app/api/entities/entities.js#L323
always query/returns all relationship/documents when saving an entity, possible solutions:
- make this optional and do not return by default ? (removing this produces 0 red tests, this suggests that this data is not usually needed when saving an entity)
- do not return realationships and whoever is using this needs to ask for relationships in another way if needed ?

https://github.com/huridocs/uwazi/blob/32f33ec5e9f167df88bedf664e285c4ca14f0a55/app/api/entities/entities.js#L325
- this is called even when there is no need, (no relationship metadata has been changed/send)
- this is still very slow when needed to perform and needs to be remade in order to be efficient and performant ( is it needed to have all relationship/hub loaded in order to perform this process ? )

**the lines above are the main Bottlenecks, both processes end up using **relationships.getByDocument()** which is pretty expensive and loads everything into memory (all relationships of the document), if we have OOM errors during this process, this is the most likely cause**

https://github.com/huridocs/uwazi/blob/32f33ec5e9f167df88bedf664e285c4ca14f0a55/app/api/entities/entities.js#L127
- this changes the label of all the entities that have the saved entity as a relationship
- this is only called when title or icon have changed.
- for entities with a lot of relationships this is slow because after the name change all affected entities have to be indexed

https://github.com/huridocs/uwazi/blob/32f33ec5e9f167df88bedf664e285c4ca14f0a55/app/api/entities/entities.js#L131
- metadata denormalization happens once every entity to sync (once per language configured), the denormalization part of the data
should be identical i think ?, we should be able to reuse it.

## https://github.com/huridocs/uwazi/blob/32f33ec5e9f167df88bedf664e285c4ca14f0a55/app/api/entities/entities.js#L328
- we are reindexing the entity including full text on every save, i think the fullText should not be needed as it can not change by saving an entity itself (needs check)

konzz:
[...]

PR #3649, opened 2021-05-06
saveEntityBasedReferences refactor

### LINKED ISSUE #3638: Refactor saveEntityBasedReferences (13 pts)
The function saveEntityBasedReferences is a function that creates / deletes references based on changes in an entity metadata. For that purpose it requests all the relationships of an entity before doing the process, that is expensive and may not be required in all scenarios.

https://github.com/huridocs/uwazi/blob/32f33ec5e9f167df88bedf664e285c4ca14f0a55/app/api/relationships/relationships.js#L323

- Refactor the function to only request the relationships needed per each property that is been evaluated

fnocetti:
fixes #3638 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/entities/routes.js
+++ b/app/api/entities/routes.js
@@ -9,6 +9,7 @@ import { parseQuery, validation } from '../utils';
 Joi.objectId = objectId(Joi);
 
 export default app => {
+<<<<<<< HEAD
   app.post(
     '/api/entities',
     needsAuthorization(['admin', 'editor', 'collaborator']),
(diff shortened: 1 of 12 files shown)


daneryl:
@fnocetti im also having OOM crashes with relatively small entities (number of connections), now i discovered that we introduced a crazy big bottleneck with https://github.com/huridocs/uwazi/pull/3602 (just after i did the first diagnostic 😢) which is the function call just after the one you just refactored.

https://github.com/huridocs/uwazi/blob/75fad04668851a6789906499e0c7651f1cb88871/app/api/entities/entities.js#L383-L386

the 385 line is the reason of the OOM/performance bottleneck, which of course uses ```relationship.getByDocument``` inside.

to actually benchmark this, i did it without that new bottleneck call to check properly the improvement of this method (10 concurrent connections over 10 seconds, with a 700 connection entity) this are my results, this looks like a pretty good improvement :+1: 

### save without sending metadata:
 - pre-refactor: 7 request/second
 - post-refactor: 110 request/second
 - current-state: [OOM crash](https://github.com/huridocs/uwazi/issues/3600#issuecomment-843033361)

### save sending NOT changing metadata:
 - pre-refactor: 6 request/second
 - post-refactor: 66 request/second
 - current-state: [OOM crash](https://github.com/huridocs/uwazi/issues/3600#issuecomment-843033361)

### save sending CHANGING metadata:
 - pre-refactor: 4 request/second
 - post-refactor: 42 request/second
 - current-state: [OOM crash](https://github.com/huridocs/uwazi/issues/3600#issuecomment-843033361)

daneryl (approved):
perf and memory usage looks good, LGTM

PR #3657, opened 2021-05-11
aggregations for inherited values

### LINKED ISSUE #3636: Filter and Aggregate by Inherited metadata (30pts)
Expected features:
- Filter by inherited metadata instead of the title of the inherited entity
- Show appropriate aggregations for each value.
- Inherited filters should match with common filters.

Implementation decisions
- ~Template properties will have type (text, relationship...) and filterType (relationships will use inherited type)~
- ~The inheritedValue will be moved from inheritedValue to value~
- ~The value will be moved to ref {value: label:} or ref / refLabel~
- ~The label should be replaced with the inherited property label~
- Do not allow inheriting from inherited

Tips
- Make sure copy from copies everything and not just the value
- check permissions ??
- Check if inherited properties are exported with the correct value and just report it but do not develop

Example of inherited values

```
{
  sharedId: 456,
  title: 'Konz the famous magician',
  template: A,
  metadata: {
    favourite-colors: [{value: 28s7d3, label: 'green'}, {value: 234js9d3, label: 'blue'}],
  },
}

{
  sharedId: 859,
  title: 'Rafa the copy cat',
  template: B,
  metadata: {
    inherited-color: [{value: [28s7d3, 234js9d3], label: ['green', 'blue'], ref: {value: '456', label: 'Konz the famous magician'}}],
  },
}
```

daneryl:
fixes #3636 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


database/elastic_mapping/mappings.ts
```diff
@@ -112,6 +112,12 @@ const relationshipType = () => ({
   label: text,
   value: id,
   type: noIndexText,
+  inheritedValue: {
+    properties: {
+      value: id,
```
RafaPolit, line 117: What happens if the inherited value is a plain-storing property like text or number or date? Should this mapping not be dependent upon inherited property?

RafaPolit:
As mentioned in the comment of the code about non-id-value properties, the problem appears to hold true in the functionality:
- inherit a text field
- select that inherited property as "use as filter" in the template
You get an immediate error when navigating to the library or uploads.

app/api/search/specs/fixtures_elastic.js
```diff
@@ -166,9 +168,47 @@ export const fixtures = {
       published: true,
       user: userId,
     },
-    //metadata filters
+    //inherited aggregations
```
RafaPolit, line 171: There is not a single property inheriting non-id values: text properties, numbers, etc.  They should not get aggregated, but they should not break the app neither.  Perhaps, add one such property to ensure all is working as expected? (which the functional tests suggest this actually isn't working as expected)

RafaPolit (changes requested):

database/elastic_mapping/mappings.ts
konzz: Indeed, we will need to put a mapping according to the property being inherited and update the mapping if that changes.

daneryl: oh! that was so obvious, i guess we were too focused on the multiselect/select combos :(

app/api/search/specs/fixtures_elastic.js
daneryl: totally right.

database/elastic_mapping/mappings.ts
daneryl: aparently id mapping works for all the cases and the crashes where produced by other conditions.

RafaPolit (changes requested):

RafaPolit:
Some questions / findings:
- Inherited properties behave a little different to "base" properties when the inherited value is empty.  Instead of not showing the label (on cards, on side-panel, on entity view), it shows the label without any value.  This is working wrongly for properties inheriting `text` fields or `relationship` fields, `image`, etc., but it appears to be correct for multiselects.  Please, check the consistency of this behavior?
- If you edit a "source" value in the library's side-panel and the entity inheriting a value is also visible in the library, the inheriting entity it will not get updated.  Should we do a re-fetch of library search upon editing?  Obviously, if the change is not yet indexed, it will not get updated and we wasted an expensive request.  So not sure about this one.
- When changing an already-existing relationship property to 'inherit', it does not warn about the "template conflict" in client, but it appears to be re-indexing nonetheless in the backend, so it may just be a "lack of warning" issue.

I found two serious bugs when using a migrated existing instance:
- Bug 1: Cannot change relationship target if it has already been assigned!
- Bug 2: TypeError: Cannot read property 'type' of undefined

But, when starting fresh from a blank state, I can no longer reproduce this. Not sure what happened.  I am still keeping this in case the documentation becomes relevant for existing users.

### To Reproduce Bug 1
- create two templates, one inheriting a property from the other
- create two entities to be potential sources of inheritance: A and B
- create an entity X that inherits from A, save
- edit that entity X and try to switch to inherit from B: `validation failed: [0].entityData.metadata['newrelationship2'][0].inheritedValue should be array, [0].entityData.metadata['newrelationship2'] should match some schema in anyOf,`

### To Reproduce Bug 2
- create a `Relationship Type` named `inherited`
- create template `Source`, add a `text` and an `image` field, no edits necessary, save
- create template `Target`, add an inherited relationship filed to the text field
- go to uploads, the error appears

app/react/Library/components/specs/FiltersFromProperties.spec.js
```diff
@@ -147,6 +156,25 @@ describe('FiltersFromProperties', () => {
     });
   });
 
+  describe('when inheriting properties', () => {
+    it('should render the filter apropiate for the inherited property', () => {
```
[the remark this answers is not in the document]
daneryl: changed.

app/api/search/specs/search.spec.js
```diff
@@ -237,35 +237,32 @@ describe('search', () => {
...
-          expect(template1en.rows.length).toBe(3);
-          expect(template2es.rows.length).toBe(1);
-          expect(allTemplatesEn.rows.length).toBe(4);
-          expect(onlyMissing.rows.length).toBe(2);
-          expect(template1AndMissing.rows.length).toBe(5);
-          done();
-        }
-      )
-      .catch(catchErrors(done));
+    ]).then(
```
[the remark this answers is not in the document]
daneryl: bad choice, changed.

app/react/Library/components/specs/FiltersFromProperties.spec.js
```diff
@@ -147,6 +156,25 @@ describe('FiltersFromProperties', () => {
...
+          name: 'countries',
+          label: 'Countries',
+          type: 'relationship',
+          inherit: true,
+          inheritProperty: '1234',
+          options: [{ label: 'option2' }],
+        },
+      ];
+      props.translationContext = 'oneContext';
+      const component = shallow(<FiltersFromProperties {...props} />).find(DateFilter);
```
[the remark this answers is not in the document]
daneryl: i do not know, what do you think ?

app/api/search/specs/fixtures_elastic.js
daneryl: this has been fixed

daneryl:
More errors found:

- inheriting geolocation produces an elastic mapping  error.
- saving from library sometimes produces validation error

database/elastic_mapping/mappings.ts
daneryl: id mapping does not work for geolocation.

konzz:
> Inherited properties behave a little different to "base" properties when the inherited value is empty. Instead of not showing the label (on cards, on side-panel, on entity view), it shows the label without any value. This is working wrongly for properties inheriting text fields or relationship fields, image, etc., but it appears to be correct for multiselects. Please, check the consistency of this behavior?

This is a problem on how the formater renders the data, and probably something we missed in the development / qa in the issue of denormalization, since this is also hapening in dev and out of the scope of filters and aggregations not going to be fix in this PR. Issue #3679 has been created.

> If you edit a "source" value in the library's side-panel and the entity inheriting a value is also visible in the library, the inheriting entity it will not get updated. Should we do a re-fetch of library search upon editing? Obviously, if the change is not yet indexed, it will not get updated and we wasted an expensive request. So not sure about this one.

The problem described is related to denormalization and not filtering or agreggations, also happening in dev. This problem also is a deeper problem that not only affects denormalization but almost anything that should be updated in the library after modifying an entity. How to solve it is not clear. No issue has been created since it's a complex topic that is not clear that we might even want to fix.

> When changing an already-existing relationship property to 'inherit', it does not warn about the "template conflict" in client, but it appears to be re-indexing nonetheless in the backend, so it may just be a "lack of warning" issue.

That re-indexation happens because you are modifying the entities of that template with new inherited values. The warning that some times appears is when there is a map conflict and you will need to re-index all the entities.

> Bug 1: Cannot change relationship target if it has already been assigned!
> Bug 2: TypeError: Cannot read property 'type' of undefined

Can no longer reproduce, probably fixed by some changes we did in that process.

PR #3662, opened 2021-05-12
3548 - Full text snippets search v2

### LINKED ISSUE #3548: Full text search (13pts)
As Uwazi devs we want to add full text search and snippets to the Search V2 so we can use it in the search snippets functionality in the document view.

- Add full text search to the new endpoint
- It should return snippets if requested
- Snippets should have the matching field
- Add e2e for search snippets in the document side panel

Screenshot 2021-04-15 at 11 30 50

mfacar:
fixes #3548 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/search.v2/buildQuery.ts
+++ b/app/api/search.v2/buildQuery.ts
@@ -13,7 +13,7 @@ async function extractSearchParams(query: SearchQuery) {
     const searchMethod = await searchStringMethod(query.filter.searchString);
     return { searchString, fullTextSearchString, searchMethod };
   }
-  return { searchMethod: 'query_string' };
+  return { searchString: query.filter?.searchString, searchMethod: 'query_string' };
 }
 
 export const buildQuery = async (query: SearchQuery, language: string): Promise<RequestBody> => {
--- a/app/api/search.v2/buildQuery.ts
+++ b/app/api/search.v2/buildQuery.ts
@@ -7,23 +7,17 @@ import { permissionsFilters } from './permissionsFilters';
 async function extractSearchParams(query: SearchQuery) {
   if (query.filter && query.filter.searchString && typeof query.filter.searchString === 'string') {
     const regex = /fullText:\s*([^,]*)/g;
-    const fullTextGroups = regex.exec(query.filter.searchString);
-    const fullTextSearchString = fullTextGroups ? fullTextGroups[1] : undefined;
-    const searchString = fullTextGroups
-      ? query.filter.searchString.replace(fullTextGroups[0], '')
-      : query.filter.searchString;
-    const searchMethod =
-      query.filter && query.filter.searchString
-        ? await searchStringMethod(query.filter.searchString)
-        : 'query_string';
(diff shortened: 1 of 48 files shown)


app/api/search.v2/buildQuery.ts
```diff
@@ -1,25 +1,84 @@
...
+                  post_tags: ['</b>'],
+                  encoder: 'html',
+                  number_of_fragments: 9999,
+                  type: 'fvh',
+                  fragment_size: 300,
+                  fragmenter: 'span',
+                  fields: {
+                    'fullText_*': {},
+                  },
+                },
```
daneryl, line 63: @mfacar this should only be needed when asking for snippets right ?

app/api/search.v2/searchResponse.ts
```diff
@@ -0,0 +1,41 @@
...
+import { SearchQuery } from 'shared/types/SearchQueryType';
+
+interface ElasticSearchResults {
+  _index: string;
+  _type: string;
+  _id: string;
+  _source: EntitySchema;
+  // eslint-disable-next-line camelcase
+  inner_hits?: { fullText: { hits: { hits: [{ highlight: {} }] } } };
+}
```
daneryl, line 11: @mfacar there is already an interface defined for elastic results, can you extend that one instead of defining a new one here ? https://github.com/huridocs/uwazi/blob/db8fa58a005d771170ee9cc06f1e02a2a34e1d87/app/api/search/elasticTypes.ts#L18

app/react/Library/actions/libraryActions.js
```diff
@@ -267,9 +267,17 @@ export function updateEntities(updatedDocs) {
...
+    qs.stringify({
+      filter: { sharedId, searchString },
+      fields: ['snippets'],
+    })
+  );
+
   return dispatch =>
-    api.searchSnippets(new RequestParams({ searchTerm, id: sharedId })).then(snippets => {
+    searchAPI.searchSnippets(requestParams).then(({ data }) => {
+      const snippets = data.length ? data[0].snippets : { total: 0, fullText: [], metadata: [] };
```
daneryl, line 280: maybe its a good default that the api returns this null snippet object ?

daneryl (changes requested):

app/react/Library/actions/libraryActions.js
daneryl: forget this.

app/api/search.v2/searchResponse.ts
mfacar: Yes, thanks, it's done

app/api/search.v2/buildQuery.ts
mfacar: I extracted the property into a function in order to make it conditional

daneryl (approved):

PR #3674, opened 2021-05-19
Expose store for pages.

### LINKED ISSUE #3670: Expose store (or at least datasets) to the Page's javascript. (5pts)
i understand the utility, but some of this datasets are huge, and we already have them loaded like this on some scenarios, for instance when you do SSR they are on ```window.__reduxData__.page.datasets``` we will duplicate what already is really big in this scenario.

maybe an alternative is to give the script access to the store itself, that has all that data loaded in memory already which give the script access not only to datasets, but also to everything else ? i guess its potentially a security issue ? but come on, this free JS script its already that.

_Originally posted by @daneryl in https://github.com/huridocs/uwazi/pull/3617#discussion_r632320502_

Zasa-san:
fixes #3670

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


e2e/helpers/entityViewPageFixtures.ts
```diff
@@ -7,4 +7,10 @@ export const contents = `<h1>My entity view</h1>
     <Value path="label" />
   </li>
 </Repeat>
-<br />`;
+<br />
+
+<p id='dataset-content'></p>
+`;
+
+export const script = `var datasets = window.store.getState().page.datasets.getIn(['currentEntity', 'sharedId']);
```
RafaPolit, line 15: This variable is called incorrectly.  This variable is not `datasets`, it is `currentEntitySharedId`. Still, I was thinking it could be useful to `append` this to the actual script, so that users could have access directly to, precisely: `datasets` inside the script, instead of having to getState().

What do you think @daneryl ?

Zasa-san: Regardless of the matter to discuss, I renamed the variable as suggested.

daneryl: @RafaPolit what do you mean ? append the datasets by default ? this is precisely what will cause size issues. if someone is advanced enough to write a custom script i think they can manage to work with the store no ?

RafaPolit: @daneryl I mean include in the page script, forced, in the first line:
```
var datasets = window.store.getState().page.datasets.toJS();
```
That way, users have access to the `datasets` variable without needing to know our store's internal structure, our store's type of objects (page is a JS object, but datasets is an Immutable).

This, more or less, the same as the previous approach in terms of user experience, yet it does not actually include all the datasets in the HTML on server side rendering,  It's just a one-liner string.  It would task the browser with more memory and extra processing, yes.  It's finding the balance between useful and optimized.

What do you think?

PR #3676, opened 2021-05-21
batch indexing with query that contains _id

mfacar:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti:
@mfacar This LGTM. I tested the issue reported: it's easily reproducible in the development branch and fixed here. Also the solution implemented is clear.
While I think this should work in every case, I'm not sure what other cases should be tested so I'll wait for @RafaPolit or @daneryl to take a look.

daneryl:
@mfacar LGTM, i have tested it with diferent datasets. im merging this, still i think we should write tests for this. im creating a tech debt for this.

PR #3684, opened 2021-05-25
test public form without SSR

### LINKED ISSUE #3685: e2e public form flaky
when filling the public form on the e2e, most of the time the JS is not ready yet, only the SSR html is ready, the elements are there so obviously puppeteer tries to fill the form. when react takes over the text added before disappears, its not part of  the state.
I think normally this is not a problem but here there is a bigger delay maybe  because we have the markdown preprocess ?

daneryl:
fixes #3685

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti (approved):
Really smart, LGTM

PR #3695, opened 2021-05-27
3693 Remove sort field from markdown elastic mapping REINDEX NEEDED

### LINKED ISSUE #3693: Elastic search mapping for markdown should not have sort (3 pts)
Our current mapping definition for the markdown field includes the elasticsearch field sort, which is used to sort. In some scenarios we have an error because the markdown text length is too long to be sorted by elasticsearch (`UTF8 encoding is longer than the max length 32766`). We have decided that we don't need sorting by this type of properties.

Task
- Update elasticsearch mapping definition for markdown fields.
- No migration is required but a reindex.

konzz:
fixes #3693

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti (commented):
@konzz the extensive changes in this PR ( :P ) look good.
Also tested the indexing process and it works just fine. I had to introduce the changes manually in the DB though, because there's still a validation in the api limiting the length of the markdown content. Is that intended?

I will wait for your response to merge. If removing the validation is **not** part of the impementation and you need to merge this urgently, you can **consider the QA approved**.

PR #3718, opened 2021-06-02
separation of schema and data validations

### LINKED ISSUE #3472: Server crash on null value (8 pts)
Sending a "null" value as a relationship value causes a server crash:
```
TypeError: value.map is not a function
    at validateRelationshipForeignIds (/home/txau/work/uwazi/app/shared/types/validateMetadataField.ts:99:28)
    at validateMetadataField (/home/txau/work/uwazi/app/shared/types/validateMetadataField.ts:144:13)
```
The code expects `[{value: 'something'}] but just `null`was sent.
We should make sure that the schema is validated before any other validation.

daneryl:
fixes #3472 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/entities/entities.js
+++ b/app/api/entities/entities.js
@@ -17,7 +17,7 @@ import translate, { getContext } from 'shared/translate';
 import { unique } from 'api/utils/filters';
 import { AccessLevels } from 'shared/types/permissionSchema';
 import { permissionsContext } from 'api/permissions/permissionsContext';
-import { validateEntity } from 'shared/types/entitySchema';
+import { validateEntity } from './validateEntity';
 import { deleteFiles, deleteUploadedFiles } from '../files/filesystem';
 import model from './entitiesModel';
 import settings from '../settings';
--- a/app/api/entities/index.js
+++ b/app/api/entities/index.js
@@ -1,5 +1,3 @@
-/** @format */
-
 import model from './entitiesModel';
 import * as endpointSchema from './endpointSchema';
 
--- a/app/api/entities/specs/entities.spec.js
+++ b/app/api/entities/specs/entities.spec.js
@@ -665,6 +665,7 @@ describe('entities', () => {
         })
         .catch(catchErrors(done));
     });
+
     it('should sanitize daterange, removing non valid dates', done => {
       const doc1 = {
         _id: batmanFinishesId,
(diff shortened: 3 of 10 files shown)


fnocetti (commented):
The changes look good to me. I still need to go through the QA, but a question that arises is: now that we separated structure and data consistency, does it make sense to do the structure validation as soon as we get the data (i.e. in the route handler before handling the data to any other process)?

fnocetti:
I did a quick QA and it seems ok.
I was able to get a 500 code by trying to save an entity sending something like metadata: { relationship: somestring }
I wasn't able to reproduce it in any other way but, as far as I understood the changes, it should be enough.

daneryl:
> The changes look good to me. I still need to go through the QA, but a question that arises is: now that we separated structure and data consistency, does it make sense to do the structure validation as soon as we get the data (i.e. in the route handler before handling the data to any other process)?

yes it makes sense to do the schema validation in the route, or at any point  where data is passed into the system, and because i think we do not know how many places we have now that do this, it is safer for now to have it in the entities save function, but this needs to be definitely revisited.

> I did a quick QA and it seems ok.
> I was able to get a 500 code by trying to save an entity sending something like metadata: { relationship: somestring }
> I wasn't able to reproduce it in any other way but, as far as I understood the changes, it should be enough.

i think a 500 should not happen, with the example you have i get a 422 response, can you provide a step by step for this 500 ?

fnocetti:
> yes it makes sense to do the schema validation in the route, or at any point where data is passed into the system, and because i think we do not know how many places we have now that do this, it is safer for now to have it in the entities save function, but this needs definitely to be revisited.

👍 

> i think a 500 should not happen, with the example you have i get a 422 response, can you provide a step by step for this 500 ?

Maybe I should say that I got that 500 when trying to reproduce the issue in the `development` branch! After your changes I indeed get a 422.
In case it is useful, the retro steps (on `development`) are:

- Have a template with a relationship property (it's configuration doesn't really matters)
- Through the API, send an usual entity creation request (I actually copied one from the chrome dev tools) but edit the body so the metadata attribute is: `{ "relationship" : "some string here" }`

fnocetti:
Agreed that this is ok. Merging.

PR #3728, opened 2021-06-03
3701 Elastic mapping improvements

### LINKED ISSUE #3701: Elastic mappings improvements (6 pts)
The amount of mapped fields could be reduced:
- Right now we have a mapping explosion because all properties are being duplicated in suggested metadata. Make sure no mappings are created for suggested metadata if the instance has no topic classification and the field is not of the type select/multiselect.
- Check that searching still works after removing the suggested metadata mapping

Zasa-san:
fixes #3701

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/database/elastic_mapping/elasticMapFactory.ts
+++ b/database/elastic_mapping/elasticMapFactory.ts
@@ -26,7 +26,10 @@ export default {
           map.properties.metadata.properties[property.name] = {
             properties: propertyMappings[property.type](),
           };
-          if (topictopicClassification && property.type === 'select') {
+          if (
+            topictopicClassification &&
+            (property.type === 'select' || property.type === 'multiselect')
+          ) {
             map.properties.suggestedMetadata.properties[property.name] = {
               properties: propertyMappings[property.type](),
             };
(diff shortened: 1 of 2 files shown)


database/elastic_mapping/elasticMapFactory.ts
```diff
@@ -16,6 +16,7 @@ export default {
 
     return templates.reduce(
       (baseMapping: any, template: TemplateSchema) =>
+        // eslint-disable-next-line max-statements
```
mfacar, line 19: @Zasa-san as far as possible we should try to avoid disabling ESLint issues, please can you extract some code into a function in order to pass the check instead

mfacar (changes requested):
LGTM! Nice work, I left just a comment about the ESLint disable line, please check it.

database/elastic_mapping/elasticMapFactory.ts
Zasa-san: I can extract lines 38:51. And replace it for 
`map = inheritedPropertyMapping(map, property, topicClassification);`
where `inheritedPropertyMapping` does what is in 38:51 right now.

Since this is all part of a reduce that appends values to `map` I would say it's ok to ignore the max statements lint issue, unless the function mapping function continues to grow.

mfacar: I agree, a refactor it's not needed, I'll merge it

Zasa-san: Thanks Mercy!

PR #3733, opened 2021-06-07
3665 custom js multiple embedd bug fix

### LINKED ISSUE #3665: Custom Javascript code on page is embedded each time user visits index page (8 pts)
# Problem

Custom javascript used on pages is re-embedded each time user returns to the homepage from other pages.

This is causing problems since all custom JS needs to be checked if it already is embedded so it can properly run, otherwise, things break.

I have tested this in Chrome and Safari, it is not a browser issue.

# Video

Screen Recording 2021-05-14 at 09 47 01 2021-05-14 09_50_48

LaszloKecskes:
Rewritten tests to check document elements instead of using mocking and snapshots. Corrected behavior.

fixes #3665

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


LaszloKecskes:
During review, please note that there was an issue where I have accidentally introduced an extra submodule reference (in [this commit](https://github.com/huridocs/uwazi/commit/54bad473fe9428b261d56a870126dcce0914d7e6)) inside uwazi to itself, which also caused a failing check. The reason is not clear, most probably I have made a mistake with the VSCode Git plugin while setting it up.

Using the git submodule reference, I have managed to remove it (in [this commit](https://github.com/huridocs/uwazi/commit/6898f950a65e7d367559a76b937b4f8f9f0c2c32)), but at this point I would consider scrapping the current branch and pull request, copy paste my changes into a new, clean one (since there isn't a lot), then create a new pull request, rather than leaving something in the git history.

Zasa-san (approved):
LGTM! Very nice =D

PR #3734, opened 2021-06-09
3729 - Old file routes should return a 404

### LINKED ISSUE #3729: (errors cleanup) old file routes should return a 404 instead of a 500 status when file not exist (5 pts)
[staticFileMiddleware](https://github.com/huridocs/uwazi/blob/development/app/api/utils/staticFilesMiddleware.ts) is [used here](https://github.com/huridocs/uwazi/blob/c05141e2af598f03bff4e5235a92a3a0820bec5c/app/server.js#L86) to serve files for routes `'/uploaded_documents/:fileName'` and `'/assets/:fileName'` when a file does not exist the returned status is a 500 instead of a 404.

the new route for serving files ```/api/files/:filename``` does this properly using a diferent implementation, we should use the same implementation if possible for both.

- fix staticFileMiddleware to return a 404 when a file is not found
- use staticFileMiddleware on ```/api/files/:filename``` endpoint to avoid code duplicity

Zasa-san:
fixes #3729

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/utils/staticFilesMiddleware.ts
+++ b/app/api/utils/staticFilesMiddleware.ts
@@ -1,9 +1,11 @@
-import { Request, Response } from 'express';
+import { NextFunction, Request, Response } from 'express';
 import { pathFunction, fileExists } from 'api/files';
+import { createError } from '.';
 
 const staticFilesMiddleware = (pathFunctions: pathFunction[]) => async (
   req: Request,
-  res: Response
+  res: Response,
+  next: NextFunction
 ) => {
   const pathToUse = await pathFunctions.reduce<Promise<pathFunction>>(async (current, path) => {
     if (await fileExists(path(req.params.fileName))) {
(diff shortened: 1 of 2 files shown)


app/api/utils/staticFilesMiddleware.ts
```diff
@@ -1,18 +1,30 @@
...
     }
     return current;
   }, Promise.resolve(pathFunctions[0]));
-
-  res.sendFile(pathToUse(req.params.fileName));
+  try {
+    await checkFilePath(req.params.fileName, pathToUse(req.params.fileName));
+    res.sendFile(pathToUse(req.params.fileName));
+  } catch (e) {
+    next();
```
mfacar, line 26: Although I'm not sure if this could be used, I think it's better to pass the exception to the next function always that it's in into a catch

mfacar (changes requested):
LGTM!, I agree it could be difficult to test this with an integration test, how about adding at least a unit test that checks the change, otherwise the added control is not tested at all

mfacar (approved):
LGTM!

PR #3736, opened 2021-06-10
HOTFIX 3702 perf issues round 2

### LINKED ISSUE #3702: Fix performance issues with denormalize inherited values in backend. Round 2 (30 pts)
When we save an entity we trigger a  "denormalize Inherited metadata" process, it process all the entities related to the entity saved and updates the denormalization, so if the entiy has modified any property thats been inherited, its updated everywhere.

We already tried a first refactor but did not achieve a good performance so we discarded the changes.

The idea will be to start by defining scenarios and create new tests for this, since one of the bigger problems in the first attempt was not having reliable tests.

It's not safe to go alone! Take this.  🗡️

fnocetti:
fixes #3702

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


RafaPolit:
Possible problem.

### To Reproduce
- Template has a multiselect property
- Entity has already assigned a value
- Edit the thesaurus entry to a different label

### Results
- Cards in library are not updated, not even after a refresh
- If I go into entity view the change is actually reflected
- If I edit from library in the side panel, the select has the changes reflected

This suggests that elastic is not picking up this changes.

RafaPolit:
Performance notice.

### To Reproduce
- Have templates A and B, have a few thousand B entities
- Create a Select property in A pointing to Thesaurus X
- Create a Relationship property in B inheriting the Select property from A
- Save

Without any data assigned, since this can come from existing relationships, it took nearly 30 seconds to save the tamplate.

RafaPolit:
Indeed, any template save that involves relationships when template has thousands of entities, results in minute-long saves, which could indicate it is hogging memory and can be potentially detrimental.

RafaPolit:
> Possible problem.
> 
> ### To Reproduce
> * Template has a multiselect property
> * Create an entity with values assigned to that Multiselect
> * Edit the thesaurus entry to a different label
> 
> ### Results
> * Cards in library are not updated, not even after a refresh
> * If I go into entity view the change is actually reflected
> * If I edit from library in the side panel, the select has the changes reflected
> 
> This suggests that elastic is not picking up this changes.

This issue is fixed with the new commit.  Apparently all is working.

RafaPolit:
Other than the performance issue on saving template described above, I have seen no problem for basic conditions of simple values and inheritances.  Have not tried deeply nested inheritances or circular references, but that should not be the address of this hotfix.

From me, green light on this!

fnocetti:
This PR is for a branch based on Development. We are now rebasing to master for the hotfix. I'm closing this.

PR #3738, opened 2021-06-14
Fixed tooltip typo

### LINKED ISSUE #3714: Settings-->Collection-->Private instance tooltip typo
The current text of the tooltip reads: “Private Instance: Check to make this instance private (only logged users access the data)”

This needs to be corrected to say "logged-in users" to avoid confusion, as "logged" by itself simply means "recorded" or "surveilled" (and since we actually have an Activity Log, could be confusing too). It could also read better to say:

"Check to make this instance private (only logged-in users **can** access the data)"

private instance tooltip

A smaller additional edit could be made to "Private instance" by putting "instance" in lowercase, instead of the current capitalized style. This would make it consistent with most of the other Collection settings options that surround it, in which only the first word is captalized:

collection setting titles

grafitto:
fixes #3714 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


Zasa-san (commented):
LGTM!

PR #3748, opened 2021-06-16
3724 search v2 projections

### LINKED ISSUE #3724: Search V2 Projections 📽️ (7 pts)
Next step in the development of the new search api will be to implement [JSON API sparse fieldsets](https://jsonapi.org/format/#fetching-sparse-fieldsets), we currently have a small implementation of this for snippets.

The specification says that when no fields are specified, the server MAY return none, all fields, or a subset, we think there should be a default of returning some or ALL fields (there is currently one, but should be discussed ⚠️ );

We can use this new implementation in the parts of Uwazi that are currently using the Search V2 API, no new parts of Uwazi will be included in this development.

mfacar:
fixes #3724

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/search.v2/buildQuery.ts
```diff
@@ -3,11 +3,12 @@ import { RequestBody } from '@elastic/elasticsearch/lib/Transport';
 import { cleanUp, extractSearchParams, snippetsHighlight } from './queryHelpers';
 import { permissionsFilters } from './permissionsFilters';
 
+const defaultFields = ['title', 'template', 'sharedId'];
 export const buildQuery = async (query: SearchQuery, language: string): Promise<RequestBody> => {
   const { searchString, fullTextSearchString, searchMethod } = await extractSearchParams(query);
   return {
     _source: {
-      includes: ['title', 'template', 'sharedId', 'language', 'documents._id'],
+      includes: query.fields || defaultFields,
```
fnocetti, line 11: Isn't it needed to remove pseudofields like `snippets`?

fnocetti (commented):
The code looks good. I left a comment out of ignorance :P 
I will proceed to the functional testing.

app/api/search.v2/buildQuery.ts
mfacar: We could, but it's not needed. Elastic ignore all the fields that don't exist in the index. That's one of the reasons why at this moment we don't need to validate or sanitize this information because it would imply fetching the valid properties from the DB. Eventually in the future, if we want to apply highlights in some fields we will need to pre-process these values.

fnocetti (approved):
This works nicely 🎉 

Merging.

PR #3770, opened 2021-06-22
3735 denormalization followup

### LINKED ISSUE #3735: Denormalization rewrite follow up (13 points)
This is following #3702 up.
To do list:

- [x] Group the properties to update by name-content-inheriting before updating to avoid multiple unneeded updates. This grouping is a first idea, further optimizations might be possible. Let's investigate.
- [x] Try to optimize that this whole process does not run if there is no changes to denormalize.
- [x] Refactor, clean-up and abstract the process.

daneryl:
fixes #3735 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/entities/denormalize.ts
+++ b/app/api/entities/denormalize.ts
@@ -1,6 +1,5 @@
 import { EntitySchema } from 'shared/types/entityType';
-import { PropertySchema } from 'shared/types/commonTypes';
-import templates from 'api/templates/templates';
+import templates, { DenormalizationUpdate } from 'api/templates/templates';
 import { search } from 'api/search';
 import { TemplateSchema } from 'shared/types/templateType';
 import { WithId } from 'api/odm';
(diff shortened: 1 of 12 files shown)


fnocetti:
@daneryl I think this is ready for review.

konzz:
@daneryl @fnocetti Found an error that I can only reproduce with the development fixtures, but not in blank-state.

- Use mecanism to inherit country code from the template country
- Add a new thesauri
- Add a property select to country using that new thesauri
- Select some values to inherited countries
- Try to edit the thesauri

An error is thrown:

`MongoError: Cannot apply array updates to non-array element inheritedValue: null`

daneryl:
the issue happens when some inheritedValue are null, our update is based on mongodb array updates, and only work when the values are actual arrays.
i have fixed it by defaulting empty values to an empty array instead of null which standarizes inheritedValue to always be an array when present.

PR #3777, opened 2021-06-25
EntityData markdown

### LINKED ISSUE #3663: Implement CurrentEntity markdown component (5pts)
As part of creating entity view pages, the CurrentEntity component will allow a much simpler interaction with the dataset.

RafaPolit:
fixes #3663

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


RafaPolit:
Proposed WIKI:
https://github.com/huridocs/uwazi-documentation/blob/creating_entity_pages/admin-docs/creating-entity-pages.md

app/react/Markdown/components/specs/EntityData.spec.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    component = renderConnectedMount(() => innerComponent, state);
+  };
+
+  describe('root properties', () => {
+    it('should print title and root dates from root of entity', () => {
+      render(<EntityData value="title" />);
+      expect(component.html()).toBe('Entity 1');
+
+      render(<EntityData value="creationDate" />);
```
txau, line 74: Can I suggest the naming convention `<EntityData print="title" />`. Alternatively `display`, `show`, `get`. `render`.

RafaPolit: The issue is that this component can print both a value or a property name.  That is why `value` and `propertyName`  where chosen.  We can go with `printValue` or `printPropertyName`, but we need to disambiguate from both.  What would you with that taken into account?

txau: I see. What do you think of `valueOf='title'`and `labelOf='title'`?

RafaPolit: Well, `valueOf` is a reserved word, so I changed it to:
`value-of`

Since internally we adhere to camel case, internally it is called `propValueOf`.  Not nice, but there was no other workaround to keep the names consistent.

app/react/Markdown/components/EntityData.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+const EntityData = ({
+  entity,
+  templates,
+  thesauri,
+  'value-of': propValueOf,
+  'label-of': propLabelOf,
+  settings,
+}: ComponentProps) => {
+  const newNameGeneration = settings.collection.get('newNameGeneration') || false;
```
daneryl, line 134: if settings is only used to get the flag then move to mapStateToProps.

app/react/Markdown/components/EntityData.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const formattedEntity = formatter.prepareMetadata(entity.toJS(), templates, thesauri);
+  let output = <></>;
+
+  try {
+    const { property, method } = prepareData(propValueOf, propLabelOf);
```
daneryl, line 139: i think names here are too generic, especially the method name, which seems to do more than 1 thing, maybe that is why is hard to name ?

app/react/Markdown/components/EntityData.tsx
```diff
@@ -0,0 +1,150 @@
...
+};
+
+const extractMetadataLabel = ({ formattedEntity, property, newNameGeneration }: Options) => {
+  const propertyData = getPropertyData({ formattedEntity, property, newNameGeneration });
+  return <Translate context={propertyData.translateContext}>{propertyData.label}</Translate>;
+};
+
+const logError = (err: any, propValueOf?: string, propLabelOf?: string) => {
+  /* eslint-disable no-console */
+  console.log('Error on EntityData: ');
```
daneryl, line 70: probably we want to use console.error ?

daneryl (changes requested):
Looks good, please check the minor comments and make changes if you consider is worth it.

app/react/Markdown/components/EntityData.tsx
RafaPolit: Renamed method to renderMethod. It can do four different things:
- extractRootProperty
- extractMetadataProperty
- extractRootLabel
- extractMetadataLabel

I think it reads better now, let me now.

RafaPolit:
> after some functional testing i found out that relationship fields with inherit do not work properly, there is an error in console, are we supposed to support inherit props in this component ?

Well, this is interesting, the formatter (formater) returns the following for a text inherited Metadata:
```
id: "08d8df21-6686-4897-8142-c71efbd9beb9"
indexInTemplate: 6
label: "Country"
localID: "w7uvhip2ggj"
name: "text"
onlyForCards: false
translateContext: "60db77a665f1b277fb63b4ad"
type: "inherit"
value: [
 {value: "Inheritted Text"}
]
_id: "60db77a665f1b277fb63b4ae"
```
So, I have no reference at all from the original property's name.  So I cannot extract its info.

So we either modify something in the formatter or simply say we don't support this for now.  Still, it is a weird decision not to include the actual name of the property in the response from the formatter.

app/react/Markdown/components/EntityData.tsx
daneryl: my main concern was the method name, i do not think prepareData says anything, actually i think you can move what that method does into the render method and everything is clearer.

daneryl:
> > after some functional testing i found out that relationship fields with inherit do not work properly, there is an error in console, are we supposed to support inherit props in this component ?
> 
> Well, this is interesting, the formatter (formater) returns the following for a text inherited Metadata:
> 
> ```
> id: "08d8df21-6686-4897-8142-c71efbd9beb9"
> indexInTemplate: 6
> label: "Country"
> localID: "w7uvhip2ggj"
> name: "text"
> onlyForCards: false
> translateContext: "60db77a665f1b277fb63b4ad"
> type: "inherit"
> value: [
>  {value: "Inheritted Text"}
> ]
> _id: "60db77a665f1b277fb63b4ae"
> ```
> 
> So, I have no reference at all from the original property's name. So I cannot extract its info.
> 
> So we either modify something in the formatter or simply say we don't support this for now. Still, it is a weird decision not to include the actual name of the property in the response from the formatter.

i think it make sense to include the original name ?

app/react/Markdown/components/EntityData.tsx
RafaPolit: Yes, I agree!  Tell that to the complexity analyzer and eslint function length.  If they agree, I am more than happy to change that! :)

RafaPolit:
> i think it make sense to include the original name ?
I actually went the other way.  I cannot think why 'name' of a property should be the inherited one.  The type, the mapping, everything... except the name!  So, name is reverted to the original one, and I have included an `inheritedName` property if there is any need.  All tests seem to pass, so I don't think this breaks anything.

daneryl (approved):

daneryl:
LGTM, @RafaPolit @txau  there is a conversation still pending ? or i can merge this ?

RafaPolit:
Yeah, conversation has been resolved, I don't think there is anything else pending.

PR #3787, opened 2021-06-30
3753 remove edit mode at leaving document

### LINKED ISSUE #3753: Edit mode stays active when leaving the document (8 pts)
1. Upload two different document to uploads
2. Go into view mode for one
3. Click edit on the side panel
4. Click on the uploads button again
5. Go into view mode
6. You see the metadata of the first document. Upon save it gets saved in the first document, but the metadata of the first document is displayed instead of the metadata of the second document.

This is also true for table of contents edition with more critical results since it overwrites the current table of contents of the document. On the top of this, there is no cancel option after entering edit mode for table of contents, forcing the user to hit save.

Edit mode of the relationship tree also is affected although without catastrophic side effects.

mfacar:
fixes #3753

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Viewer/PDFView.js
```diff
@@ -53,6 +54,10 @@ class PDFView extends Component {
     }
   }
 
+  componentWillUnmount() {
+    this.context.store.dispatch(leaveEditMode());
```
konzz, line 58: I think we should connect PDFView instead of using actions this way.

mfacar: It's done. I had some problems at testing but decided to update just for the changed part.

PR #3790, opened 2021-07-02
template to thesauri count now with elastic aggs

### LINKED ISSUE #3786: api/thesauris does very expensive db queries.
this route merges dictionaries and templates into a single "thesauri" collection, is this still needed ?
if yes, then we need to find a solution to make it more performant.

daneryl:
fixes #3786

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/search/search.js
+++ b/app/api/search/search.js
@@ -855,6 +855,21 @@ const search = {
     const templates = await templatesModel.get();
     return updateMapping(templates);
   },
+
+  async countPerTemplate(language) {
+    const queryBuilder = documentQueryBuilder()
+      .language(language)
+      .includeUnpublished()
+      .limit(0);
+
+    return (
+      await elastic.search({ body: queryBuilder.query() })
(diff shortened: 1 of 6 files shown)


app/api/search/search.js
```diff
@@ -852,6 +852,21 @@ const search = {
     const templates = await templatesModel.get();
     return updateMapping(templates);
   },
+
+  async countPerTemplate(language) {
+    const queryBuilder = documentQueryBuilder()
+      .language(language)
+      .includeUnpublished()
```
RafaPolit, line 859: While this, apparently, replicates the original functionality... if you have hundreds of published and unpublished entities, wouldn't this show the 'wrong' amount of available extra options for unlogged users?

daneryl: @RafaPolit Actually this is not the case, the permissions feature takes care of that on the documentQueryBuilder.

PR #3808, opened 2021-07-13
Custom template title translation cleanup

### LINKED ISSUE #3567: Custom template title in translations does not clean previous values (5 pts)
Translations correctly append the custom title you can configure on a template.

However, if you change that value, the previous translation is kept when that key should no longer exist on that context.

## To reproduce

- Change the name of the "title" property in the template (for ex. to "Another name")
- Check that the translations correctly have this new title added as a key
- Edit the name in the template once more to "Something different"
- Go into translations and check that BOTH "Another Name" and "Something Different" are present there

Remember to add a migration to remove the wrong values from the translations dictionaries.

LaszloKecskes:
fixes #3567 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/migrations/migrations/47-delete-superfluous-title-translations/index.js
```diff
@@ -0,0 +1,42 @@
...
+        translation.contexts.map(async context => {
+          if (context.type === 'Entity') {
+            const template = await db
+              .collection('templates')
+              .findOne({ _id: ObjectId(context.id) });
+
+            if (template) {
+              const templateLabelSet = new Set([
+                template.name,
+                'Title',
```
fnocetti, line 29: Why is Title needed here?

app/api/migrations/migrations/47-delete-superfluous-title-translations/index.js
```diff
@@ -0,0 +1,42 @@
...
+                template.commonProperties.find(p => p.name === 'title').label,
+                ...template.properties.map(p => p.label),
+              ]);
+              context.values = context.values.filter(element => templateLabelSet.has(element.key));
+            }
+          }
+          return Promise.resolve();
```
fnocetti, line 36: Is this return needed?

fnocetti (commented):
The code LGTM. I left 3 minor comments.
I will proceed to the manual testing.

fnocetti:
From the manual QA it looks like the functionality and the migration work as expected, but they're always keeping the translations for 'Title' in place, which is something that looks like it's by design in your code. Is that ok?

app/api/templates/utils.ts
```diff
@@ -135,3 +136,12 @@ export function getDeletedProperties(
     .filter(notIncludedIn(flattenProperties(newProperties)))
     .map(property => property[prop]);
 }
+
+export function getDeletedCommonProperties(
```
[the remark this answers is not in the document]
LaszloKecskes: It seems to me that the `notIncludedIn` function tests inclusion by id. In this case, the commonProperty does not get deleted, just the name is changed, it has the same id.

I agree that the design of the function is a bit wonky, since at the moment, this is the only commonProperty where we care about this, so putting it into a separate function might be too much. In this case, I went with readability, since this way it follows the same pattern as with the properties. Probably the naming could've been different. We can discuss this.

app/api/migrations/migrations/47-delete-superfluous-title-translations/index.js
LaszloKecskes: It isn't. I thought Promise.all will complain, but it looks like it's more clever than I gave it credit for.

LaszloKecskes:
> From the manual QA it looks like the functionality and the migration work as expected, but they're always keeping the translations for 'Title' in place, which is something that looks like it's by design in your code. Is that ok?

It was intentional. When Rafa presented the issue, it was my impression that we want to keep that. It might be that I misunderstood, we can ask him.

app/api/templates/utils.ts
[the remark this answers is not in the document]
fnocetti: Makes total sense, thanks for clarifying! Maybe yes, we can try to figure out a better naming. In fact, common properties cannot be deleted (which also isn't the case for your algorithm). I think something around `renamed` would be better. Or even better, let's make it refer right to the title, since it's the only thing it's acting on. Yeah, let's have a chat!

app/api/migrations/migrations/47-delete-superfluous-title-translations/index.js
fnocetti: It's a combination. Promise.all will accept non-promise values and treat them as if you returned Promise.resolve(value). But also, since you made your callback `async` it is always returning a promise. In your case, since you include no return value, it is returning a promise that resolves to an empty value, which is in fact the same you tried to achieve with your return :)

fnocetti:
> It was intentional. When Rafa presented the issue, it was my impression that we want to keep that. It might be that I misunderstood, we can ask him.

Let's have a chat about this!

fnocetti (approved):
This LGTM. Waiting for the checks to run

PR #3819, opened 2021-07-19
Multiinheritance

### LINKED ISSUE #3667: Relationships multiinheritance 🚀 (20 pts)
Tasks:
- Allow multiple inheritance via different fields
- On edit we will group all relationships that are the same in only one relationship field
- The API should validate that we are sending the same values or return an error
- Remove template validation that prevents multiple relationship fields with the same configuration

WARNING ⚠️ 
We have two proceses that are affected by this:
- Generate relationships based on the metadata
- Modify metadata based on relationships

konzz:
fixes #3667

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/entities/validation/validateMetadataField.ts
```diff
@@ -122,6 +123,41 @@ const validateRelationshipForeignIds = async (
...
+    template.properties?.filter(
+      p =>
+        p.type === propertyTypes.relationship &&
+        p.content?.toString() === property.content?.toString() &&
+        p.relationType?.toString() === property.relationType?.toString()
+    ) || [];
+
+  const valid = Boolean(
+    sameProps.every(
+      p => !entity.metadata || JSON.stringify(entity.metadata[p.name]) === JSON.stringify(value)
```
fnocetti, line 146: Is there a chance for the metadata values to be the same but be in different order, making this stringify comparison fail?

app/react/Metadata/components/MetadataFormFields.js
```diff
@@ -120,6 +148,7 @@ export class MetadataFormFields extends Component {
             options={thesauri}
             totalPossibleOptions={totalPossibleOptions}
             prefix={_model}
+            onChange={this.relationshipChange.bind(this, property)}
```
fnocetti, line 151: Please check this eslint warning. The recommended approach is to do the binding in the class constructor.

app/react/Metadata/helpers/groupRelationshipFields.ts
```diff
@@ -0,0 +1,31 @@
+import { PropertySchema } from 'shared/types/commonTypes';
+
+export function groupSameRelationshipFields(fields: PropertySchema[]) {
```
fnocetti, line 3: This function looks like it's the same implemented in `MetadataFormFields` and it's not used anywhere. Can you please take a look?

fnocetti (changes requested):
The code LGTM, I left a couple of quick comments.

app/react/Metadata/components/MetadataFormFields.js
konzz: How do you suggest doing the binding in the class constructor but also bind different properties?

app/react/Metadata/helpers/groupRelationshipFields.ts
konzz: yep, deleted it, it was for the previous approach of the form

app/api/entities/validation/validateMetadataField.ts
konzz: there is a new implementation

fnocetti (approved):
The code LGTM. Now @RafaPolit is in charge of the manual testing.

RafaPolit:
I'm really impressed with this.  It mostly works as expected.  Very minor details, mostly related to the user experience, nothing critical, mostly presentational.

Functional QA:

- The warning message upon property creation is wrong.  When first configuring a second relationship of the same type, it reads: "Relationship fields must have diferent (with just one F!!!) relationship or diferent (again one F!) type of entity".  This is no longer true.  This only shows when the "any entity" is selected.  So, I think that something like "Cannot use 'any entity or document' if another relationship of the same type is already with a specific entity". Or something like that.
- The "(affects...)" concatenation is not putting spaces between multiple inherited fields.  It reads `(affects "Place of Detention""Current Country""Guarded by")`
- Inherited properties that have a related-entity value, but on which the target entity has NO value, ends up showing the label with no value displayed.  So, if Entity A inherits property X from Entity B, and Entity B has property X empty, you see the label for the property, but no value.  The property label should not appear.  When there is more than one entity in the same property and one of the targets have empty values, we get weird comma concatenations: "Value 1, ,Value 3,".  For example here, "Value 2" and "Value 4" are empty, but the formatter or renderer is still trying to include those values.
- Inheriting from entities that have the same value, result in the same value appearing multiple times.  For example: Entity A is inheriting property X from Entities B and C.  Entities B and C both have value "bla" on property X.  On entity A you see: "bla, bla".

konzz:
> * The warning message upon property creation is wrong.  When first configuring a second relationship of the same type, it reads: "Relationship fields must have diferent (with just one F!!!) relationship or diferent (again one F!) type of entity".  This is no longer true.  This only shows when the "any entity" is selected.  So, I think that something like "Cannot use 'any entity or document' if another relationship of the same type is already with a specific entity". Or something like that.

Changed to the suggested message.

> * The "(affects...)" concatenation is not putting spaces between multiple inherited fields.  It reads `(affects "Place of Detention""Current Country""Guarded by")`

Added a blank space, but sure if we prefer a coma.

> * Inherited properties that have a related-entity value, but on which the target entity has NO value, ends up showing the label with no value displayed.  So, if Entity A inherits property X from Entity B, and Entity B has property X empty, you see the label for the property, but no value.  The property label should not appear.  When there is more than one entity in the same property and one of the targets have empty values, we get weird comma concatenations: "Value 1, ,Value 3,".  For example here, "Value 2" and "Value 4" are empty, but the formatter or renderer is still trying to include those values.

This is not really related to multinheritance but any inheritance, and I think @fnocetti is working in a fix for this.

> * Inheriting from entities that have the same value, result in the same value appearing multiple times.  For example: Entity A is inheriting property X from Entities B and C.  Entities B and C both have value "bla" on property X.  On entity A you see: "bla, bla".  This one we could let go for now, as it is not sure how this needs to work.

Yep, this is a complex problem, but is about inheritance and not only affecting multinheritance.

RafaPolit:
There are a couple of codeclimate issues complaining about the use of the quote.  Also, maybe that string be made into a translatable item?

PR #3833, opened 2021-07-23
2911 error on page parsing

### LINKED ISSUE #2911: error on page parsing crashes application. (8 pts)
should we catch errors on custom pages to prevent the app from crashing ? showing an error on the page instead, this way the user can still use the app.

mfacar:
fixes #2911

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/utils/appContextMiddleware.ts
```diff
@@ -4,6 +4,7 @@ import { appContext } from 'api/utils/AppContext';
 const appContextMiddleware = (_req: Request, _res: Response, next: NextFunction) => {
   appContext
     .run(async () => {
+      appContext.set('requestId', Math.floor(Math.random() * 10000));
```
daneryl, line 7: @mfacar i think this is better placed in its own middleware like the multitenant one ?

app/react/Pages/components/Script.js
```diff
@@ -13,6 +13,7 @@ class Script extends Component {
 
   componentDidMount() {
     this.appendScript();
+    window.onerror = this.props.onError;
```
daneryl, line 16: @mfacar this seems like will have a similar problem the script had before.
window.onerror will not be changed until you navigate to a page, from that point window.onerror will always be this.props.onError, even if you navigate out of the page, should we remove callback using the same logic as the script itself ?
see ```componentDidUpdate``` and ```componentWillUnmount``` or maybe there is another way to catch errors on the script ?

app/react/App/ErrorHandling/ErrorUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+    error?: string;
+    prettyMessage: string;
+    requestId?: number;
+  };
+  status?: number;
+}
+
+export const getRenderError = (apiResponse: APIError) => {
+  if (!apiResponse || !apiResponse.json) {
+    return {};
```
daneryl, line 19: @mfacar why is this case needed ? and if its actually needed at least the returned error should have a default name or message ? like the one below ? ```unexpected error```

app/react/App/ErrorHandling/ErrorUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+export interface APIError {
+  json: {
+    error?: string;
+    prettyMessage: string;
+    requestId?: number;
+  };
+  status?: number;
+}
+
+export const getRenderError = (apiResponse: APIError) => {
```
daneryl, line 17: @mfacar do not understand the name of the function, what is this supposed to do ? normalize the error ?

daneryl (changes requested):
looks good in general @mfacar, im concerned with the way the new requestId is handled on the specs, i have expanded on this on one of the code comments. still missing the functional qa, but i think there is enough here for now.

app/api/files/specs/downloadRoute.spec.ts
```diff
@@ -15,6 +15,7 @@ jest.mock(
     next();
   }
 );
+jest.mock('api/utils/AppContext');
```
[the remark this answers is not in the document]
mfacar: As agreed, testingEnvironment is now responsible for the setup. At the moment just used by the affected tests.

app/api/utils/appContextMiddleware.ts
mfacar: added requestIdMiddleware

app/react/Pages/components/Script.js
mfacar: Yes, thanks, I changed to `window.removeEventListener` instead and used the same logic than script

app/react/App/ErrorHandling/ErrorUtils.ts
mfacar: It was there just to add the condition, I moved to a function before calling

app/react/App/ErrorHandling/ErrorUtils.ts
mfacar: Yes, it adapt the object to the needed by the UI, I changed the name

daneryl (changes requested):
@mercy changes look good. 
im having problems with the tests, there is something that is conflicting and making some tests flaky with the new setUp function. 

this is the example i have used to replicate:

```jest app/api/search.v2 app/api/files``` when launching this all in parallel the majority of times the tests fail. when i change the setup of search.v2 permissions.spec to the old clearAndLoad all works as expected.

mfacar:
thank you @daneryl, the command to replicate is very useful, I am looking for the problem.

mfacar:
@daneryl, I found the cause for the flaky tests, it's related to the nested promises at setting up the fixtures, I did a split like a workaround for the problem but I'm not sure about a final solution, maybe it's needed to implement a more robust handler of asynchronous queues.

daneryl (changes requested):
changes looks good, all tests green for me now.
after the functional qa i found only one thing left, i think we should also add eventlistener for ```onunhandledrejection``` event to handle errors inside a async execution on the custom script explained [here](https://stackoverflow.com/questions/55178612/why-does-onerror-not-intercept-exceptions-from-promises-and-async-functions?answertab=active#tab-top).

to replicate this you can throw an error inside a promise in the custom script in uwazi and see that it is not properly handled.

mfacar:
> changes looks good, all tests green for me now.
> 
> after the functional qa i found only one thing left, i think we should also add eventlistener for `onunhandledrejection` event to handle errors inside a async execution on the custom script explained [here](https://stackoverflow.com/questions/55178612/why-does-onerror-not-intercept-exceptions-from-promises-and-async-functions?answertab=active#tab-top).
> 
> to replicate this you can throw an error inside a promise in the custom script in uwazi and see that it is not properly handled.

great I missed that totally, thanks @daneryl, it's added

daneryl (approved):

PR #3845, opened 2021-08-02
3799 information extraction dashboard

### LINKED ISSUE #3799: Information Extraction Dashboard (7 pts)
For the feature extraction we need a dashboard to access each property that can be extracted. This section will be behind a feature toggle / configuration that for the MVP will be done manually in the DB and won't have a configuration panel. This configuration should include the template and which properties are being extracted.

This panel will only represent the metadata extraction configuration, buttons wont do anything for now, and will be implemented in consecutive issues.

Link to the [designs](https://www.figma.com/file/h3mhxBGiRctmF72HLf0qld/UWAZI---ML---information-extraction?node-id=1270%3A2824)

Next screenshot is just for estimation purposes and quick reference, for actual size of the elements, refer to the original designs in the link above
Screenshot 2021-07-08 at 11 37 14

LaszloKecskes:
fixes #3799 

PR checklist:
- [x] Update READ.me ?
- [x] Update API documentation ?

QA checklist:
- [x] Test for side effects
- [x] UI responsiveness
- [x] Cross browser testing
- [x] Code review


gabriel-piles:
One small thing, in templates settings the properties are grouped by "name" but in this dashboard they are grouped by "label".

const prop =
          template.get('properties')?.find(p => p?.get('label') === propLabel) ||
          template.get('commonProperties')?.find(p => p?.get('label') === propLabel);

This doesn't work well when you have labels like :

* Date Adopted
* date adopted

They are the same property in template settings, but not in this dashboard.

LaszloKecskes:
Nice catch. [This function](https://github.com/huridocs/uwazi/blob/e95450312781d3ece88ac5a57fd4ae5a0aafa383/app/react/Templates/components/FilterSuggestions.tsx#L39) handles finding the shared properties in the template property creation. Based on this, I'll just do the same fpr this comparison (i.e. .trim().toLowerCase() ). Also, the string that is written in the table is the string that is given in the settings, so that isn't affected.

However for a more general thing that is out of the scope of this issue: date_adopted would be different, but transformed to the same name. Can this be a problem? I.e. the existence of labels 'one property' and 'one_property' with the same name 'one_property', but potentially different types.

LaszloKecskes:
Scratch that, it will be affected. So, design decision: given a bunch of equivalent strings via capitalization (shared, Shared, sHaReD), which one would we like to be used in the table? Options I see:
1, default all-lowercase
2, the first encountered 
3, all of them, separated by coma

gabriel-piles:
@LaszloKecskes How about we activate the feature toggle using the property name, that is stable across properties, and we avoid the uppercase problem?

gabriel-piles:
When a property name, from the database toggle list, is not found, the Metadata Extraction page fails to load with the following error:

Property "Title2" not found on template "New_template"

This can happen if an admin change a property name in the templates setting so probably it is better to ignore the not matching properties names.

PR #3869, opened 2021-08-17
Lazy load

### LINKED ISSUE #1853: Lazy load libraries (13 pts)
[Edited by Konz]

We want to lazy load some heavy weight libraries to lighten our build, so users have a better experience loading Uwazi. This was done before in [Configure2fa](https://github.com/huridocs/uwazi/blob/development/app/react/Auth2fa/Configure2fa.tsx) to lazy load 'qrcode.react'.

- Lazy load mapbox.
- Lazy load recharts.

It would be great to know how much the build is reduced and the impact on loading Uwazi.
⚠️  Will this have a negative impact on e2e ?

konzz:
fixes #1853

Development:
| Branch | Total size | JS | Total load localhost | Total load Fast 3g |
| --- | --- | ---- | ----- | ---- |
| Development | 2.1mb | 1.7mb | 862ms | 13.98s |
| 1853-lazy-load | 1.1mb | 942kb | 797ms | 8.61s |

PR checklist:
- [x] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Viewer/components/index.js
```diff
@@ -0,0 +1,3 @@
+import Document from './Document.js';
+
+export { Document };
```
daneryl, line 3: use export from syntax ```export { default as Document } from './Document.js';```

app/shared/detectLanguage.ts
```diff
@@ -0,0 +1,6 @@
+import franc from 'franc';
+import get from 'shared/languagesList';
+
+export default {
```
daneryl, line 4: do not use export default for new files, i think is time we add https://github.com/import-js/eslint-plugin-import/blob/master/docs/rules/no-default-export.md to our rule set as a warning maybe ?

package.json
```diff
@@ -285,6 +286,7 @@
     "terser-webpack-plugin": "^4.2.0",
     "wait-for-expect": "^3.0.2",
     "webpack": "4.44.1",
+    "webpack-bundle-analyzer": "^4.4.2",
```
daneryl, line 289: this is nice to have and i think we should maintain it here, can you add a yarn script for ease of use please ? or maybe document how to use it in the README ?

app/react/Viewer/components/Viewer.js
```diff
@@ -28,11 +28,12 @@ import {
 import { openPanel } from '../actions/uiActions';
 import { selectDoc } from '../selectors';
 import ConfirmCloseForm from './ConfirmCloseForm';
-import SourceDocument from './SourceDocument';
-import TargetDocument from './TargetDocument';
+
 import ViewMetadataPanel from './ViewMetadataPanel';
 import ViewerDefaultMenu from './ViewerDefaultMenu';
 import ViewerTextSelectedMenu from './ViewerTextSelectedMenu';
+import SourceDocument from './SourceDocument';
```
daneryl, line 35: why is this import order change needed ? does this has something to do with circular dependencies ?

daneryl (changes requested):
@konzz  looks good, mostly minor comments.

the thing im not convinced on is to put the loadables in the index, which forces us to import and then export on the index.js, this has been how we do manage index files for a long time but i think we should not do it anymore, the ```export from``` syntax is shorter and  helps with circular dependency issues. 
im good with this for now since only a couple of index where touched but we should keep an eye or create a tech debt ?

app/react/Viewer/components/Viewer.js
konzz: no, just that at first I did those loadables, and then I changed it back, and ended in a different order

app/react/Viewer/components/index.js
konzz: changed

app/shared/detectLanguage.ts
konzz: changed, I did it export default to keep it as it was in the previous file.

app/api/files/PDF.js
```diff
@@ -1,7 +1,7 @@
 import EventEmitter from 'events';
 import fs from 'fs';
 import path from 'path';
-import languages from 'shared/languages';
+import { detect } from 'shared/detectLanguage';
```
daneryl, line 4: this variable name is too generic, please change to detectLanguage or something similar ?

daneryl (changes requested):
only one more comment on the code, also please review and try to fix codeclimate error.

app/api/files/PDF.js
konzz: changed

daneryl (approved):

PR #3890, opened 2021-08-28
Checkt translations

konzz:
Script to check untranslated texts in Uwazi

The script parses all react code and alerts of texts not enclosed in a Translate element or "t" function and reports File / Text / Element. Right now still does not detect other scenarios like texts in the Confirm or Notify methods. 

It also checks every string in the app against the System translation and reports missing translations in this context and reports File / Text.

If the scripts finds any text in any of the categories exits with code 1.

Screenshot 2021-08-28 at 20 02 03

- Added script `yarn check-translations` (runs in ~2s)
- Added CI workflow check-translations (it only takes ~1min)
- Added ~227 Missing `<Translate>` elements
- Migration to add 383 Missing translations to the system context
- Updated several "t" functions to the newer component Translate

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


.github/workflows/ci_check_translations.yml
```diff
@@ -0,0 +1,43 @@
...
+    branches:
+      - master
+      - development
+      - release
+  pull_request:
+
+jobs:
+  check_translations:
+    runs-on: ubuntu-18.04
+    services:
```
LaszloKecskes, line 14: Do we need the services for this? We are just running static code analysis.

.github/workflows/ci_check_translations.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          node-version: '14.6.x'
+      - name: Cache node modules
+        uses: actions/cache@v2
+        with:
+          path: ./node_modules
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+      - name: install dependencies
+        run: yarn install
+      - run: yarn blank-state
```
LaszloKecskes, line 37: Do we need the steps after this? Also, can't we get away with just installing the dependencies strictly for the script (babel, and anything else)?

app/api/migrations/migrations/50-add_system_key_translations/specs/fixtures.js
```diff
@@ -0,0 +1,88 @@
...
+      ],
+    },
+    {
+      _id: db.id(),
+      locale: 'en',
+      contexts: [
+        {
+          ...commonContext,
+          values: commonContext.values.concat([
+            { key: 'Priority sorting', value: 'Priority sorting' },
```
LaszloKecskes, line 67: I could not find a reference to this value in the tests.

checkTranslations.mjs
```diff
@@ -0,0 +1,171 @@
...
+      const res = resolve(dir, dirent.name);
+      return dirent.isDirectory() ? getFiles(res) : res;
+    })
+  );
+  return Array.prototype
+    .concat(...files)
+    .filter(
+      file =>
+        !file.match('.spec') &&
+        (file.endsWith('.js') || file.endsWith('.ts') || file.endsWith('.tsx'))
```
LaszloKecskes, line 21: While the project does not seem to contain any, I would still add jsx as well, just to be safe.

.github/workflows/ci_check_translations.yml
konzz: It is connecting to the DB to check if the values are in the context

.github/workflows/ci_check_translations.yml
konzz: Yes, we need the blank-state to check what is in the system context, we could not install other dependencies, but since it's cached from other CI workflows we are not increasing the time

.github/workflows/ci_check_translations.yml
LaszloKecskes: Of course, got it.

checkTranslations.mjs
```diff
@@ -0,0 +1,171 @@
...
+  process.stdout.write(
+    ` === Found \x1b[31m ${textsNotInTranslations.length} \x1b[0m texts not in translations collection ===\n`
+  );
+};
+
+async function checkTranslations(dir) {
+  const files = await getFiles(dir);
+
+  const results = await Promise.all(files.map(file => parseFile(file)));
+  const allTexts = results.reduce((acc, curr) => acc.concat(curr), []);
```
LaszloKecskes, line 155: Maybe mistaken and extremely nitpicky, but maybe results.[flat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat)()?

checkTranslations.mjs
```diff
@@ -0,0 +1,171 @@
...
+    let key = text.key || text.text;
+    key = key.trim().replace(/\n\s*/g, ' ');
+    return !systemContext.values.find(t => t.key === key);
+  });
+
+  client.close();
+  return textsNotInTranslations;
+};
+
+const reportNoTranslaeElement = textsWithoutTranslateElement => {
```
LaszloKecskes, line 124: Typo in "...Translate...".

app/react/App/Home.js
```diff
@@ -1,15 +0,0 @@
-import React, { Component } from 'react';
```
LaszloKecskes, line 1: Was this an unused component?

app/react/Documents/components/DocumentSidePanel.js
```diff
@@ -368,11 +371,11 @@ export class DocumentSidePanel extends Component {
               <TabContent for="toc" className="toc">
                 <div className="tocHeader">
                   <h1>
-                    <Translate>Table of contents </Translate>
+                    <Translate>Table of contents</Translate>
                   </h1>
                   &nbsp;
                   <TocGeneratedLabel file={this.props.file}>
-                    <Translate>auto-created ⓘ </Translate>
+                    <Translate>auto-created ⓘ</Translate>
```
LaszloKecskes, line 378: Do we want to translate with the small icon?

LaszloKecskes:
I've left some minor comments, but I consider only the jsx one as somewhat noteworthy. Aside from that, looks great and works as intended. 
I've learned a lot, babel traverse is nice.

app/react/App/Home.js
konzz: yes

checkTranslations.mjs
konzz: done

app/react/Documents/components/DocumentSidePanel.js
konzz: moved the icon outside

checkTranslations.mjs
konzz: much better, changed

LaszloKecskes (approved):
Looks good, merging.

PR #3893, opened 2021-08-31
Fixed error display when csv upload fails

### LINKED ISSUE #3865: Errors at importing not valid entities aren't handled on the unified Library
Before #3704, errors at importing entities were displayed in the side panel, now the import is suspended after receiving an error and no message is displayed.

Steps to reproduce:
1. Go to Library
2. Click on Import CSV button
3. Choose a file with an error, for example, an entity without a title.
4. Click on Import

Expected:
Screen Shot 2021-08-16 at 20 49 05

Actual:
Screen Shot 2021-08-16 at 20 41 17

grafitto:
fixes #3865 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Library/Library.js
```diff
@@ -6,7 +6,6 @@ import LibraryModeToggleButtons from 'app/Library/components/LibraryModeToggleBu
 import requestState from 'app/Library/helpers/requestState';
 import LibraryLayout from 'app/Library/LibraryLayout';
 import { wrapDispatch } from 'app/Multireducer';
-import ImportProgress from 'app/Uploads/components/ImportProgress';
```
mfacar, line 9: @grafitto It seems like you brought some of the code from this component. Why do we need to replicate the code instead of reusing it?

app/react/Library/Library.js
```diff
@@ -21,13 +20,7 @@ export default class Library extends RouteHandler {
...
   }
 
-  static renderTools() {
-    return (
-      <>
-        <ImportProgress />
-      </>
-    );
-  }
+  static renderTools() {}
```
mfacar, line 23: Without `Import progress` why are you keeping this definition?

app/react/Uploads/components/ImportPanel.js
```diff
@@ -103,11 +126,26 @@ export class ImportPanel extends Component {
...
+            <Icon icon="exclamation-triangle" />
+            <div className="force-ltr">
+              <Translate>The import process threw an error:</Translate>
+            </div>
+            <Icon style={{ cursor: 'pointer' }} icon="times" onClick={this.closeError} />
+          </div>
+          <StackTrace message={this.importError.error} validations={this.importError.validations} />
+        </>
+      );
+    }
```
mfacar, line 146: This section is the same in `ImportProgress`, I think the solution is not to duplicate this code but diagnose why it was not working correctly

app/react/Uploads/components/specs/ImportPanel.spec.js
```diff
@@ -56,5 +55,17 @@ describe('ImportPanel', () => {
...
     });
+    describe('error', () => {
+      describe('when there is an error', () => {
+        it('should render an error message', () => {
+          props.importStart = false;
+          props.importProgress = 0;
+          props.uploadProgress = 0;
+          render();
+          component.setState({ showError: true });
+          expect(component).toMatchSnapshot();
```
mfacar, line 66: This test is pretty basic and is using the undesired snapshot comparison, also, it's pretty similar to the existing at `ImportProgress.spec.js`

mfacar (changes requested):
@grafitto, I'm not sure about the approach that you took to solve the issue, please review the comments in the code. I'll start doing the functional revision.

app/react/Uploads/components/specs/ImportPanel.spec.js
grafitto: Removed

app/react/Uploads/components/ImportPanel.js
grafitto: I have update the codebase. It seems the server returns `importError.prettyMessage` instead of `importError.message`

app/react/Library/Library.js
grafitto: Reused

app/react/Library/Library.js
```diff
@@ -6,7 +6,7 @@ import LibraryModeToggleButtons from 'app/Library/components/LibraryModeToggleBu
 import requestState from 'app/Library/helpers/requestState';
 import LibraryLayout from 'app/Library/LibraryLayout';
 import { wrapDispatch } from 'app/Multireducer';
-import ImportProgress from 'app/Uploads/components/ImportProgress';
+// import ImportProgress from 'app/Uploads/components/ImportProgress';
```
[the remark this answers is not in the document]
grafitto: Yea you are right. the server seems to return `importError.prettyMessage` instead of `importError.message`

app/react/Uploads/components/specs/__snapshots__/ImportPanel.spec.js.snap
```diff
@@ -1,5 +1,54 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`ImportPanel rendering states error when there is an error should render an error message 1`] = `
```
mfacar, line 3: Please update, I think this snapshot is not required now

mfacar (changes requested):
@grafitto LGTM!, there are a couple of snapshots that should be updated. I did the functional test and everything works as expected, I'll merge it as soon as the unit tests pass unless you still are working to add an e2e test for this?

app/react/Library/Library.js
grafitto: In `App` component, the function is called in a loop of many components, and all components looped are expected to have the function, otherwise there is a `function is undefined` error

app/react/Uploads/components/specs/__snapshots__/ImportPanel.spec.js.snap
grafitto: Done

grafitto: Done

mfacar (approved):
LGTM!

PR #3913, opened 2021-09-08
3815 removing references to pdfinfo

### LINKED ISSUE #3815: pdfInfo error on creating text references (8 pts)
image

mfacar:
fixes #3815 

QA checklist:
- [ ] Test for side effects
- [ ] Code review


daneryl (changes requested):
@mercy i see that the migration removes `pdfInfo` from entities, i thought this was wrong since the `pdfInfo` should be in files, but its true entities also have it on some datasets, i do not know why, you should add to the migration removing also from files collection, which is where the majority of the `pdfInfo` are. 

everything else looks good, will do a functional QA when this is done so i can test on real dataset.

mfacar:
> @mercy i see that the migration removes `pdfInfo` from entities, i thought this was wrong since the `pdfInfo` should be in files, but its true entities also have it on some datasets, i do not know why, you should add to the migration removing also from files collection, which is where the majority of the `pdfInfo` are.
> 
> everything else looks good, will do a functional QA when this is done so i can test on real dataset.

You're totally right. I've added the removal of pdfInfo from files, thank you @daneryl.

app/api/migrations/migrations/51-remove-pdfinfo-property/index.js
```diff
@@ -8,9 +8,17 @@ export default {
...
       .updateMany({}, { $unset: { pdfInfo: 1 } });
-    process.stdout.write(`${modifiedCount} entities updated...\r\n`);
+
+    process.stdout.write(`${modifiedEntitiesCount} entities updated...\r\n`);
+
+    const { modifiedFilesCount } = await db
+      .collection('files')
+      .updateMany({}, { $unset: { pdfInfo: 1 } });
+
+    process.stdout.write(`${modifiedFilesCount} files updated...\r\n`);
```
daneryl: @mfacar i think this change broke the report because there is no modifiedFilesCount or modifiedEntitiesCount on the returned object from db, now the migration always says ```undefined entities updated...``` and ```undefined files updated...```

daneryl (changes requested):
@mfacar one thing broke with the change and is commented on the code, everything else looks and works good, ill merge when this is fixed.

app/api/migrations/migrations/51-remove-pdfinfo-property/index.js
mfacar: ohh!, thanks @daneryl, I changed the variable name by mistake, it's fixed

daneryl (approved):

PR #3914, opened 2021-09-10
Removed x-requested-with header from auth middleware

grafitto:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/api.js
```diff
@@ -1,9 +1,11 @@
 /* eslint-disable global-require */
 import activitylogMiddleware from './activitylog/activitylogMiddleware';
+import headersMiddleware from './auth/headersMiddleware';
```
daneryl, line 3: @grafitto change the name to be more descriptive, headers is too generic, this middleware is here specifically to prevent CSRF attack

app/api/auth/headersMiddleware.ts
```diff
@@ -0,0 +1,12 @@
+import { Request, Response, NextFunction } from 'express';
+
+export default (req: Request, res: Response, next: NextFunction) => {
+  if (!['POST', 'DELETE', 'PUT', 'OPTIONS'].includes(req.method)) {
+    return next();
+  }
+  if (req.get('X-Requested-With') === 'XMLHttpRequest') {
+    return next();
+  }
+  res.status(401);
```
daneryl, line 10: 401 means Unauthorized, maybe look if there is a more fitting status code for this case ?

app/api/auth/headersMiddleware.ts
```diff
@@ -0,0 +1,12 @@
...
+  return res.json({ error: 'Unauthorized', message: 'X-Requested-With header was not found!' });
```
daneryl, line 11: maybe change the message to header was not sent on the request, to be more clear.
also match the error with what the new code (if changed) means

app/api/documents/specs/deprecatedRoutes.spec.js
```diff
@@ -12,6 +12,10 @@ jest.mock('../../utils/languageMiddleware.ts', () => (req, _res, next) => {
   next();
 });
 
+jest.mock('../../auth/headersMiddleware.ts', () => (_req, _res, next) => {
+  next();
+});
+
```
daneryl, line 18: why is this mock needed ? i think now the middleware is not being used on the test environment right ?

app/api/sync/specs/uploadRoute.spec.ts
```diff
@@ -26,6 +26,13 @@ jest.mock(
   }
 );
 
+jest.mock(
+  '../../auth/headersMiddleware.ts',
+  () => (_req: Request, _res: Response, next: NextFunction) => {
+    next();
+  }
+);
+
```
daneryl, line 35: again, is this needed ? can we avoid it somehow ?

daneryl (changes requested):
here is a quick code review, let me know what you think, and will check the functionality when this is ready.

app/api/documents/specs/deprecatedRoutes.spec.js
grafitto: Yes it is not used. Removed

app/api/sync/specs/uploadRoute.spec.ts
grafitto: Removed

app/api/auth/headersMiddleware.ts
grafitto: Is `400 Bad Request` okay?

app/api/api.js
grafitto: How about `CSRFMiddleware`. I am not sure whether that is good 😂

app/api/auth/headersMiddleware.ts
daneryl: i do not know, can you check what is that people is using in other apps when doing this kind of checks ? maybe 403 makes more sense ?

app/api/api.js
daneryl: yes, definitely better

app/api/auth/headersMiddleware.ts
grafitto: Will go with that then

app/api/auth/headersMiddleware.ts
grafitto: done

daneryl (approved):

PR #3917, opened 2021-09-15
Added storeKey document viewer

### LINKED ISSUE #3901: Image field popup not showing correct values in entity view (4 pts)
Image selection popup shows different values in library and PDF Document.

When you have an image selected from the supporting files:

Library:
Screenshot 2021-09-07 at 11 33 16

Document View:
Screenshot 2021-09-07 at 11 31 16
Screenshot 2021-09-07 at 11 31 11

Probably is because in the entity view we don't have the attachments, so the popup does not recognise the value as an attachment and shows the raw url.

grafitto:
fixes #3901 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Metadata/components/MetadataFormFields.js
```diff
@@ -356,8 +356,8 @@ export const mapStateToProps = (state, ownProps) => {
     attachments = selectedDocuments.size ? selectedDocuments.get(0).get('attachments') : undefined;
   }
 
-  if (storeKey === 'documentView') {
-    const entity = state.documentView.doc;
+  if (storeKey === 'documentView' || storeKey === 'documentViewer') {
```
fnocetti, line 359: A quick search shows that there are actually no uses of `documentView` in the application. Most probably this was a typo and not an unhandled case.

fnocetti (changes requested):
There are no tests to check that the image is actually shown. Maybe an e2e is worth exploring.

app/react/Metadata/components/MetadataFormFields.js
grafitto: Removed

e2e/suites/convert-entity-template.test.ts
```diff
@@ -0,0 +1,62 @@
...
+const setupTest = async () => {
+  await createTemplate('Without image');
+  await createTemplate('With image');
+  await createEntity('Without image', {
+    pdf: `${__dirname}/test_files/valid.pdf`,
+    supportingFile: `${__dirname}/test_files/batman.jpg`,
+  });
+};
+
+describe('Convert entity template', () => {
```
fnocetti, line 26: I don't think this is testing "Convert entity template" but actually testing that the image is being rendered. We need to find a better description and file name.

e2e/suites/convert-entity-template.test.ts
```diff
@@ -0,0 +1,62 @@
...
+        value: optionEl.getAttribute('value') as string,
+      }));
+      return value;
+    });
+
+    const optionValues: any[] = await Promise.all(optionsValues);
+    const { value } = optionValues.find(option => option.text === 'With image');
+
+    await expect(page).toMatchElement('select.form-control');
+    await page.select('select.form-control', value);
```
fnocetti, line 54: Wasn't it possible to use the template with image right from the beginning? I think it would serve the purpose and save a lot of code. Maybe I'm missing something :D

fnocetti (changes requested):
I left a couple comments. Please check.

e2e/suites/convert-entity-template.test.ts
grafitto: I also think that would help, the problem is I did not find any template that has an image out of all the ones I looked

fnocetti: Yeah, but wouldn't it be useful to create the entity directly with template "With image" rather than start with "Without image" and then changing it to "With image"? My concern is: does that template changing add anything to the test?

fnocetti (changes requested):
Please adjust the name of the test file as well. 'convert-entity-template' is missleading.

fnocetti:
Apart from the changes requested, the functionality works as expected.

e2e/suites/convert-entity-template.test.ts
```diff
@@ -0,0 +1,45 @@
+/*global page*/
+
+// import { ElementHandle } from 'puppeteer';
```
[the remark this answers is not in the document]
grafitto: Deleted

e2e/suites/convert-entity-template.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const setupTest = async () => {
+  await createTemplate('With image');
+  await createEntity('With image', {
+    pdf: `${__dirname}/test_files/valid.pdf`,
+    supportingFile: `${__dirname}/test_files/batman.jpg`,
+  });
+};
+
+describe('Image is rendered when switching entity template', () => {
```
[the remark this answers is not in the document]
grafitto: Updated description

fnocetti (approved):

PR #3922, opened 2021-09-17
Always assume multitenant

### LINKED ISSUE #3755: Error reported as localhost (5 pts)
The tenant needs to be displayed in the error for proper debugging:
```
2021-06-19T20:16:14.337Z [localhost] 
import.csv file not found
```

RafaPolit:
fixes #3755 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/log/specs/errorLog.spec.js
```diff
@@ -35,7 +37,7 @@ describe('errorLog', () => {
       tenants.add({ name: 'tenant' });
 
       await tenants.run(async () => {
-        anErrorLog.error('a message', { shouldBeMultiTenantContext: true });
+        anErrorLog.error('a message');
```
daneryl, line 40: the describe for this test is outdated now after removing the flag right ?

daneryl (changes requested):
@RafaPolit all looks good, just a minor comment on the code.

a question/change maybe ? while trying to test this i noticed that localhost was still being reported when an error occurs, which is the default behavior when the tenant is the default one,
i am not sure why we are doing this, the localhost is hard coded on the loggers to be reported as default, but basically gives no info.
we can remove this code and a couple of conditions and just always report the tenant name (which will be "default" instead of "localhost" for local development), what do you think ? another option is to report the database name instead of the tenant name if you want to disambiguate on single tenants uwazi environments ?

app/api/log/specs/errorLog.spec.js
RafaPolit: Yes, I have reverted the tests:
- default behavior now has no describe
- added describe for "unexpected" behavior

daneryl (approved):

PR #3927, opened 2021-09-21
3242 slow csv import with thesaurus creation

### LINKED ISSUE #3242: Import CSV slows down if creation of thesaurus values is needed [timebox 1 day]
Import CSV runs very fast when the row uses existing thesaurus values.

If the row needs to create one or more thesaurus values in different thesauri, it really slows down 10x or more.

Pre-computing the values to be added, adding them, and then moving on with the import could significantly improve import time.

This is probably also true for relationships.

LaszloKecskes:
fixes #3242

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/csv/importEntity.ts
```diff
@@ -66,6 +71,120 @@ type Options = {
...
+    if (
+      normalized &&
+      !thesauriIdToExistingValues.get(id).has(normalized) &&
+      !thesauriIdToNormalizedNewValues.get(id).has(normalized)
+    ) {
+      thesauriIdToNewValues.get(id)?.add(original);
+      thesauriIdToNormalizedNewValues.get(id).add(normalized);
+    }
+  }
+  await csv(await file.readStream(), errorContext?.stopOnError)
```
daneryl, line 140: @LaszloKecskes errorContext is confusing, its being used as error "emiter" and in this case its used to grab an option, and it actually its not clear what this object represents, it is the parent, csvLoader, i think its better if we do not pass the whole csvLoader, but instead the things we need here as parameters, like options for the stopOnError.

for the errors, can we make this handle its own errors and just throw them, the csvLoader can handle them in whatever manner.

this change basically means that this function should know nothing about the parent, throw errors in case it has and receive params like stopOnError, what do you think ?

app/api/csv/importEntity.ts
```diff
@@ -66,6 +71,120 @@ type Options = {
...
+  const result: { [k: string]: any } = {};
+  keys.forEach(k => {
+    if (input.hasOwnProperty(k)) {
+      result[k] = input[k];
+    }
+  });
+  return result;
+};
+
+const arrangeThesauri = async (
```
daneryl, line 84: @LaszloKecskes this function is really difficult to follow and understand, i have write a couple of comments, but i think its best if this is first refactored before proceeding with more code review (there is an issue on codeclimate pointing out this cognitive complexity).
some general things i think are important and can improve the complexity, readability:

- move this into a separate file, this clearly does not belong to importEntity as does another thing.
- try to maintain as less state as possible, there is 4 variables that you need to keep track when reading the function.
- split the function into smaller pieces with clear responsabilities
- throw errors instead of using the ```errorContext``` strategy (pointed out in another comment)

app/api/csv/importEntity.ts
```diff
@@ -1,18 +1,23 @@
+/* eslint-disable max-statements */
```
daneryl, line 1: this should only be allowed on old, files, never added unless there is a really good reason.

app/api/csv/specs/importEntity.spec.js
```diff
@@ -0,0 +1,141 @@
+import path from 'path';
```
daneryl, line 1: @LaszloKecskes is this spec really necessary ? i mean this development is a performance improvement right, the "result" of everything should be the same, are we doing new things that were not done before ?, if this is the case do you think we can add those those tests to any existing spec that tests the import itself (the whole module) like csvLoader.spec ?

app/api/csv/specs/csvLoader.spec.js
```diff
@@ -162,6 +169,22 @@ describe('csvLoader', () => {
...
+          thesauri2: 'thesauri2',
+        });
+        expect(tr.contexts.find(c => c.label === 'multi_select_thesaurus').values).toMatchObject({
+          multi_select_thesaurus: 'multi_select_thesaurus',
+          multivalue1: 'multivalue1',
+          multivalue2: 'multivalue2',
+          multivalue3: 'multivalue3',
+        });
+      });
+    });
```
daneryl, line 186: im not sure why this is needed, i believe thesauri save already saves/updates the translation context ?

daneryl (changes requested):

app/api/csv/importEntity.ts
LaszloKecskes: removed.

app/api/csv/importEntity.ts
```diff
@@ -66,6 +71,120 @@ type Options = {
...
+          nameToThesauriIdMultiselects[`${p.name}__${suffix}`] = thesarusID;
+        });
+      }
+    }
+  });
+  const allRelatedThesauri = await Promise.all(
+    Array.from(
+      new Set(thesauriRelatedProperties?.map(p => p.content?.toString()).filter(t => t))
+    ).map(async id => thesauri.getById(id))
+  );
```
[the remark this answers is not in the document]
LaszloKecskes: nice catch. done.

app/api/csv/importEntity.ts
LaszloKecskes: Done. I'm not confident about how the load function handles the errors thrown by the arrangeThesauri, we could probably doublecheck that.

app/api/csv/importEntity.ts
LaszloKecskes: Moved, and split up. As discussed, there's nothing I see about the state, I've tried to organize it more clearly.
It turns out the thesauri saving at the end was responsible for the most of it, as it's still a 7 over 5 on code climate. Can we pair that?

app/api/csv/specs/importEntity.spec.js
LaszloKecskes: Rewritten as discussed: to do the testing via CSVLoader.load, rather then directly the implementation. Still kept in a separate file, to not clutter the rest with unrelated checks.

app/api/csv/specs/csvLoader.spec.js
LaszloKecskes: Kept as it was a smoke test for an actual error that occurred. Moved out from this file to the other, thesauri related load tests.

app/api/csv/arrangeThesauri.ts
```diff
@@ -0,0 +1,180 @@
...
+};
+
+const syncSaveThesauri = async (
+  allRelatedThesauri: ThesaurusSchema[],
+  thesauriIdToNewValues: Map<string, Set<string>>
+) => {
+  for (let i = 0; i < allRelatedThesauri.length; i += 1) {
+    const thesaurus = allRelatedThesauri[i];
+    if (thesaurus?._id) {
+      const newValues: { label: string }[] = Array.from(
```
daneryl, line 111: @LaszloKecskes allRelatedThesauri has a ThesaurusSchema[] type, if you check what thesauri.get returns its actually a diferent type, which includes the _id, making the condition on line 110 redundant and ensuring you will always have an _id, there should be a better way to achieve this, but it has not been worked out yet.

app/api/csv/arrangeThesauri.ts
```diff
@@ -0,0 +1,180 @@
...
+  thesauriIdToNormalizedNewValues: Map<string, Set<string>>;
+};
+
+const setupIdValueMaps = (allRelatedThesauri: ThesaurusSchema[]): ThesauriValueData => {
+  const thesauriIdToExistingValues = new Map();
+  const thesauriIdToNewValues = new Map();
+  const thesauriIdToNormalizedNewValues = new Map();
+
+  allRelatedThesauri.forEach(t => {
+    if (t._id) {
```
daneryl, line 74: @LaszloKecskes same as a previous comment, thesauris.get already ensures the existence of _id, you just need to type the parameter properly with something like WithId<ThesaurusSchema>[]

app/api/csv/csvLoader.ts
```diff
@@ -56,7 +56,16 @@ export class CSVLoader extends EventEmitter {
...
+    try {
+      await arrangeThesauri(file, template, availableLanguages, this.stopOnError);
+    } catch (e) {
+      if (e instanceof ArrangeThesauriError) {
+        const _e: ArrangeThesauriError = e;
+        this.emit('loadError', _e.source, toSafeName(_e.row), _e.index);
+      } else {
+        throw e;
+      }
+    }
```
daneryl, line 68: @LaszloKecskes this whole try/catch block i believe is not tested, only the "no error" portion is. Also this is adding an eslint error for the current function having to many statements, maybe move the block into its own method?

app/api/csv/arrangeThesauri.ts
```diff
@@ -0,0 +1,180 @@
...
+  stopOnError: boolean = true
+) => {
+  const thesauriRelatedProperties = template.properties?.filter(p =>
+    ['select', 'multiselect'].includes(p.type)
+  );
+
+  let [
+    nameToThesauriIdSelects,
+    nameToThesauriIdMultiselects,
+  ] = separateSelectAndMultiselectThesauri(thesauriRelatedProperties, languages);
```
daneryl, line 139: @LaszloKecskes do you think we can avoid this separation and treat every property as the "same" this i think is possible and removes a lot of code.
- everything is a thesauri, it does not matter if its multi or not, you can parse it assuming always multiple values.

app/api/csv/arrangeThesauri.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  const thesauriValueData = setupIdValueMaps(allRelatedThesauri);
+
+  await csv(await file.readStream(), stopOnError)
+    .onRow(async (row: CSVRow, index: number) => {
+      if (index === 0) {
+        const columnnames = Object.keys(row);
+        nameToThesauriIdSelects = filterJSObject(nameToThesauriIdSelects, columnnames);
+        nameToThesauriIdMultiselects = filterJSObject(nameToThesauriIdMultiselects, columnnames);
+      }
```
daneryl, line 155: is this giving us that much performance ? can we remove it and executed on each iteration ?

app/api/csv/arrangeThesauri.ts
```diff
@@ -0,0 +1,180 @@
...
+      Object.entries(nameToThesauriIdSelects).forEach(([name, id]) => {
+        const label = row[name];
+        if (label) {
+          const normalizedLabel = normalizeThesaurusLabel(label);
+          handleLabels(id, label, normalizedLabel, thesauriValueData);
+        }
+      });
+      Object.entries(nameToThesauriIdMultiselects).forEach(([name, id]) => {
+        const labels = splitMultiselectLabels(row[name]);
+        if (labels) {
```
daneryl, line 165: `splitMultiselectLabels` is always guaranteed to return an object, the if is not needed here.

app/api/csv/arrangeThesauri.ts
```diff
@@ -0,0 +1,180 @@
...
+          Object.entries(labels).forEach(([normalizedLabel, originalLabel]) => {
+            handleLabels(id, originalLabel, normalizedLabel, thesauriValueData);
```
daneryl, line 167: `handleLables` its not clear what it does, i think something like `addNewLabel` is more descriptive of what the function does ?
also this function modifies a passed parameter and it makes it dificult to follow, consider modifying the valueData in here.

app/api/csv/arrangeThesauri.ts
```diff
@@ -0,0 +1,180 @@
...
+  allRelatedThesauri: ThesaurusSchema[],
+  thesauriIdToNewValues: Map<string, Set<string>>
+) => {
+  for (let i = 0; i < allRelatedThesauri.length; i += 1) {
+    const thesaurus = allRelatedThesauri[i];
+    if (thesaurus?._id) {
+      const newValues: { label: string }[] = Array.from(
+        thesauriIdToNewValues.get(thesaurus._id.toString()) || []
+      ).map(tval => ({ label: tval }));
+      if (newValues.length > 0) {
```
daneryl, line 114: since we are only saving thesaurus once, i do not think this gives us much performance gain, you can just save it and avoid a condition ?

app/api/csv/arrangeThesauri.ts
daneryl: `newValues` type should be inferred by the compiler for you, no need to explicitly type.

daneryl (changes requested):

app/api/csv/importEntity.ts
daneryl: ok, just reviewed it, i think its fine but needs to be tested and moved probably.

app/api/csv/importEntity.ts
daneryl: sure lets do it !

app/api/csv/arrangeThesauri.ts
LaszloKecskes: changed.

app/api/csv/arrangeThesauri.ts
LaszloKecskes: This is one of those "why didn't I think of that" moments. Of course you can do that.

app/api/csv/arrangeThesauri.ts
LaszloKecskes: Since we do not separate selects and multiselects, this thing is used only once, so the separate function is removed and the modification is done here.

app/api/csv/arrangeThesauri.ts
LaszloKecskes: Done.

app/api/csv/arrangeThesauri.ts
LaszloKecskes: Done.

app/api/csv/csvLoader.ts
LaszloKecskes: Removed during pairing.

app/api/csv/arrangeThesauri.ts
LaszloKecskes: Removed.

app/api/csv/arrangeThesauri.ts
LaszloKecskes: Moved condition to a filter prior to the loop.

daneryl:
LGTM

PR #3935, opened 2021-09-22
Upgrade mongoose

### LINKED ISSUE #3933: Mapth Critical vulnerability
Modules security check / yarn-audit has detected a critical vulnerability in mpath
| ⚠️ Critical | Type confusion |
| - | - |
| Package | mpath |
| Patched in | >=0.8.4 |
|Dependency of |mongoose |
|Path | mongoose > mpath|
| More info| [https://www.npmjs.com/advisories/1785](https://www.npmjs.com/advisories/1785)|

ℹ️ There was an attempt to update mongoose to version 5.13.9 since this version already has mpath 0.8.4, but some problems with the types appeared. In mongoose 5.13.9 [@types/mongoose](https://www.npmjs.com/package/@types/mongoose) is deprecated and should be removed, because mongoose already includes its own types, after that some type errors still happen.

RafaPolit:
fixes #3933

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/odm/ModelWithPermissions.ts
```diff
@@ -150,7 +150,11 @@ export class ModelWithPermissions<T> extends OdmModel<T> {
   async getById(id: any, select?: string) {
     const user = permissionsContext.getUserInContext();
     const doc = await this.db.findOne(
-      appendPermissionQuery({ _id: id || null }, AccessLevels.READ, user),
+      appendPermissionQuery(
+        { _id: id || null } as PermissionsUwaziFilterQuery<T>,
```
daneryl, line 154: @RafaPolit im pretty sure this is not needed, we should be able to pass the proper object. what is the reason ?, i think we typed wrongly the DataModelType?

app/api/odm/models.ts
```diff
@@ -8,12 +8,16 @@ export type WithId<T> = T & {
   _id: Schema.Types.ObjectId;
 };
 
-export type UwaziFilterQuery<T> = FilterQuery<T> & { _id?: any };
+export type DataModelType<T> = WithId<T> & Document;
```
daneryl, line 11: this DataModelType is probably not correct, we are using this on UwaziFilterQuery and in that case the _id can be a string, this will not work in that case, can you check if this is true ?, options:
- actually enforce the ObjectId type for _id on the whole application (this really has advantages, but is probably a longer and more difficult development)
- use a diferent WithId for the UpdateQuery with supports strings as the _id

daneryl: another issue with this type, there is another Data Type defined with the "old" definition in model.ts
https://github.com/huridocs/uwazi/blob/eec9203126096f24731f543f2bea8b33bb121541/app/api/odm/model.ts#L9 and its being used already in many places (meaning we are using now 2 different data types), we should remove the new one created and use the old one ? also move all this types to model.ts where i think make more sense ?

daneryl (changes requested):
@RafaPolit i think there is some confusion on the datamodel types, check the comments, let me know what you think.

app/api/odm/ModelWithPermissions.ts
RafaPolit: It is not needed with Typescirpt 4.3, it does not give any error.  It fails on 3.9.

app/api/odm/models.ts
RafaPolit: As for your first question, it apparently takes the string without any issue:
MongooseTypescript

RafaPolit: As for your second question, I am not really sure we want the DataType to always be a Partial?  The ReadOnly is probably more secure, but I'm not sure the implications.

RafaPolit:
Closing in favor of #3950

PR #3938, opened 2021-09-29
3852 - Loss of session in public forms of private instance with multi-tenant

### LINKED ISSUE #3852: public form does not work properly on a private instance. [Timebox]
to replicate:
- set up two private instance.
- ? sync them
- set up a remote public form one to another.
- use the form.
- session is lost.

the bug is that the session is lost every time you send the form, the form is sent succesfully.

mfacar:
fixes #3852

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/files/jsRoutes.js
```diff
@@ -127,6 +127,7 @@ const routes = app => {
       },
       proxyReqOptDecorator(proxyReqOpts) {
         const { tenant, ...headers } = proxyReqOpts.headers;
+        delete headers.cookie;
         return {
           ...proxyReqOpts,
           headers: { ...headers },
```
daneryl, line 133: @mercy can we instead of modifying headers, only delete it on the new returned object ?
`
headers: { ...headers, cookie: undefined or {} }
`
this way we avoid side effects.

another option is change line 129 to 
`
const { tenant, cookie, ...headers } = proxyReqOpts.headers;
`

app/api/files/specs/jsRoutes.spec.js
```diff
@@ -156,27 +156,44 @@ describe('upload routes', () => {
...
+      remoteServer = remoteApp.listen(54321, async () => {
+        const response = await request(app)
+          .post('/api/remotepublic')
+          .send({ title: 'Title' })
+          .set('tenant', 'tenant')
+          .expect(200);
+
+        const headersOnRemote = JSON.parse(response.text);
+        expect(headersOnRemote.tenant).not.toBeDefined();
+        done();
```
daneryl, line 179: @mfacar the description does not match what this is testing.

daneryl (changes requested):
@mfacar couple of comments, nice work on finding what was going on here

app/api/files/specs/jsRoutes.spec.js
mfacar: I unified with the previous test and updated the description

app/api/files/jsRoutes.js
mfacar: Sure thanks, I changed to the second option.

daneryl:
LGTM !

PR #3950, opened 2021-10-01
upgrade typescript packages

### LINKED ISSUE #3395: ci tasks run multiple times on internal PR

RafaPolit:
fixes #3395

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/odm/ModelWithPermissions.ts
```diff
@@ -115,14 +121,15 @@ const controlPermissionsData = <T>(data: T & { published?: boolean }, user?: Use
 };
 
 export class ModelWithPermissions<T> extends OdmModel<T> {
-  async save(data: DataType<T & { permissions?: PermissionSchema[] }>) {
+  async save(data: WithPermissionsDataType<T>) {
     const user = permissionsContext.getUserInContext();
+    const query = { _id: data._id } as PermissionsUwaziFilterQuery<T>;
```
RafaPolit, line 126: Not sure why I am required to do this `casts`.  In Typescript 4.x, they are not needed, in 3.x it does not accept this is an argument for the below function.  I am open to ideas as to what may be wrong.

app/api/migrations/migrations/19-rename-uploads-to-files/specs/19-rename-uploads-to-files.spec.js
```diff
@@ -43,14 +43,19 @@ describe('migration rename-uploads-to-files', () => {
   });
 
   describe('when files already exists', () => {
+    beforeEach(async () => {
+      const collections = await testingDB.mongodb.listCollections().toArray();
+      if (!collections.find(c => c.name === 'files')) {
+        await testingDB.mongodb.createCollection('files');
+      }
+    });
+
```
RafaPolit, line 52: New mongo driver fails on `createCollection` of existing collections, so this prior check is now needed.

app/api/odm/ModelWithPermissions.ts
```diff
@@ -115,14 +120,15 @@ const controlPermissionsData = <T>(data: T & { published?: boolean }, user?: Use
 };
 
 export class ModelWithPermissions<T> extends OdmModel<T> {
-  async save(data: DataType<T & { permissions?: PermissionSchema[] }>) {
```
RafaPolit, line 118: I believe this was wrong to start with, the `& permissions` should have been OUTSIDE of the <T> bracket.

app/api/odm/MultiTenantMongooseModel.ts
```diff
@@ -1,10 +1,16 @@
...
+  UwaziUpdateQuery,
+  UwaziQueryOptions,
+  PartialDataType,
+} from './model';
 import { tenants } from '../tenants/tenantContext';
 import { DB } from './DB';
 
 class MultiTenantMongooseModel<T> {
-  dbs: { [k: string]: mongoose.Model<WithId<T> & mongoose.Document> };
+  dbs: { [k: string]: mongoose.Model<DataType<T>> };
```
RafaPolit, line 13: In the first iteration, `DataType` included an `& Document`.  This is no longer true.  This means that `dbs` no longer explicitly has the `mongoose.Document` type extended, but I think that `T` should already have that passed along on instance.

app/api/odm/MultiTenantMongooseModel.ts
```diff
@@ -20,15 +26,16 @@ class MultiTenantMongooseModel<T> {
...
+        this.collectionName,
+        this.schema
+      );
     }
 
     return this.dbs[currentTenant.name];
   }
 
-  findById(id: any | string | number, select?: any) {
+  findById(id: any, select?: any) {
```
RafaPolit, line 38: I was confused that this had `any | string | number`... `any` is already any.

app/api/odm/MultiTenantMongooseModel.ts
```diff
@@ -38,20 +45,20 @@ class MultiTenantMongooseModel<T> {
...
-    update: Readonly<Partial<T>> & { _id?: any },
-    options: any = {}
+    update: UwaziUpdateQuery<T>,
+    options: UwaziQueryOptions
   ) {
     return this.dbForCurrentTenant().findOneAndUpdate(query, update, options);
   }
 
-  async create(data: Readonly<Partial<T>> & { _id?: any }) {
+  async create(data: PartialDataType<T>) {
```
RafaPolit, line 54: Nothing in the new implementation has `Readonly`.  Any hint as to why this could have been needed?  I just removed it from everywhere and nothing is failing.  I'm open to being enlightened about this as well.

app/api/odm/specs/logHelper.spec.ts
```diff
@@ -19,6 +19,7 @@ describe('NoLogger', () => {
   it('should return empty promises for logging methods', async () => {
     const logHelper = createUpdateLogHelper('activitylog');
     await expect(logHelper.getAffectedIds(undefined)).resolves.toBeEmpty;
+    // @ts-ignore
     await expect(logHelper.upsertLogMany({ _id: null })).resolves.toBeEmpty;
```
RafaPolit, line 23: This is testing a value that is not valid as _id.  This type of errors should, in theory, be caught by TS from now on, so this test scenario should ideally not arise.  Unsure if this merits a deletion of both expects (this and the next one, which is more or less the same situation), but I'm keeping them as I am unsure of the scenario this was written in.

app/api/odm/ModelWithPermissions.ts
daneryl: I am not sure how to solve but what i think is wrong is:
this method save uses `data: WithPermissionsDataType<T>` while appendPermissionsQuery uses as the first param `query: PermissionsUwaziFilterQuery<T>`
the first is based on `PartialDataType`, which means, _id can be not defined, the second is based on `DataType` which means _id can NOT be undefined. i think despite the difficult to read error this is the issue.

app/api/odm/ModelWithPermissions.ts
daneryl: im not sure this is the case with the current implementation, lets take a look together.

app/api/odm/MultiTenantMongooseModel.ts
daneryl: not sure, @fnocetti @mfacar ?

mfacar: I don't think it's needed for the permissions model 🤔

PR #3995, opened 2021-10-28
Fix failing screenshot tests on mac

### LINKED ISSUE #3912: Visual regression tests on key screens of the application (8 pts)
To detect visual bugs we want to add image snapshots on key steps of the e2e. Currently mac users have a problem with this type of snapshot expects because the resolution of the image does not match.

We will be adding them to the puppeteer suites.

Tasks:
- Fix problem for mac users
- Add an e2e to cover the majority of the app.
- Add image snapshots on suites that have complex ⚠️ **key steps**

⚠️ ISSUE UPDATE ⚠️
- Develop it only for linux, do an spike to check that works on Linux, CI and different distributions.
- Should be a separate suite
- Move the test that fails on mac to the new suit

Previous attempt is in the branch `3912-adds-more-visual-regression-tests`

Edited by konz

grafitto:
fixes #3912

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


e2e/suites/graphs.test.ts
```diff
@@ -64,15 +74,16 @@ describe('Graphs in Page ', () => {
...
-    it('should display Bar chart graph in page with no more than a 3% difference', async () => {
+    it('should display Bar chart graph in page with no more than a 7% difference', async () => {
       const graphsPage = await displayGraph();
-      const chartContainer = ensure<ElementHandle>(
-        await graphsPage.$('.recharts-responsive-container')
+      const chartScreenshot = await getChartContainerScreenshot(
+        graphsPage,
+        '.recharts-responsive-container'
       );
+      const resizedChart = await resizeImage(chartScreenshot, 1000, 320);
```
fnocetti, line 83: I don't think this approach will escalate. We need to find some way to avoid passing height and width every time. Maybe resize always to the same width while keeping the ration for the height?
cc. @daneryl

fnocetti (commented):
Please take a look at my comment.
Also, I see we never set the viewport for the puppeteer tests (https://github.com/puppeteer/puppeteer/blob/v4.0.0/docs/api.md#pagesetviewportviewport). Maybe that's the reason behind the size miss-matching?

grafitto:
> Please take a look at my comment. Also, I see we never set the viewport for the puppeteer tests (https://github.com/puppeteer/puppeteer/blob/v4.0.0/docs/api.md#pagesetviewportviewport). Maybe that's the reason behind the size miss-matching?

I have tried setting the viewport, was still behaving the same

e2e/suites/graphs.test.ts
grafitto: Done

package.json
```diff
@@ -175,6 +175,7 @@
     "rtlcss": "^2.6.0",
     "sanitize-filename": "^1.6.3",
     "serialize-javascript": "^5.0.1",
+    "sharp": "^0.29.2",
```
fnocetti, line 178: Can you please install sharp as a devDependency instead, please?

grafitto: Ooh yea

grafitto: Done

fnocetti (approved):
Nice job! Merging now

PR #4005, opened 2021-11-09
[Hotfix] Translated "Select..." option in dropdowns

### LINKED ISSUE #4000: [Hotfix] The placeholder "Select..." is not translated (1 pts)
The form component [Select](https://github.com/huridocs/uwazi/blob/development/app/react/Forms/components/Select.js) has a placeholder that is not translated. By default it's 'Select...' but this is not translated and it can also receive a different value, like for example is used in the [Templates config](https://github.com/huridocs/uwazi/blob/421a01d699f057d3c8e647838dab2f4f8b3c9c14/app/react/Templates/components/FormConfigRelationship.js#L93) to pass "Any entity or document"

- Make it translatable.

Screenshot 2021-11-08 at 16 36 50

grafitto:
fixes #4000 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Forms/components/Select.js
+++ b/app/react/Forms/components/Select.js
@@ -3,7 +3,7 @@
 import { advancedSort } from 'app/utils/advancedSort';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
-
+import { t } from 'app/I18N';
 export default class Select extends Component {
   render() {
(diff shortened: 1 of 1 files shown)


konzz:
@grafitto this is a Hotfix, so it should point to master

PR #4017, opened 2021-11-12
Emit socket event

daneryl:
- small cleanup of root folder
- client can receive a socket event "forceReconnect" which will reconnect and retry in case the connection fails.
- yarn emitSocketEvent to send events to clients, specific tenant or all (requires cluster mode with redis)
execute ```yarn emitSocketEvent``` to see help and options, in order for this to work uwazi needs to run in cluster_mode ```CLUSTER_MODE=true yarn hot``` and redis running


app/api/socketio/standaloneEmitSocketEvent.ts
```diff
@@ -0,0 +1,23 @@
...
+const emitSocketEvent = async (tenant: string, event: string, data: string) =>
+  new Promise((resolve, reject) => {
+    const redisClient = createClient({ host: config.redis.host, port: config.redis.port });
+
+    redisClient.on('error', reject);
+    redisClient.on('end', resolve);
+
+    redisClient.on('ready', () => {
+      const io = new Emitter(redisClient);
+      if (tenant === 'all') {
```
RafaPolit, line 14: This is susceptible to collide with a tenant named `all`: we have discussed and passing an INTENTIONAL empty string for `all` could be a better approach.

RafaPolit (changes requested):
Apart from the minor comment, LGTM!

app/api/socketio/standaloneEmitSocketEvent.ts
daneryl: done

PR #4025, opened 2021-11-12
3958 - Added check mapping logic to template creation 

### LINKED ISSUE #3958: Check mapping logic in template creation is controlled by the front-end leaving the API obscure (8 pts)
When calling to api/templates for template creation, the template mapping will not be automatically created unless an explicit call to another end-point (/api/templates/check_mapping) is performed first. This particularity is obscure and undocumented.

Zasa-san:
fixes #3958

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/templates/specs/routes.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+  name: 'template4',
+  properties: [],
+  commonProperties: templateCommonProperties,
+};
+
+const emitToCurrentTenantSpy = jasmine.createSpy('emitToCurrentTenant');
+
+describe('templates routes', () => {
+  const app: Application = setUpApp(templateRoutes, (req, _res, next: NextFunction) => {
+    (req as any).user = {
```
konzz, line 35: This is a bit confusing, the authMiddleware is mocked, but also the testingEnvironment has it's own function to setup the user, is this a legacy approach left there when migrating to TypeScript?

Zasa-san: It's a mistake, I started by setting up the user, then added the middleware mock, but forgot to remove this. Removed now.

PR #4042, opened 2021-11-20
Information extraction

### LINKED ISSUE #3800: Suggestions Review Panel
For the Feature Extraction we need a panel to review  the suggestions. In this section we will see the suggestions for a specific template / property.

- This suggestions will come from mongo. (Different issue will integrate with the service)
- Pagination TBD
- Order TBD
- "Accept" action will put the suggestion in to the property and mark the suggestion as accepted.
- "Mark" action will be implemented later.
- Dashboard button will return to the #3799
- Review button in the dashboard should open this section
- Once a suggestion is accepted, the accept button should disappear

The possible status are:
- Accepted: when the suggestion has been accepted
- Empty: When the property has no value
- Pending: The property has a value and the suggestion has not been accepted.

Link to the [designs](https://www.figma.com/file/h3mhxBGiRctmF72HLf0qld/UWAZI---ML---information-extraction?node-id=1270%3A2824)

Screenshot 2021-07-08 at 12 02 47

### LINKED ISSUE #3929: Accept entity suggestion (8 pts)
⚠️ Depends on #3928 and #3800

In the suggestions review panel add the accept button and make it update the suggested value in to the entity property and update the suggestion status.
See #3800 for all the details.

mfacar:
fixes #3800
fixes #3929

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/socketio/setupSockets.ts
```diff
@@ -24,8 +24,16 @@ declare global {
   }
 }
 
+let io: SocketIoServer;
+const emitToTenant = (tenantName: string, event: string, ...data: any[]) => {
```
konzz, line 28: @fnocetti This is the method for emiting to tenants

app/api/migrations/migrations/58-add_system_key_translations/index.js
```diff
@@ -32,60 +38,24 @@ async function insertSystemKeys(db, newKeys) {
...
-        key: 'Find suggestions',
-        value: 'Find suggestions',
-      },
-      {
-        key: 'per page',
-        value: 'per page',
-      },
-    ];
-    await insertSystemKeys(db, systemKeys);
+
```
LaszloKecskes, line 55: What was the issue with the non-file based approach? I sort of like that better, since it avoids importing the csv facilities.

app/api/services/tasksmanager/specs/ExternalDummyService.ts
```diff
@@ -18,45 +18,65 @@ export class ExternalDummyService {
...
+  results: any | undefined;
 
   redisClient: RedisClient | undefined;
 
   fileResults: string | undefined;
 
   private readonly serviceName: string;
 
-  constructor(port = 1234, serviceName = 'dummy') {
+  materialsDataPartams: any;
```
LaszloKecskes, line 35: Typos in the variable names (lines 35-41).

app/react/MetadataExtraction/SuggestionsAPI.ts
```diff
@@ -1,14 +1,27 @@
...
-    resolve({
-      suggestions: data,
-      totalPages: 30,
-    });
-  });
-
-export { getSuggestions };
+import api from 'app/utils/api';
+
+const getSuggestions = async (requestParams: RequestParams) => {
```
LaszloKecskes, line 4: For everything in this file: usually, our *API definitions in the front end go like this (example from PagesAPI).
```
  save(requestParams) {
    return api.post('pages', requestParams).then(response => response.json);
  },
```

From style perspective I don't think the difference is important, but if I read it correctly, the example returns a Promise and these do not. Is returning a Promise something we would like to keep to in our frontent API definitions?

app/react/MetadataExtraction/specs/SugestionsAPI.spec.ts
```diff
@@ -0,0 +1,33 @@
+import api from 'app/utils/api';
```
LaszloKecskes, line 1: Typo in the filename.

app/react/MetadataExtraction/specs/SugestionsAPI.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  describe('status', () => {
+    it('should return the status of the training', async () => {
+      const request = new RequestParams();
+      spyOn(api, 'get').and.returnValue(Promise.resolve({ json: 'data' }));
+      const result = await ixStatus(request);
+      expect(result).toEqual('data');
+    });
+  });
```
LaszloKecskes, line 32: SuggestionsAPI.acceptEntitySuggestion does not need to be tested?

app/setUpJestServer.js
```diff
@@ -1,3 +1,4 @@
+process.env.EXTERNAL_SERVICES = true;
```
LaszloKecskes, line 1: I guess the goal is to always have this true for tests. Just out of curiosity: is this a common pattern in JS (i.e. setting environment variables in code)? 
Also, can this go after the import? My eslint complains (with just a warning).

package.json
```diff
@@ -213,7 +213,7 @@
     "@babel/preset-react": "7.10.4",
     "@babel/preset-typescript": "7.10.4",
     "@babel/traverse": "^7.15.0",
-    "@testing-library/jest-dom": "^5.14.1",
+    "@testing-library/jest-dom": "^5.16.1",
```
LaszloKecskes, line 216: Just to double-check: did we discuss that it is okay to upgrade this here?

scripts/checkServicesTasks.js
```diff
@@ -0,0 +1,122 @@
...
+  .option('deleteOne', {
+    alias: 'd',
+    type: 'string',
+    describe: 'ID of the message to be deleted',
+  })
+
+  .demandOption(['service'], '\n\n').argv;
+
+const run = async () => {
+  const redisClient = await Redis.createClient(redisUrl);
```
LaszloKecskes, line 43: While there are these in other places of the code as well, my typescript tells me that these two awaits may be not necessary.

scripts/checkServicesTasks.js
```diff
@@ -0,0 +1,122 @@
...
+  const deleteMessage = async id => {
+    await redisSMQ.deleteMessageAsync({
+      qname: `${service}_tasks`,
+      id,
+    });
+  };
+
+  const readAllTaskMessages = async () => {
+    const messages = [];
+    while (true) {
```
LaszloKecskes, line 68: how about something like:
```
let message = await readFirstTaskMessage();
while(message){
messages.push(message);
message = await readFirstTaskMessage();
}
```
This avoids the constant in the loop warning from eslint.

scripts/checkServicesTasks.js
```diff
@@ -0,0 +1,122 @@
...
+      id: message.id,
+      task: messageData.task,
+      tenant: messageData.tenant,
+      params: messageData.params,
+      sent: new Date(message.sent).toLocaleString(),
+    };
+  };
+
+  if (deleteOne) {
+    console.log(`💣 \u001b[31m DELETING message \u001b[32m${deleteOne}\u001b[37m`);
```
LaszloKecskes, line 93: Eslint unexpected console statement warning.
In other scripts we seem to use `process.stdout.write()` and `process.stderr.write()`, whichever is needed.

app/api/migrations/migrations/58-add_system_key_translations/index.js
mfacar: I changed it because since it increased the number of translations, the file ended to be so long

app/react/MetadataExtraction/SuggestionsAPI.ts
mfacar: I'm not sure about this point. I think we don't have a convention for that, let's talk about it in the meeting

app/react/MetadataExtraction/specs/SugestionsAPI.spec.ts
mfacar: Of course yes, I'll add the test

package.json
mfacar: This was not discussed because this dependency was added to test React Table, is not present in development, the upgrade is fine.

app/react/MetadataExtraction/SuggestionsAPI.ts
mfacar: Discussed in meeting, both ways are valid.

app/api/services/tasksmanager/specs/ExternalDummyService.ts
mfacar: done

app/react/MetadataExtraction/specs/SugestionsAPI.spec.ts
mfacar: done

scripts/checkServicesTasks.js
mfacar: right, done

app/setUpJestServer.js
mfacar: I think it's fine for the purposes of the test. I moved the line.

app/api/migrations/migrations/58-add_system_key_translations/index.js
LaszloKecskes: Make sense; same reason we originally did it.

app/react/MetadataExtraction/SuggestionsAPI.ts
LaszloKecskes: Yes, I have misread it. Got to remember that async's return Promises automatically. :)

app/react/MetadataExtraction/specs/SugestionsAPI.spec.ts
LaszloKecskes: Checked, done.

app/api/suggestions/routes.ts
```diff
@@ -0,0 +1,124 @@
...
+      res.json(suggestionsList);
+    }
+  );
+
+  app.post(
+    '/api/suggestions/train',
+    serviceMiddleware,
+    needsAuthorization(['admin']),
+    validateAndCoerceRequest({
+      properties: {
```
LaszloKecskes, line 42: Same as starting on line 69, maybe refactor in this file.

app/api/suggestions/routes.ts
```diff
@@ -0,0 +1,124 @@
...
+          additionalProperties: false,
+          required: ['property'],
+          properties: {
+            property: { type: 'string' },
+          },
+        },
+      },
+    }),
+    async (req, res, _next) => {
+      if (!IX) {
```
LaszloKecskes, line 53: Same as starting on line 80, maybe refactor the service availability checking and error message into a function in this file.

app/api/suggestions/suggestions.ts
```diff
@@ -0,0 +1,164 @@
...
+  ) => {
+    const suggestion = await IXSuggestionsModel.getById(acceptedSuggestion._id);
+    if (!suggestion) {
+      throw new Error('Suggestion not found');
+    }
+
+    const query = allLanguages
+      ? { sharedId: acceptedSuggestion.sharedId }
+      : { sharedId: acceptedSuggestion.sharedId, _id: acceptedSuggestion.entityId };
+    const storedEntities = await entities.get(query, '+permissions');
```
LaszloKecskes, line 143: This is more of a question: we cannot use entities.multipleUpdate here, because that one updates the current language of multiple sharedId's, where here we want to go along "other axis", i.e. update one or more languages for the same sharedId?

app/api/suggestions/suggestions.ts
```diff
@@ -0,0 +1,164 @@
...
+        permissions: entity.permissions || [],
+      }));
+      await entities.saveMultiple(entitiesToUpdate);
+    } else {
+      const entitiesToUpdate = storedEntities.map((entity: EntitySchema) => ({
+        ...entity,
+        title: suggestion.suggestedValue,
+      }));
+      await Promise.all(
+        entitiesToUpdate.map(async (entity: EntitySchema) => entities.save(entity, params))
```
LaszloKecskes, line 160: Can't we send this one through entities.saveMultiple as well?

app/api/suggestions/suggestions.ts
```diff
@@ -0,0 +1,164 @@
...
+            segment: 1,
+            currentValue: 1,
+            state: 1,
+            page: 1,
+          },
+        },
+      ],
+      {
+        stage1: [{ $group: { _id: null, count: { $sum: 1 } } }],
+        stage2: [{ $skip: offset }, { $limit: limit }],
```
LaszloKecskes, line 117: I don't have a better solution at the moment, but this request could become a potential performance bottleneck for a large amount of entities (i.e. it needs to go through all the data for every page).

app/api/suggestions/routes.ts
mfacar: Actually, it's different, the first one is expecting a value into the body while the second one is expecting the value into the query. I extracted it into a function that receives this root option.

app/api/suggestions/routes.ts
mfacar: done

app/api/suggestions/suggestions.ts
mfacar: Yes, that's part of the reason. The main reason however is that the way multipleUpdate/entity.save work, they were overwriting the changes across the different languages. Additionally, it does not merit going into the details of the saving because we don't need here all the complexity that entity save has.

app/api/suggestions/suggestions.ts
mfacar: Totally, it's done.

app/api/suggestions/suggestions.ts
mfacar: Probably it could be, we need to do some tests with a large number of entities, even for the training phase.  Since it's against the database directly the indexes should help in order to speed the query, probably we'll need additional ones depending on the load tests.

app/api/suggestions/suggestions.ts
LaszloKecskes: Got it.

LaszloKecskes: Got it.

app/api/suggestions/suggestions.ts
LaszloKecskes: Agreed. Let's handle this as a tech debt, since I don't think it should hold up the rest of the pr, if you agree. I have meant this more as a conversation opener.

app/api/suggestions/specs/routes.spec.ts
```diff
@@ -0,0 +1,260 @@
...
+
+  afterAll(async () => testingEnvironment.tearDown());
+
+  describe('GET /api/suggestions', () => {
+    it('should return the suggestions filtered by the request language and the property name', async () => {
+      const response = await request(app)
+        .get('/api/suggestions')
+        .query({ filter: { propertyName: 'super_powers' } })
+        .expect(200);
+      expect(response.body.suggestions).toMatchObject([
```
LaszloKecskes, line 65: How does the "filtered by language work"? We mock an English request, but we still expect the Spanish entity returned as well?

app/api/services/informationextraction/InformationExtraction.ts
```diff
@@ -0,0 +1,448 @@
...
+      language: languages.get(file.language!, 'ISO639_1'),
+    });
+
+    return entity;
+  };
+
+  coerceSuggestionValue = (suggestion: RawSuggestion, templates: TemplateSchema[]) => {
+    const allProps: PropertySchema[] = [];
+
+    templates.forEach(template => {
```
LaszloKecskes, line 152: Maybe something like:
const allProps = template.map(t=>t.properties).filter(Boolean).flat()

app/api/services/informationextraction/InformationExtraction.ts
```diff
@@ -0,0 +1,448 @@
...
+      propertyName: property,
+      status: 'processing',
+    });
+
+    const [suggestion] = await IXSuggestionsModel.get({
+      propertyName: property,
+      status: 'processing',
+    });
+
+    if (currentModel) {
```
LaszloKecskes, line 372: Maybe move this between the two requests above, since we don't need `suggestion` if we return from this first "if".

app/api/services/informationextraction/InformationExtraction.ts
```diff
@@ -0,0 +1,448 @@
...
+      status: 'processing',
+      creationDate: new Date().getTime(),
+      propertyName: property,
+    });
+  };
+
+  getTemplatesWithProperty = async (property: string) => {
+    const settingsValues = await settings.get();
+    const templates: ObjectIdSchema[] = [];
+    settingsValues.features?.metadataExtraction?.templates?.forEach(t => {
```
LaszloKecskes, line 408: Maybe something like:
`const templates = settingsValues.features?.metadataExtraction?.templates?.filter(t=>t.properties.includes(property)).map(t=>new ObjectId(t.template))`

app/react/MetadataExtraction/EntitySuggestions.tsx
```diff
@@ -1,128 +1,65 @@
 /* eslint-disable react/no-multi-comp */
```
LaszloKecskes, line 1: There's another one (SuggestionsTable) but I'm asking only here: since the adherence to eslint rules has been a topic, do we want to keep this one? In these particular cases, if I see it correctly, the culprits are the cell definitions for the tables (actionCell, suggestionCell), where it makes sense to refactor them out, and hopefully not too much effort.

app/api/suggestions/specs/routes.spec.ts
mfacar: The returned is actually the title from the English entity, but the language of the suggestion is in Spanish, we are not filtering the suggestions by their language but by the language of entities. I added the language to the title to make it clear.

app/api/services/informationextraction/InformationExtraction.ts
mfacar: done

app/api/services/informationextraction/InformationExtraction.ts
mfacar: done

app/api/services/informationextraction/InformationExtraction.ts
mfacar: done

app/react/MetadataExtraction/EntitySuggestions.tsx
mfacar: There are some complications moving the functions outside these two files.
In this one there's a dependency with the function `showConfirmationModal` that is not possible to pass directly to the cell function given it has a specific definition, we could wrap it in another function but I think it could add unnecessary complexity.
In `SuggestionsTable` every Header and Cell is also pretty small functional components, so I think it merits disabling the rule here also.
Indeed the team agreed in the past that for functional components like these it's ok to disable this rule.

LaszloKecskes: Got it!

LaszloKecskes:
> @LaszloKecskes thanks, the two issues about entity deletion are solved.

Checked, works. Nice.

LaszloKecskes:
There is an issue with training the same entities with a different property. Steps:

1. Create any template with at least two properties for extraction (one can be the title). Create the data (2-3 entities is enough). Create the labeled data for at least one.
2. Pick a property. Extract. The process should finish normally, one should see the suggestions on the dashboard.
3. Now, pick a different property. Extract the information. The suggestions do not show up.

What I could figure out is that somehow the existing suggestions seem to block the process. That is:
- If the suggestions are manually deleted from the db after step 2 (the first extraction), the second one can be then completed normally.
- If between the first and second extraction a new entity is introduced with files, then the second extraction produces result for only that.

@konzz : @mfacar has confirmed that she sees the issue on her setup as well.

Also, an additional, possible related issue is that for me, any property other than title (i.e. metadata), does not produce suggestions for entities without labeled data, only for entities where I went through the click-to-fill process on the pdf.

LaszloKecskes:
> Found an error with Lazslo: The information extraction checks for entities that have segmentation in order to send them to information extraction, but does not check for the status, so it happend that the information extraction tried to send an entity that was still in processsing for the segmentation and throws an error.

This is also fixed.

LaszloKecskes:
This is more of a question: so far I have been doing the labeling through the click-to-fill function, in the entity view with a pdf. For me, for selects, this is not available, therefore for selects the extraction fails through "no labeled data" error. Are there other flows for labeling data, or should I be able to do this for selects?

mfacar:
> Also, an additional, possible related issue is that for me, any property other than title (i.e. metadata), does not produce suggestions for entities without labeled data, only for entities where I went through the click-to-fill process on the pdf. @mfacar can you check whether you see this as well?

yes, @LaszloKecskes, it's the same for me

> Also, the dates do not work for me at all. Similarly, as all in the above cases, the "missing" suggestions do show up in the db, but are permanently stuck in a "processing" state. There are no tasks in the redis queues.

I also tested numbers and dates without good predictions, I think we need to check that with @gabriel-piles

LaszloKecskes:
> There is an issue with training the same entities with a different property. Steps:
> 
> 1. Create any template with at least two properties for extraction (one can be the title). Create the data (2-3 entities is enough). Create the labeled data for at least one.
> 2. Pick a property. Extract. The process should finish normally, one should see the suggestions on the dashboard.
> 3. Now, pick a different property. Extract the information. The suggestions do not show up.
> 
> What I could figure out is that somehow the existing suggestions seem to block the process. That is:
> 
> * If the suggestions are manually deleted from the db after step 2 (the first extraction), the second one can be then completed normally.
> * If between the first and second extraction a new entity is introduced with files, then the second extraction produces result for only that.
> 
> @konzz : @mfacar has confirmed that she sees the issue on her setup as well.
> 
> Also, an additional, possible related issue is that for me, any property other than title (i.e. metadata), does not produce suggestions for entities without labeled data, only for entities where I went through the click-to-fill process on the pdf. @mfacar can you check whether you see this as well?

@mfacar: Konz's previous commit seems to have fixed the second issue, and we have paired a fix for the first one this morning. Again, can you doublecheck that these work for you as well?

mfacar:
@LaszloKecskes unfortunately I couldn't set up my environment locally to test the latest changes, however, we identified with @gabriel-piles an error when the file language is 'other', I added a change to control that and also the page property expected from the updated pdf-information-extraction service.

LaszloKecskes:
Thanks, I've checked the changes in the morning. Let me approve it.

LaszloKecskes:
As discussed on the call: the error messages got do not get stored in the database as error, the suggestions remain there as 'pending'.

PR #4049, opened 2021-11-24
3934 - Geolocation search query only in map view

### LINKED ISSUE #3934: Geolocation search query is very slow [timeboxed 1 day]
Debug geolocation bottleneck, that apparently happens also when the number of geolocation points is 0.

Zasa-san:
fixes #3934

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/App/scss/elements/_map.scss
+++ b/app/react/App/scss/elements/_map.scss
@@ -549,11 +549,9 @@ a.mapboxgl-ctrl-logo {
 }
 
 @media print {
-  /* stylelint-disable */
   .mapbox-improve-map {
     display: none;
   }
-  /* stylelint-enable */
 }
 
 .library-map {
--- a/app/react/App/scss/elements/_map.scss
+++ b/app/react/App/scss/elements/_map.scss
@@ -548,8 +548,27 @@ a.mapboxgl-ctrl-logo {
   @include tooltip('right');
 }
 
+@media print {
+  /* stylelint-disable */
+  .mapbox-improve-map {
+    display: none;
+  }
+  /* stylelint-enable */
+}
+
 .library-map {
   overflow: hidden;
+
+  .documents-counter {
+    & > span {
+      padding: 5px;
+      background-color: #fff;
+      border-color: #ccc;
+      border-style: solid;
+      border-width: 1px;
+      border-radius: 10px;
+    }
+  }
 }
 
 .mapbox-navigation {
--- a/app/react/Library/LibraryMap.js
+++ b/app/react/Library/LibraryMap.js
@@ -7,7 +7,7 @@ import LibraryModeToggleButtons from 'app/Library/components/LibraryModeToggleBu
 
 export default class LibraryMap extends Library {
   static async requestState(requestParams, globalResources) {
-    return requestState(requestParams, globalResources, { mapMarkers: true });
+    return requestState(requestParams, globalResources, { geolocation: true });
   }
 
   render() {
--- a/app/react/Library/LibraryMap.js
+++ b/app/react/Library/LibraryMap.js
@@ -7,7 +7,7 @@ import LibraryModeToggleButtons from 'app/Library/components/LibraryModeToggleBu
 
 export default class LibraryMap extends Library {
   static async requestState(requestParams, globalResources) {
-    return requestState(requestParams, globalResources, 'markers');
+    return requestState(requestParams, globalResources, { mapMarkers: true });
   }
 
   render() {
--- a/app/react/Library/LibraryMap.js
+++ b/app/react/Library/LibraryMap.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import requestState from 'app/Library/helpers/requestState';
+import { requestState } from 'app/Library/helpers/requestState';
 import MapView from 'app/Library/components/MapView';
 import LibraryLayout from 'app/Library/LibraryLayout';
 import Library from 'app/Library/Library';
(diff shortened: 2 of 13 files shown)


app/react/Library/LibraryMap.js
```diff
@@ -7,7 +7,7 @@ import LibraryModeToggleButtons from 'app/Library/components/LibraryModeToggleBu
 
 export default class LibraryMap extends Library {
   static async requestState(requestParams, globalResources) {
-    return requestState(requestParams, globalResources, 'markers');
```
RafaPolit, line 10: So, was this actually passing `markers` as the value for `calculateTableColumns`? Something was wrong here to begin with.  Did you trace this problem when you debugged this issue?

app/react/Library/helpers/requestState.js
```diff
@@ -50,67 +50,60 @@ export function processQuery(params, globalResources, key) {
...
-        templates,
-        globalResources.thesauris.toJS(),
-        globalResources.relationTypes.toJS(),
-        request.data.quickLabelThesaurus
-          ? getThesaurusPropertyNames(request.data.quickLabelThesaurus, templates)
-          : []
-      );
+  if (options.geolocation) {
+    documentsRequest = request.set({ ...docsQuery, geolocation: true });
+  }
```
RafaPolit, line 66: I am confused by this: the previous behavior was that, any query, be it Cards or Map or Table, the library -> filters -> documentTypes were the result of `request.set` with the `tocGenerationUtils.aggregations`.  Here, you are changing that behavior for the map view when geolocation is asked.   Please, ensure this does not have adverse effects on the filters and aggregations on the side panel.

RafaPolit: Indeed, this is changing the values of aggregations when you change to map view, I don't think that is the intended behavior.  The aggregations on the library, in whichever "mode", be it Cards, Map or Table View, should show the same aggregations.  The map should show all the map geolocation values, but the aggregation numbers should not change depending on the tab I am in.  They are just different views of the same collection.

app/react/Library/LibraryMap.js
Zasa-san: I think the `markers` parameter is legacy code that passed unnoticed since the files involved were not typed. This `markers ` parameter was passed to the `processQuery` function in `requestState` to set geolocation to true or false. Process query no longer handles geolocation on the query. 
Right now, on development, this triggers functionality for the table view when entering map view as well, with no side effects that I can tell.

app/react/Library/helpers/requestState.js
Zasa-san: Map view well do 2 calls like before: one for aggregations, and one with `geolocation` to get markers.

app/react/App/scss/elements/_map.scss
```diff
@@ -548,8 +548,27 @@ a.mapboxgl-ctrl-logo {
   @include tooltip('right');
 }
 
+@media print {
+  /* stylelint-disable */
```
RafaPolit, line 552: I don't think this disable is needed? I removed the comments and my linter is not complaining.

app/react/Library/helpers/requestState.js
```diff
@@ -50,17 +50,18 @@ export function processQuery(params, globalResources, key) {
   };
 }
 
-export default function requestState(request, globalResources, calculateTableColumns = false) {
+export default function requestState(
+  request,
+  globalResources,
+  options = { calculateTableColumns: false, geolocation: false }
```
RafaPolit, line 56: For just two values, I believe we usually just used named properties, instead of equating them to options:

```
export default function requestState(request, globalResources, { calculateTableColumns = false, geolocation = false })
```
Something like that?

app/react/Library/helpers/requestState.js
```diff
@@ -50,17 +50,18 @@ export function processQuery(params, globalResources, key) {
...
   const documentsRequest = request.set(
     tocGenerationUtils.aggregations(docsQuery, globalResources.settings.collection.toJS())
   );
 
-  const templatesWithGeolocation = globalResources.templates.find(template =>
-    template.get('properties').find(property => property.get('type') === 'geolocation')
-  );
-
-  const markersRequest = templatesWithGeolocation
+  const markersRequest = options.geolocation
```
RafaPolit, line 64: There is a slight chance in logic here: previously, this happened always, but it only happened if there was, at least, one template with geolocation.  Now, we are doing the request only on map, but it happens even if there isn't a single geolocation property configured.  Are we sure we don't want to also limit this to not happen if there isn't a single geolocation property?

app/react/App/scss/elements/_map.scss
Zasa-san: Sure, I'll remove it.

app/react/Library/helpers/requestState.js
Zasa-san: Changed!

Zasa-san: I like this approach since you could omit the 3rd value if you don't need to pass any of the options. By using named properties the function will throw an error if I don't pass least an empty object.

app/react/Library/helpers/requestState.js
Zasa-san: Users are not be able to navigate to the map view if there are no templates with geolocation. They could manually input the map view URL to go there (in that case the query would happen). I wouldn't think is necessary to also check in `requestState` if there are templates with geolocation, since the query for geolocations only happens in that map view, I thought it was like checking twice.

PR #4064, opened 2021-12-02
3951 - OCR Stage 1

### LINKED ISSUE #3951: Support OCR Stage 1 (20 pts)
Original issue #64 ( from 2016! 👴 )
We will implement OCR support in two steps, first one being a button in the document to trigger an OCR for that document. This button should keep track of the state.

- Single button on the Document Viewer (for each PDF that could be accessed) that triggers the external service for OCRing.
- It replies immediately with a "OCR in progress"
- The "task manager" will eventually callback with some sort of socket interaction with the client when finished
- Call backend for status when reloading the document
- The button should be display when you have the configuration for the OCR service
- Add a confirm for the button.
- Delete the old PDF, replace it with the new one, and run all the process that normally happen ( delete text references, extract text...)
- e2e ?
- Keep the original as an attachment

Zasa-san:
fixes #3951

Important things to consider during testing:
- You need to run uwazi with EXTERNAL_SERVICES=true to enable the redis integration. If you need to change the redis port, you need REDIS_PORT=#### as well
- Test with multiple tenants.
- Tests real-time notifications with a couple of tabs having different files, and from different tenants as well.
- Test text references.

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/services/ocr/specs/OcrManager.spec.ts
```diff
@@ -0,0 +1,481 @@
...
+  settings: [
+    {
+      features: {
+        ocr: {
+          url: 'serviceUrl',
+        },
+      },
+    },
+  ],
+};
```
RafaPolit, line 128: If the fixtures are half the length of our max amount of lines, please extract them into their own file.

RafaPolit (changes requested):
Half way reviewing, please start tackling these issues.  Thanks.

app/api/files/ocrRoutes.ts
```diff
@@ -0,0 +1,74 @@
...
+  properties: {
+    params: {
+      properties: {
+        filename: { type: 'string' },
+      },
+    },
+  },
+};
+
+async function fileFromRequest(request: Request) {
```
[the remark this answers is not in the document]
Zasa-san: Fixed this, we'll check other files for this as well.

app/api/files/ocrRoutes.ts
```diff
@@ -0,0 +1,74 @@
...
+  });
+
+  if (!file || !(await fileExists(uploadsPath(file.filename)))) {
+    throw createError('file not found', 404);
+  }
+
+  return file;
+}
+
+// eslint-disable-next-line import/no-default-export
```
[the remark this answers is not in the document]
Zasa-san: Done, exported named.

app/api/files/ocrRoutes.ts
```diff
@@ -0,0 +1,74 @@
...
+    validation.validateRequest(ocrRequestDecriptor),
+    async (req, res, next) => {
+      try {
+        const file = await fileFromRequest(req);
+
+        const status = await OcrManager.getStatus(file);
+
+        res.json(status);
+      } catch (e) {
+        next(e);
```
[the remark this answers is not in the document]
Zasa-san: Yes, removed the try catchs.

app/api/files/specs/ocrRoutes.spec.ts
```diff
@@ -0,0 +1,227 @@
...
+      { overwriteRoutes: true }
+    );
+  }
+
+  beforeEach(async () => {
+    spyOn(search, 'indexEntities').and.returnValue(Promise.resolve());
+    await testingEnvironment.setUp(FIXTURES);
+    testingEnvironment.setPermissions(adminUser);
+    requestMockedUser = adminUser;
+    jest.spyOn(Date, 'now').mockReturnValue(1000);
```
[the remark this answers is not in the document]
Zasa-san: Unsure if spies can survive. For safety we are doing a restoreAllMocks.

app/api/services/ocr/specs/OcrManager.spec.ts
Zasa-san: Sure!

app/api/services/ocr/specs/OcrManager.spec.ts
```diff
@@ -0,0 +1,374 @@
...
+          { _id: fixturesFactory.id('resultToDelete2') },
+        ],
+      });
+      let records = await OcrModel.get({ sourceFile: { $in: filesToCleanup.map(f => f._id) } });
+      await OcrManager.cleanupRecordsOfFiles(filesToCleanup.map(f => f._id));
+      records = await OcrModel.get({ _id: { $in: records.map(r => r._id) } });
+      expect(records).toHaveLength(0);
+    });
+  });
+});
```
RafaPolit, line 374: This entire file has the wrong naming conventions on describes and its.

Every `describe` is a scenario or conditional: "on error", "when there is no tenant", "if promise rejected".  It should never start with a `should`.  Every `it`, on the other hand, MUST start with a `should`.

On rare occasions, an `it` also involves a single conditional. In order to avoid a new `describe` with a single `it`, we put the condition after the `should`.  So: "on cleanup" -> "should modify record source to null when source is deleted".

But always start with a should.

app/api/services/ocr/specs/fixtures/fixtures.ts
```diff
@@ -0,0 +1,113 @@
...
+      features: {
+        ocr: {
+          url: 'serviceUrl',
+        },
+      },
+    },
+  ],
+};
+
+export { FIXTURES };
```
RafaPolit, line 113: This is a new naming convention with all caps.  I'm fine with it for now, let's discuss this further.

app/react/Settings/components/CollectionSettings.tsx
```diff
@@ -53,6 +53,7 @@ const CollectionSettings = ({
   register('mapStartingPoint');
   register('cookiepolicy');
   register('newNameGeneration');
+  register('toggleOCRButton');
```
RafaPolit, line 56: This is not toggling on a button, it's toggling on an entire feature.  Is this the name in the collection as well?  Please, rename it to: "toggleOCRService".

app/react/Settings/components/CollectionSettings.tsx
```diff
@@ -198,6 +199,21 @@ const CollectionSettings = ({
...
+          </h2>
+
+          <SettingsFormElement label="Document OCR trigger" tip={tips.ocrTrigger}>
+            <ToggleButton
+              checked={Boolean(watch('toggleOCRButton'))}
+              onClick={() => {
+                setValue('toggleOCRButton', !getValues('toggleOCRButton'));
+              }}
+            />
+          </SettingsFormElement>
```
RafaPolit, line 214: Same as above, this is not a button we are enabling.

app/react/Viewer/actions/ocrActions.ts
```diff
@@ -0,0 +1,15 @@
+import api from 'app/utils/api';
+
+const postToOcr = async (filename: string) => {
+  await api.post(`files/${filename}/ocr`);
```
RafaPolit, line 4: It is correct that this function doesn't return anything? Are not doing anything with the response of the POST either to acknowledge it received or to catch any errors?

app/react/Viewer/components/OCRDisplay.tsx
```diff
@@ -0,0 +1,190 @@
...
+  useEffect(() => {
+    if (ocrIsToggled) {
+      getOcrStatus(file.filename || '')
+        .then(({ status, lastUpdated }) => {
+          setOcrStatus({ status, lastUpdated });
+          if (status === 'inQueue') {
+            // @ts-ignore
+            socket.on('ocr:ready', listenOnSuccess);
+            // @ts-ignore
+            socket.on('ocr:error', listenOnError);
```
RafaPolit, line 58: Did we try to add types for socket? Why are these ignores required?

app/react/Viewer/components/OCRDisplay.tsx
```diff
@@ -0,0 +1,190 @@
...
+};
+
+const mapDispatchToProps = (dispatch: any) =>
+  bindActionCreators({ loadDocument: reloadDocument }, dispatch);
+
+const connector = connect(mapStateToProps, mapDispatchToProps);
+type mappedProps = ConnectedProps<typeof connector>;
+type ComponentProps = OCRDisplayProps & mappedProps;
+
+const OCRDisplay = ({ file, ocrIsToggled, locale, loadDocument }: ComponentProps) => {
```
RafaPolit, line 35: I think this component is wrongly named.  OCRDisplay sounds more like the actual result of the OCR.  This is more like OCRStatus? This includes the name of the file.

app/react/Viewer/components/OCRDisplay.tsx
```diff
@@ -0,0 +1,190 @@
...
+        </div>
+      );
+      break;
+
+    default:
+      break;
+  }
+
+  return (
+    ocrIsToggled && (
```
RafaPolit, line 180: The entire component is irrelevant if `orcIsToggled` is off... yet we wait all the logic before "dismissing" the component?  Doesn't it make more sense to put this conditional on line 1 of the component: `if (!ocrIsToggled) return null;` and we prevent all this case switching and stuff if we don't want to render anything?

app/react/Viewer/components/Viewer.js
```diff
@@ -243,4 +245,5 @@ const mapDispatchToProps = dispatch =>
     dispatch
   );
 
-export default connect(mapStateToProps, mapDispatchToProps)(Viewer);
+const ConnectedViewer = connect(mapStateToProps, mapDispatchToProps)(Viewer);
+export { ConnectedViewer, Viewer };
```
RafaPolit, line 249: If we are exporting the `Viewer` only for testing, it is better to just export the connected component as Viewer and use the `wrappedComponent` in the test.

RafaPolit (changes requested):
Another batch of comments.

app/react/Viewer/utils/ocrDisplayTips.ts
```diff
@@ -0,0 +1,26 @@
...
+    ),
+  unsupportedLang: (language: string) => {
+    let tip = "The document's language is not supported.";
+    if (language === 'other') tip = 'Please select a language for this document';
+    return t('System', tip, null, false);
+  },
+  cantProcess: (time: string) =>
+    `${t(
+      'System',
+      'The OCR engine couldn’t read the document. Try uploading the document in a different format. Last updated',
```
RafaPolit, line 19: This one is tricky!  Format can only be PDF, so "another format" is misleading.  Let's look into this after we finish. Please create an issue.

app/react/components/Elements/FeatureToggle.tsx
```diff
@@ -21,11 +21,13 @@ FeatureToggle.defaultProps = {
 };
 
 function mapStateToProps({ settings }: any, ownProps: OwnPropTypes) {
-  const features = settings.collection.get('features') || Immutable.fromJS({});
+  const features = settings.collection.get('features');
+
   return {
-    featureActivated: features.get(ownProps.feature),
+    featureActivated: features ? Boolean(get(features.toJS(), ownProps.feature)) : false,
```
RafaPolit, line 27: Why is this change to lodash needed? Why not use `Boolean(features.get(ownProps.feature))`?  `toJS` operations tend to be expensive.

app/react/socket.js
```diff
@@ -1,17 +1,17 @@
...
 }
 
-export default socket;
-
 const reconnectSocket = () => {
   socket.disconnect();
   socket.connect();
 };
 
+export default socket;
```
RafaPolit, line 16: Try to avoid use export default.  I know this was here before... is it too hard to change it?  Maybe not worth it right now. Thanks for putting the export at the bottom of the file.

app/server.js
```diff
@@ -162,6 +163,8 @@ DB.connect(config.DBHOST, dbAuth).then(async () => {
           );
 
           segmentationRepeater.start();
+
+          OcrManager.start();
```
RafaPolit, line 167: I think that having all external services into a single ENV variable is really, really wrong.

app/shared/types/settingsSchema.ts
```diff
@@ -119,6 +119,7 @@ export const settingsSchema = {
     customCSS: { type: 'string' },
     mapTilerKey: { type: 'string' },
     newNameGeneration: { type: 'boolean', enum: [true] },
+    toggleOCRButton: { type: 'boolean' },
```
RafaPolit, line 122: So, for sure, this name in the settings schema is wrong.  This is enabling the entire OCR service for a collection, not a button.

RafaPolit (changes requested):

app/api/services/ocr/specs/fixtures/fixtures.ts
fnocetti: Fixed

app/api/services/ocr/specs/OcrManager.spec.ts
fnocetti: Will take a look

app/react/Settings/components/CollectionSettings.tsx
fnocetti: Agreed. We are changing it to "ocrServiceEnabled". We think it's more descriptive than an "action" phrase.

app/server.js
RafaPolit: Obviously, we are not going to fix this now.  It's a comment for another time.  Do not change this right now.

app/react/Viewer/actions/ocrActions.ts
fnocetti: The UX design (or the issue) specified that the button immediately changes to the "In queue" without waiting for the response. If I'm not wrong this is a UX pattern in which you predict the successful action. Now you say this, probably we still need to catch the error and "rollback" the UI, but since you still get an error notification and the UX design didn't specified it. We basically overlooked it.
If you think it's a need we can implement it, in other case we can iterate on it in the future.

app/react/Viewer/components/OCRDisplay.tsx
fnocetti: We will fix this by adding the needed parameters to the JS declarations, not by migrating to TS that can take longer.

app/server.js
fnocetti: That env variable doesn't control the services, but controls if uwazi will try to connect to a redis service. The service-related configuration is not in this env variable. We can separate the variable for the OCR, I'm not completely against it, but I think it's mostly a decision about semantics.

app/react/Viewer/components/OCRDisplay.tsx
Zasa-san: Yes, this totally correct.

app/react/components/Elements/FeatureToggle.tsx
Zasa-san: We will avoid using toJS by using getIn.

app/api/services/ocr/OcrManager.ts
```diff
@@ -0,0 +1,233 @@
...
+  url: string;
+}
+
+const LANGUAGES_MAP: { [key: string]: string } = {
+  arb: 'ar',
+  deu: 'de',
+  eng: 'en',
+  fra: 'fr',
+  spa: 'es',
+};
```
RafaPolit, line 35: Please, do not use a locally defined hashmap!
We have `app/shared/languagesList.ts` for this exact purpose.

fnocetti: Working on it. For some reason we thought it was under the frontend package.

app/api/services/ocr/OcrManager.ts
```diff
@@ -0,0 +1,233 @@
...
+    } catch (e) {
+      handleError(e);
+    }
+  }, message.tenant);
+};
+
+const validateLanguage = async (language: string, ocrSettings?: { url: string }) => {
+  const _ocrSettings = ocrSettings || (await getSettings());
+  const supportedLanguages = await fetchSupportedLanguages(_ocrSettings);
+  return supportedLanguages.includes(LANGUAGES_MAP[language]);
```
RafaPolit, line 146: Replace this local hashmap with the languagesList ISO639_1 value

app/api/services/ocr/OcrManager.ts
```diff
@@ -0,0 +1,233 @@
...
+      file.filename,
+      fileContent
+    );
+
+    await this.ocrTaskManager!.startTask({
+      task: this.SERVICE_NAME,
+      tenant: tenant.name,
+      params: {
+        filename: file.filename,
+        language: LANGUAGES_MAP[file.language || 'other'],
```
RafaPolit, line 224: When 'other', is this returning `undefined`? Because the `other` key does not exist on the local hashmap.  Still, replace this with the external one and please look into the `other` issue.

fnocetti: It's not an issue. The default is needed here because the FileType has language defined as optional (is it really optional) so we need a default for typescript to work. Any way, if the flow reached this point, file.language will have a valid supported language.
I'll change the strategy to forcing it non-optional inline.

app/react/Viewer/components/OCRStatus.tsx
```diff
@@ -0,0 +1,184 @@
...
+};
+
+const mapDispatchToProps = (dispatch: any) =>
+  bindActionCreators({ loadDocument: reloadDocument }, dispatch);
+
+const connector = connect(mapStateToProps, mapDispatchToProps);
+type mappedProps = ConnectedProps<typeof connector>;
+type ComponentProps = OCRStatusProps & mappedProps;
+
+// eslint-disable-next-line max-statements
```
RafaPolit, line 35: This component can really be refactored.
Abstract more each case for the status, looks like most could be a simple function that returns a `div` with a `p` and a `Tranlsate`, only leave those that are really different (like the "noOcr" case) 
If there isn't enough time now, please create an issue to refactor this as tech debt so we don't leave it hanging there for too long.

app/react/Viewer/components/specs/OCRStatus.spec.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+    it('should change to display that the ocr has failed when the service reports error', async () => {
+      await renderAndSubmit();
+      act(() => {
+        mockSocketOn['ocr:error']('file_id');
+      });
+      expect(await screen.findByText('Could not be processed')).not.toBeNull();
+    });
+
+    it('should liste to the ocr service if the document is in the ocr queue when the component loads', async () => {
```
RafaPolit, line 148: spelling: liste

app/react/socket.js
```diff
@@ -1,17 +1,17 @@
 import io from 'socket.io-client';
 import { isClient } from 'app/utils';
 
-let socket = { on: () => {} };
+let _socket = { on: (_event, _listener) => {}, off: (_event, _listener) => {} };
```
RafaPolit, line 4: Why the underscored variable name?

app/react/socket.js
```diff
@@ -1,17 +1,17 @@
...
-
 const reconnectSocket = () => {
-  socket.disconnect();
-  socket.connect();
+  _socket.disconnect();
+  _socket.connect();
 };
 
+export const socket = _socket;
 export { reconnectSocket };
```
RafaPolit, line 17: This could have been `export { socket, recconetSocket }` and we avoid the underscored name, right?

app/react/socket.js
```diff
@@ -1,17 +1,17 @@
...
-export default socket;
-
 const reconnectSocket = () => {
-  socket.disconnect();
-  socket.connect();
+  _socket.disconnect();
+  _socket.connect();
 };
 
+export const socket = _socket;
```
RafaPolit, line 16: This could have been `export { socket, recconetSocket }` and we avoid the underscored name, right?

RafaPolit (commented):
Very minor comments now.  I think the code QA is finished, I'll start the functional QA.

app/react/socket.js
fnocetti: Because the parameter is not being used in the body, so eslint will complain, but we need them for the arity of the function to be correctly infered on ts files.

app/react/socket.js
fnocetti: There's an eslint rule stating that we shouldn't export variables (`let`) but constants.

app/react/Viewer/components/OCRStatus.tsx
Zasa-san: Will create the issue!

RafaPolit:
Functional QA:
- When first uploading a scanned PDF, the file is assigned language "other", which is correct, yet the OCR status still reads "Unsupported OCR Language" instead of "please, select a language for the file".  The tooltip does instruct of this, so this is a Nice to Have, we can live with what is right now.

The problem arises next:
- If I edit the document and save the language of the file, the response does not affect the current status
- if the language I assigned is supported, I still see the same "unsupported OCR language" with the tooltip asking me to assign a language, which I have just done! 
- if the language is unsupported, I also still see the tooltip to assign a language which I already did

The real problem here is that I need to refresh in order to get the right status, which makes no sense.

app/shared/detectLanguage.ts
```diff
@@ -1,5 +1,6 @@
 import franc from 'franc';
-import get from 'shared/languagesList';
+import { language as get } from 'shared/languagesList';
```
RafaPolit, line 2: Get is really a very poor name here!  I get it you are just reassigning the import, but this one is really a sucky name. :(

fnocetti: Fixed

RafaPolit:
LGTM!

PR #4072, opened 2021-12-08
Allow link without label  on Link properties

### LINKED ISSUE #3977: Have a default value/allow for blank label in link fields ( 3 pts)
Link fields do not allow for blank or default value for the label. I think this is UX issue since adding a URL becomes necessarily a 2 step action when sometimes all you need to do is to add the URL and use the URL as label as well (ie. you want the domain name and the rest of the URL to be visible).

image

IMO we should allow blank labels, when there is no label present, the URL is displayed as label.

grafitto:
fixes #3977 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Metadata/helpers/specs/validator.spec.js
+++ b/app/react/Metadata/helpers/specs/validator.spec.js
@@ -61,9 +61,6 @@ describe('metadata validator', () => {
       expect(labelAndUrl({ label: 'cat', url: '' })).toBe(false);
       expect(labelAndUrl({ label: 'cat' })).toBe(false);
     });
-    it('should be true when url is present and label is not', () => {
-      expect(labelAndUrl({ label: '', url: 'cat.pic.com' })).toBe(true);
-    });
   });
 
   describe('generate', () => {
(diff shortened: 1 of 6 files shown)


app/react/Metadata/helpers/specs/validator.spec.js
```diff
@@ -51,17 +51,18 @@ describe('metadata validator', () => {
...
 
-    it('should be false when one value is present and the other is not', () => {
+    it('should be false when label is present and url is not', () => {
       expect(labelAndUrl({ label: 'cat', url: '' })).toBe(false);
-      expect(labelAndUrl({ label: '', url: 'cat.pic.com' })).toBe(false);
       expect(labelAndUrl({ label: 'cat' })).toBe(false);
-      expect(labelAndUrl({ url: 'cat.pic.com' })).toBe(false);
+    });
+    it('should be true when url is present and label is not', () => {
+      expect(labelAndUrl({ label: '', url: 'cat.pic.com' })).toBe(true);
```
konzz, line 65: This scenario is the same as in line 55

grafitto: Removed

PR #4076, opened 2021-12-10
4075 cannot change file language

### LINKED ISSUE #4075: Cannot change file language
- Have an entity with a main document.
- Go to entity view and click on the `Edit` green button near the file, under `Primary documents`.
- Change the language and save.
- A validation error saying `.body should not have additional properties` shows and the language is not updated.

This is not happening in production.

mfacar:
fixes #4075


app/api/entities/entities.js
```diff
@@ -347,7 +347,7 @@ const withDocuments = async (entities, documentsFullText) => {
   const sharedIds = entities.map(entity => entity.sharedId);
   const allFiles = await files.get(
     { entity: { $in: sharedIds } },
-    documentsFullText ? '+fullText ' : ' '
+    documentsFullText ? '+fullText -__v ' : ' -__v '
```
RafaPolit, line 350: I don't think this works.  You cannot both include and exclude values in the same query.  So, the right hand part will work, the left one will give an error or, at least, ignore the flag.  Please confirm that.

app/api/files/files.ts
```diff
@@ -20,7 +21,8 @@ export const files = {
     if (index) {
       await search.indexEntities({ sharedId: savedFile.entity }, '+fullText');
     }
-    return savedFile;
+    const { __v, ...document }: WithId<FileType> & { __v?: number } = savedFile;
```
RafaPolit, line 24: Why is this called "document"?  Is this meant to be a "mongo document"? Is it `fileInfo`? should it just be "response"? or maked savedFile a variable and reassign it?  Not sure.

Still, if we omit the value from the model, should that not take care of the value returning in the save operation?

RafaPolit (commented):
Not sure about the actual implementation of the approach. Please, check the comments.

app/api/entities/entities.js
mfacar: I confirmed that it is possible to have both include exclude in the same query, it works well. I can't add a test to check that because I don't know why but the mongo memory instance for the test is not returning the versionKey field. Thats the reason I just added an e2e test.

app/api/files/files.ts
mfacar: Totally agree, that's a better solution, actually, there's a specific option to don't include the versionKey

app/api/entities/entities.js
RafaPolit: Can you please link any documentation mentioning you can both include and exclude fields? This is the official mongodb documentation:
https://docs.mongodb.com/v4.4/reference/method/db.collection.find/#projections
>The projection parameter specifies which fields to return. The parameter contains either include or exclude specifications, not both, unless the exclude is for the _id field.

RafaPolit: Furthermore, Mongoose, up to it's last version also limits this:
https://mongoosejs.com/docs/api/query.html#query_Query-select
> A projection must be either inclusive or exclusive. In other words, you must either list the fields to include (which excludes all others), or list the fields to exclude (which implies all other fields are included). The _id field is the only exception because MongoDB includes it by default.

RafaPolit: Ok, so, just to explain: this is actually NOT a include and exclude operation in the same operation.  The `+` operator forces the inclusion of a path that would have otherwise been hidden by the schema definition.  So this is not a "select only fullText and exclude __v", it's a "select all, ADD fullText and exclude __v".

Issue #4087, opened 2021-12-15
CSV import missing thesauri values


LaszloKecskes:
If there are multiple select/multiselect properties, and only some of those have new values, the pre-flight might miss saving values. Then, when it gets to saving the entity, it tries to look for a thesauri value that does not exist and various error might occur.

The solution is to make sure that the saving picks up the proper set of thesauri to save.

Steps to reproduce:

- create two separate thesauri with a value each
- create two selects in a template, link each to a different thesauri
- try to import an entity, that does not have a new value in the first select, but has a new value in the second

PR #4088, opened 2021-12-15
Csv import: column names and thesauri save fix

### LINKED ISSUE #4059: CSV import composite issue (5 pts)
This is a composite issue that describes multiple bugs with the csv import process, which we decided to keep together since they involve the same part of the code.
The csv import is a two steps process. First, a pre-flight (csvLoader - arrangeThesauri) collects the new values from the select and multiselect fields, then the main loop after performs the actual entity import one-by-one.

Bug 1: Column Name Sanitization

In the pre-flight step, the function assumes that the csv column names are already safe property names (lowercase, underscore instead of spaces). This is not the case; the file could contain the raw label, or even different capitalization (Example Select Value, instead of example_select_value). This results in the new values not loaded in the thesauri, then the entity import fails at the first new value.
Solution: The pre-flight needs to sanitize the column names before matching to properties. The sanitization needs to be consistent with what already happens in properties, preferably through the same function.
With this bug, ping me (@LaszloKecskes ) for help about the pre-flight, if needed.

Steps to reproduce:

- create a thesauri
- create a select or a multiselect property with a label that will be different from the safe name
- try to import an entity with a new value, where the column uses the label

<del>Bug 2: Triplicate Imports</del>

<del>While testing the previous bug, for some people the entity import resulted in triplicate imports (i.e. the same set of entities imported three times).</del>
<del>I have not been able to reproduce the problem. @nwyu  reported this, and could supply more information.</del>

<del>Bug 3: #3764 - CSV import of nested thesaurus values fails</del>

<del>If you CSV import an entity with a select or multiselect with the value of a nested thesaurus item, the value appears on cards (denormalized?) but when you edit the entity, it doesn't show on the selectors.</del>

LaszloKecskes:
fixes #4059, #4087

For the PR, please note that there are two issues, and functional testing should cover both.

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/csv/specs/csvLoaderSelects.spec.js
+++ b/app/api/csv/specs/csvLoaderSelects.spec.js
@@ -21,9 +21,9 @@ const commonTranslationContexts = (id1, id2) => [
   },
   {
     id: id1.toString(),
-    label: 'select_thesaurus',
+    label: 'Select Thesaurus',
     values: [
-      { key: 'select_thesaurus', value: 'select_thesaurus' },
+      { key: 'Select Thesaurus', value: 'Select Thesaurus' },
       { key: 'A', value: 'A' },
     ],
     type: 'Dictionary',
@@ -42,17 +42,19 @@ const commonTranslationContexts = (id1, id2) => [
 
 const fixtures = {
   dictionaries: [
-    fixtureFactory.thesauri('select_thesaurus', ['A']),
+    fixtureFactory.thesauri('Select Thesaurus', ['A']),
     fixtureFactory.thesauri('multiselect_thesaurus', ['A', 'B']),
   ],
   templates: [
     fixtureFactory.template('template', [
       fixtureFactory.property('unrelated_property', 'text'),
       fixtureFactory.property('select_property', 'select', {
-        content: fixtureFactory.id('select_thesaurus'),
+        content: fixtureFactory.id('Select Thesaurus'),
+        label: 'Select Property',
       }),
       fixtureFactory.property('multiselect_property', 'multiselect', {
         content: fixtureFactory.id('multiselect_thesaurus'),
+        label: 'Multiselect Property',
       }),
     ]),
     fixtureFactory.template('no_selects_template', [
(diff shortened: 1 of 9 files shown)


app/api/csv/specs/csvLoaderSelects.spec.js
```diff
@@ -9,7 +9,7 @@ import { CSVLoader } from '../csvLoader';
 
 const fixtureFactory = getFixturesFactory();
 
-const commonTranslationContexts = (id1, id2) => [
+const commonTranslationContexts = (id1, id2, id3) => [
```
mfacar, line 12: @LaszloKecskes Probably not related to this PR, but if I delete all the commonTranslationContexts except system, the tests continue passing.

mfacar (changes requested):
LGTM! I did some tests with different cases, everything works as expected.  
I left a comment about some fixtures that apparently are not needed, please take a look.

app/api/csv/specs/csvLoaderSelects.spec.js
LaszloKecskes: That one's tricky, since it is something that we did not test first and it is a smoke test for an issue we had during writing. 
We have assumed that the thesaurus saving will take care of the translations. However, there was an issue with concurrency: we are saving multiple thesauri, and when we did it in an async manner, they have overwritten each other's changes to the translations, and only the last one got entered properly. So, for the test I have decided to have a "translations" structure as close to the real one as possible, and then test that the changes go through as they should.
However, since this was a smoke test, and the related problem was taken care of (this is the reason for the somewhat convoluted syncSaveThesauri) , now it seems like this is not testing anything. Nevertheless, I would leave it as is, or just add a comment somewhere (in the test, next to the fixtures, both...) to indicate that this was a smoke test.

mfacar: Ok, it's fine I think. I'll merge it as it is.

PR #4100, opened 2021-12-20
Sanitize _id in nested filters

### LINKED ISSUE #3983: Collection Filters not sanitizing _id upon drag and fails validation (8 pts)
If you have already defined collection filters, all "root placed" templates have an `_id` property assigned after save.
Upon re-entering the settings, those `_id`s are also carried over if you drag them inside a group.  Upon saving, this fails validation, because `_id` is not a valid `items` property.

Way to reproduce:
- Ensure collection:filters are empty, save
- Add a template X to collection filters (any template will do) save
- RELOAD!
- Create a Group and drag the template X inside the group, save

You get an error of validation, claiming that `_id` is not a valid property for `items`.

We could do any of the following approaches:
- sanitize this in client upon dragging, or upon saving (this one requires parsing, but is probably more secure due to scenarios mentioned below)
- accept the `_id` in the route and then sanitize that
- add the `_id` to the items in the DB as well for consistency

At any rate, the solution should work with multiple operations:
- adding from scratch root properties and groups with children
- dragging an in-filter root property inside a newly created group (this is the failing scenario of this issue)
- dragging a new non-used property inside a new group
- dragging a non-used property inside an existing group
- dragging an existing child OUT of the group into the root of the filters
- dragging something inside a group and then dragging it back (without a save in between)

There are lots of combinations that could end up deleting `_id`s or assigning the wrong thing here, so please don't just address the described bug, but ensure that the entire flow is working.

Perhaps an e2e for this is important? We have had several bugs now concerning the filter grouping, which is not tested through our integration tests.

fnocetti:
fixes #3983 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [x] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [x] Code review


Zasa-san (approved):
LGTM!

PR #4117, opened 2022-01-05
Export CSV respecting inherited values

### LINKED ISSUE #3945: Export CSV should respect inherited values, not title of the related entity
It has been reported that CSV export of inherited properties is exporting the title of the Entity, and not really the inherited property.

Since this value is already denormalized, it should not be too complex to change this.

elreplicante:
fixes #3945 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/shared/types/commonTypes.d.ts
+++ b/app/shared/types/commonTypes.d.ts
@@ -63,13 +63,19 @@ export type PropertyValueSchema =
   | LatLonSchema
   | LatLonSchema[];
 
+export interface InheritedValueSchema {
+  value: string;
+}
+
 export interface MetadataObjectSchema {
   value: PropertyValueSchema;
   label?: string;
   suggestion_confidence?: number;
   suggestion_model?: string;
   provenance?: '' | 'BULK_ACCEPT';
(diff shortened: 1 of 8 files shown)


app/shared/types/commonTypes.d.ts
```diff
@@ -71,7 +71,7 @@ export interface MetadataObjectSchema {
   provenance?: '' | 'BULK_ACCEPT';
   inheritedValue?: {
     [k: string]: unknown | undefined;
-  };
+  }[];
```
daneryl, line 74: @elreplicante this type file is auto-generated from a json schema definition https://github.com/huridocs/uwazi/blob/1ffb53f40def29a8865b598e75cd8addb667b5f6/app/shared/types/commonSchemas.ts#L105, you will need to change that file and then run yarn emit-types to update this file.

daneryl (changes requested):
@elreplicante looks good!, only a minor comment regarding the type definition.

daneryl (approved):

PR #4123, opened 2022-01-11
4149 - Transaction over adding a language

### LINKED ISSUE #4113: Do a low-level entity duplicate on language creation (13 pts)
Language creation should not use the entities model, it should duplicate entities with the db.collection methods as is done with the sync. This process should be wrapped in a mongo transaction (starting with the language creation) so that, if it fails for any reason, we are not left with inconsistent databases. All this avoids any checks that we don't need to happen ONLY on language creation. This requires an attention to indexing those into elastic, but that is already being taken care of in the sync process, it could be used as a starting point.

split from #3955

### LINKED ISSUE #4149: Transaction over language creation and duplication of entities
Split from  #4113 

The processing of adding a language should be wrapped in a mongo transaction (starting with the language creation) so that, if it fails for any reason, we are not left with inconsistent databases.

mfacar:
fixes #4149

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


mfacar:
Closing until a further discussion about transactions over MongoDB

PR #4146, opened 2022-01-17
Fixed aggregations wrong limit bug

### LINKED ISSUE #4094: Multiselect filters that have groups should use the parent in the URL instead of the children (8 pts)
When you have a thesauri with groups, and you click the parent, that action puts all the children in the url as filters, but if then you use that URL in a page, and also add more values to that group, you will expect that when you visit that url the ne sub values are also been filtered by.

Example:
- You have library filters like this:
Screenshot 2021-12-17 at 15 55 54

- You use the url to create a link pointing to Cool People
- Someone adds new cool people, let's add Santi and Joan
- And then when you visit again that filter you get this:

Screenshot 2021-12-17 at 15 57 08

### LINKED ISSUE #4142: Wrong aggregations returned to client (5 pts)
When sending back aggregations to the client we limit the amount of aggregations we send back to the client by certain amount to avoid scenarios on where the options are thousands.

This limit was wrongly applied over all the aggregations, instead of the previously filtered aggregations that have results, and because of this an aggregation with results but at the bottom of that list will never reach the client.

In line 370 we filter the aggregations, but then the limit in the line 373 is applied over the original object.

https://github.com/huridocs/uwazi/blob/4e4487f8b8c5b24f610c4f6467e67cb14848bb6a/app/api/search/search.js#L369-L379

konzz:
fixes #4142

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


elreplicante (approved):
LGTM!

PR #4153, opened 2022-01-19
Add reindex flag to migrations

### LINKED ISSUE #2872: Reindex as a migration
Every time a development requires a reindex, it should include a migration that does the reindex.  This solves the problem of having to 'flag' a PR or master status as 'requiring reindex'.  It also allows users hosting themselves to just run the server and have it run the reindex if needed.

### LINKED ISSUE #4122: Add reindex flag to migrations (5 pts)
We want to avoid human errors when deploying so if a deploy needs a reindex it should be marked in a migration.

Possible options:
- The migration has a flag to say reindex or not.
- The migration has a flag but can be changed on the fly when an instance does not need to be migrated
- The migration returns some information including if it requires a migration or not.

⚠️ Beware of making this backwards compatible with previous migrations (it should not throw errors if flag is not defined)

elreplicante:
fixes #4122
fixes #2872

Migrations will now be created with a property `reindex` set to `false` by  default.  A developer can check programatically whenever a reindex is needed or not and change that property's value.  If any of the pending migrations needs  a reindex, the `yarn migrate` command will output a JSON that will be handled by Ansible to decide if the `yarn reindex` command will be executed or not.  Here's an  example output:

Screenshot 2022-01-19 at 11 18 30

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/migrations/migrate.ts
+++ b/app/api/migrations/migrate.ts
@@ -29,8 +29,8 @@ export const run = async () => {
   errorLog.closeGraylog();
   await DB.disconnect();
 
-  const reindexNeed = migrations.some(migration => migration.reindex === true);
-  process.stdout.write(`{ reindex: ${reindexNeed} }`);
+  const reindexNeeded = migrations.some(migration => migration.reindex === true);
+  process.stdout.write(JSON.stringify({ reindex: reindexNeeded }));
 };
 
 run().catch(async e => {
--- a/app/api/migrations/migrate.ts
+++ b/app/api/migrations/migrate.ts
@@ -18,16 +18,19 @@ if (process.env.DBUSER) {
   };
 }
 
-const run = async () => {
+export const run = async () => {
   await DB.connect(config.DBHOST, auth);
   const { db } = await DB.connectionForDB(config.defaultTenant.dbName);
-
+  let migrations: any[] = [];
   await tenants.run(async () => {
-    await migrator.migrate(db);
+    migrations = await migrator.migrate(db);
   });
   //@ts-ignore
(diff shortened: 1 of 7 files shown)


LaszloKecskes:
@daneryl , @elreplicante 
I've been looking through this since I need to use the same for the migration I'm currently working on, and I've been wondering: what would you think about adding to the tests template some reminder (maybe in the form of two dummy tests with appropriate descriptions) that one should test what the reindex flag gives back? 
We will probably not need it for some cases, and also these may need more than just one test each for false/true, but I think some reminder could be useful.

elreplicante:
> @daneryl , @elreplicante I've been looking through this since I need to use the same for the migration I'm currently working on, and I've been wondering: what would you think about adding to the tests template some reminder (maybe in the form of two dummy tests with appropriate descriptions) that one should test what the reindex flag gives back? We will probably not need it for some cases, and also these may need more than just one test each for false/true, but I think some reminder could be useful.

@LaszloKecskes makes sense.  I've added a test to the template that will always fail checking on `migration.reindex`.  Do you think is this enough?

https://github.com/huridocs/uwazi/pull/4153/files#diff-458c4edfedc3499ee081e8c68881699eb84fbe617980eeb16cce79dd12fe4328R22-R25

Thanks for the suggestion!

LaszloKecskes:
@elreplicante 
I think yes, the reminder is there, then now the developer can decide whether they need more paths (tests) or not. Thanks!

PR #4162, opened 2022-01-21
fixed flaky by properly awaiting on the previous test ?

### LINKED ISSUE #3832: Flaky unit test in app/api/entities/specs/entites.spec.js
A specific unit test sometimes fails on the github CI. Usually, in those cases one cannot reproduce the error locally, and re-running the unit tests resolves the issue without any change.

An example error report from the CI:
```
Summary of all failing tests
FAIL app/api/entities/specs/entities.spec.js (26.183s)
  ● entities › save › when generatedToc is undefined › should not replicate the value to all languages

    TypeError: Cannot read property 'template' of undefined

      134 |   const docLanguages = await this.getAllLanguages(entity.sharedId);
      135 |   if (
    > 136 |     docLanguages[0].template &&
          |                     ^
      137 |     entity.template &&
      138 |     docLanguages[0].template.toString() !== entity.template.toString()
      139 |   ) {

      at Object.template [as updateEntity] (api/entities/entities.js:136:21)
          at runMicrotasks (<anonymous>)
      at Object.save (api/entities/entities.js:416:7)


Test Suites: 1 failed, 603 passed, 604 total
Tests:       1 failed, 3793 passed, 3794 total
Snapshots:   423 passed, 423 total
Time:        523.648s
Ran all test suites in 2 projects.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Error: Process completed with exit code 1.
```

daneryl:
fixes #3832 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/entities/specs/entities.spec.js
+++ b/app/api/entities/specs/entities.spec.js
@@ -284,7 +284,7 @@ describe('entities', () => {
       });
     });
 
-    it('should ignore the permissions parameter', () => {
+    it('should ignore the permissions parameter', async () => {
       const doc = {
         _id: unpublishedDocId,
         sharedId: 'other',
@@ -292,12 +292,12 @@ describe('entities', () => {
         permissions: [],
       };
 
-      entities.save(doc, { language: 'en' }).then(updatedDoc => {
-        expect(updatedDoc.permissions).toEqual([
-          expect.objectContaining({ refId: 'user1' }),
-          expect.objectContaining({ refId: 'user2' }),
-        ]);
-      });
+      const updatedDoc = await entities.save(doc, { language: 'en' });
+
+      expect(updatedDoc.permissions).toEqual([
+        expect.objectContaining({ refId: 'user1' }),
+        expect.objectContaining({ refId: 'user2' }),
+      ]);
     });
 
     describe('when generatedToc is undefined', () => {


LaszloKecskes (changes requested):
Code looks good, but check comment.

daneryl:
> It is a reasonable guess! The same pattern is used in lines 166, 187, 227, 262(*), 295 (fix), 335, 447, 541, 564, 604, 649, 997, 1037, 1056, 1086, 1110, 1129, 1165, 1176, 1187, 1253 (I thought it will be less when I've started... :) ). I think the one that actually causes the problem might be not this one, but the one on 262, because that is the save that matches sharedId with the flaky test. However, if we fix this one, then this pattern might cause the same thing in other places, so probably we should try to fix the entire file. Also, I've only listed saves, updates and deletes, but there's the same pattern in "get" tests, but there I think it should be okay, since there are no data changes with those.

@LaszloKecskes i was not sure at the time of the PR why would this be the cause but after thinking about it i think i have a clearer idea now, the key difference between the tests on the lines you mention and the fix is that all the others actually await for the test to finish (using the ```done``` parameter instead of ```await```, both are valid), the test i fixed was not awaiting in any way, setting up a race condition, i think it goes like this:

starts execution of not "awaiting test" and immediately start execution of the next beforeEach and next test. the beforeEach cleans the DB just before the previous test is requesting for entities, returns an empty array producing the error. and jest thinks is in the test being executed right now. The error happens in the async code from previous test but jest reports it as if it was failing in the current one, makes sense ?

this is a post explaining a similar issue dariacaraway.com/blog/jest-async-flakes

LaszloKecskes:
Yes, that make sense. I didn't know about the `done` option, sorry about that. Then it's all good!

LaszloKecskes (approved):
Approved after discussion.

PR #4179, opened 2022-01-26
Run oneup aggregation only when enable clasification is enabled

### LINKED ISSUE #4147: Wrap oneup review aggregations in a feature toggle conditional (5 pts)
The [QueryBuilder](https://github.com/huridocs/uwazi/blob/development/app/api/search/documentQueryBuilder.js) has a method to ask for the aggregations for an array of properties called [aggregations()](https://github.com/huridocs/uwazi/blob/development/app/api/search/documentQueryBuilder.js#L347). 

This method also asks for special aggregations used only in the oneup review, and adds them a `__` prefix to identify them, this is very costly since its run in every search even when not needed, even when the instance does not have this feature activated.

The proposed solution is to wrap this part in a conditional based on if the instance has this feature activated or not.

The feature togle is not in settings but in the thesauris as:
` "enable_classification" : true`

grafitto:
fixes #4147 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/search/search.js
```diff
@@ -683,7 +683,8 @@ const buildQuery = async (query, language, user, resources) => {
   queryBuilder.filterMetadata(filters);
   queryBuilder.customFilters(query.customFilters);
   // this is where the query aggregations are built
-  queryBuilder.aggregations(aggregations, dictionaries);
+  const isClasificationEnabled = await dictionariesModel.count({ enable_classification: true });
+  queryBuilder.aggregations(aggregations, dictionaries, isClasificationEnabled);
```
RafaPolit, line 687: This is the problem with comments: the comment that belonged to the `queryBuilder...` line, now is commenting the `isClasification...` (and incorrectly spelled!) line.

app/api/search/documentQueryBuilder.js
```diff
@@ -353,14 +353,16 @@ export default function () {
...
       // suggested has an implied '__' as a prefix
-      properties.forEach(property => {
-        baseQuery.aggregations.all.aggregations[`__${property.name}`] = propertyToAggregation(
-          property,
-          dictionaries,
-          baseQuery,
-          true
-        );
-      });
+      if (clasificationIsEnabled) {
```
RafaPolit, line 356: Again, this comment is commenting now an `if`.

app/api/search/specs/search.spec.js
```diff
@@ -578,12 +579,24 @@ describe('search', () => {
...
 
     describe('allAggregations', () => {
-      it('should return all aggregations', async () => {
+      it('should return all aggregations when clasification is not enabled', async () => {
         userFactory.mock(undefined);
         const allAggregations = await search.search({ allAggregations: true }, 'en');
         const aggregationsIncluded = Object.keys(allAggregations.aggregations.all);
         expect(aggregationsIncluded).toMatchSnapshot();
       });
+      it('should return all aggregations when clasification is enabled', async () => {
```
RafaPolit, line 588: Again, wrong spelling in classification.

RafaPolit (changes requested):
Every line has a spelling error, and the approach is not what is expected.

LaszloKecskes (approved):
Approved after pair review with Gabriel.

app/api/search/search.js
LaszloKecskes: Resolved in previous commit.

app/api/search/documentQueryBuilder.js
LaszloKecskes: Moved in previous commit.

app/api/search/specs/search.spec.js
LaszloKecskes: Approach and tests changed.

PR #4183, opened 2022-01-26
Sorts templates alphabetically

### LINKED ISSUE #3997: Request from partner to be able to order templates alphabetically (3 pts)
Partner asked if it is possible to organize their templates alphabetically as they do with the thesauri list.

grafitto:
fixes #3997 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Metadata/components/MetadataForm.js
```diff
@@ -26,6 +26,7 @@ const selectTemplateOptions = createSelector(
   templates => templates.map(tmpl => ({ label: tmpl.get('name'), value: tmpl.get('_id') }))
 );
 
+// eslint-disable-next-line import/exports-last
```
fnocetti, line 29: Please do not add eslint disable directives.

app/react/Settings/components/EntityTypesList.js
```diff
@@ -13,6 +13,7 @@ import { Icon } from 'UI';
 import { notificationActions } from 'app/Notifications';
 import Tip from '../../Layout/Tip';
 
+// eslint-disable-next-line import/exports-last
```
fnocetti, line 16: Same, remove please.

app/react/Settings/components/EntityTypesList.js
```diff
@@ -72,14 +73,21 @@ export class EntityTypesList extends Component {
     );
   }
 
+  // eslint-disable-next-line class-methods-use-this
```
fnocetti, line 76: What about changing this method to something like `getSortedTemplates()` and instead of accepting a `templates` parameter, you just use `this.props.templates` right in the function body? That way we get rid of this eslint error and also make it more cohesive with the class.

app/react/Settings/components/EntityTypesList.js
```diff
@@ -72,14 +73,21 @@ export class EntityTypesList extends Component {
...
+  sortTemplates(templates) {
+    const def = templates.filter(t => t.default);
+    const sorted = templates.filter(t => !t.default).sort((a, b) => a.name > b.name);
```
fnocetti, line 79: I think we agreed that the default template wasn't going to be treated specially, but just sorting it as well, didn't we?

app/react/Metadata/components/MetadataForm.js
```diff
@@ -50,6 +51,15 @@ export class MetadataForm extends Component {
 
   renderTemplateSelect(templateOptions, template) {
     if (templateOptions.size) {
+      const templateId = template.get('_id');
+      let sortedTemplates = templateOptions
+        .toJS()
+        .filter(defTemp => defTemp.value !== templateId)
```
fnocetti, line 57: I think that probably even just `temp` is better than `defTemp`. In this expression you are testing against an arbitrary template from the templates list, not against a default template.

app/react/Metadata/components/MetadataForm.js
```diff
@@ -50,6 +51,15 @@ export class MetadataForm extends Component {
...
+        .sort((a, b) => a.label > b.label);
+
+      const defTemplate = templateOptions.toJS().filter(temp => temp.value === templateId);
+      sortedTemplates = [...defTemplate, ...sortedTemplates];
```
fnocetti, line 61: Same question as in the other case: did we decide to treat the selected template specially?

fnocetti (changes requested):
@grafitto I left some comments, please review.

app/react/Metadata/components/MetadataForm.js
grafitto: Removed

app/react/Settings/components/EntityTypesList.js
grafitto: Removed

app/react/Settings/components/EntityTypesList.js
grafitto: Done

app/react/Settings/components/EntityTypesList.js
grafitto: I might not have catched that, changed

app/react/Settings/components/EntityTypesList.js
grafitto: Removed

app/react/Metadata/components/MetadataForm.js
grafitto: removed

app/react/Metadata/components/MetadataForm.js
grafitto: Removed

grafitto:
> @grafitto I left some comments, please review.

@fnocetti looked at your comments

fnocetti (changes requested):
@grafitto The implementation is not working. The `Array.sort`'s comparator callback expects a return type ` -1 | 0 | 1` and your callback is returning `boolean`.
This is an example of how it should be done: https://github.com/huridocs/uwazi/blob/df450c6c0fbc83542e724117f9613deaefbc30c0/app/react/Charts/utils/arrayUtils.js#L5

Please fix. Also, please add some testing.

fnocetti (changes requested):
It looks good to me. The functionality works as expected. Please take a look at my comment.

app/react/Metadata/components/specs/MetadataForm.spec.js
```diff
@@ -97,6 +97,25 @@ describe('MetadataForm', () => {
...
+        label: 'Zues',
+        value: 'zues',
+      },
+      { label: 'Yezzy', value: 'yezzy' },
+      { label: 'Template1', value: 'template1' },
+      { label: 'Aaron', value: 'Aaron' },
+    ]);
+    render();
+    const templateDropdown = component.find(SimpleSelect);
+    // console.log(component.debug());
```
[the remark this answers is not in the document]
grafitto: Removed

fnocetti (approved):

PR #4247, opened 2022-02-07
Update metadata on template update only when needed

### LINKED ISSUE #4230: Assigning a page to a template takes an unreasonable amount of time
When using pages for entity template rendering the configuration of the page is saved in the template settings. Still it takes a long time to save this change. Probably entities are being reprocessed and re-indexed without need.

grafitto:
fixes #4230 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/templates/templates.ts
+++ b/app/api/templates/templates.ts
@@ -19,6 +19,7 @@ import {
   denormalizeInheritedProperties,
   updateExtractedMetadataProperties,
 } from './utils';
+import { compareShallowProperties } from './reindex';
 
 const removePropsWithNonexistentId = async (nonexistentId: string) => {
   const relatedTemplates = await model.get({ 'properties.content': nonexistentId });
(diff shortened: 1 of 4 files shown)


app/api/templates/templates.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-statements */
```
Zasa-san, line 1: I think it's ok to move the `shouldUpdateMetadata` logic to an outside function, to avoid this.

app/api/templates/templates.ts
```diff
@@ -174,10 +176,25 @@ export default {
...
+    );
+    const shouldUpdateMetadata = changedProperties.find(
+      (key: string) =>
+        (key.startsWith('properties') &&
+          key !== 'properties.style' &&
+          key !== 'properties.nestedProperties' &&
+          key !== 'properties.fullWidth') ||
+        key.startsWith('commonProperties')
+    );
+    if (shouldUpdateMetadata || reindex) {
```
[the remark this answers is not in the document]
grafitto: > If `shouldUpdateMetadata` is true, but `reindex` is false, this condition will pass as true, but `updateMetadataProperties` will get called with `reindex = false`. Is this intended?

Yes that is intended. In some cases, update metadata is needed without necessarily having to reindex. That's how it was previously.

app/api/templates/templates.ts
grafitto: Removed

PR #4290, opened 2022-02-14
4227 - Search return correct aggregations for templates

### LINKED ISSUE #4227: Search is not computing the correct aggregations per template (5 pts)
When filtering by template search should be displaying the amount of entities for other templates. This filters is a special case and does not work as the rest of the aggregations. Instead it gives 0 for other templates:

image

Zasa-san:
fixes #4227 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti (approved):
This LGTM.

PR #4302, opened 2022-02-15
Sentry tweaks

elreplicante:
- Mongoose integration
- Capture errors only on the 5xx response codes.

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/server.js
+++ b/app/server.js
@@ -18,6 +18,7 @@ import { DistributedLoop } from 'api/services/tasksmanager/DistributedLoop';
 
 import { appContextMiddleware } from 'api/utils/appContextMiddleware';
 import { requestIdMiddleware } from 'api/utils/requestIdMiddleware';
+import Error from 'api/utils/Error';
 import uwaziMessage from '../message';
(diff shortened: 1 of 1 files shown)


app/server.js
```diff
@@ -115,7 +118,13 @@ DB.connect(config.DBHOST, dbAuth).then(async () => {
   apiRoutes(app, http);
   serverRenderingRoutes(app);
   if (config.sentry.dsn) {
-    app.use(Sentry.Handlers.errorHandler());
+    app.use(
+      Sentry.Handlers.errorHandler({
+        shouldHandleError(error) {
+          return error.status >= 500;
```
daneryl, line 124: @elreplicante We do not have status in our errors i think ? there is two type of errors we want to handle that i can think of, either is one of our errors which has ```code``` property not status, the other one is an instance of Error, something like:
```return error instanceof Error || error.code >= 500```

PR #4339, opened 2022-02-23
leaflet library

### LINKED ISSUE #4249: Spike Leaflet map [timebox 3 days]
For some of our partners the quality of the tiles of our current map provider is not good enough, it may happen also that we want to support more than one tile provider for out maps, after a brief spike by @txau we want to test a different map library ([Leaflet](https://leafletjs.com/)).

- Do an spike to replace our current map library, react-map-gl for leaflet
- Take into account things like markers, clustering markers etc... 
- Check how easily will be to switch and support different tile providers in Uwazi

mfacar:
fixes #4249

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


mfacar:
@konzz after your review I added some code for translations and removed other unreachable code lines.
The template label has the set color now.

PR #4404, opened 2022-03-15
Adds conditional display on pages

### LINKED ISSUE #4231: Be able to condition page templates to a particular value of the entity (8 pts)
This would allow to use different templates depending on the actual content and a more fine grained control over custom pages for templates.

Another approach that could be even better is to have conditional blocks as page tags. Something like:
```
<renderIf property="some property name" has-value="some expected value">
  Other code and elements here
</renderIf>

### PR DESCRIPTION
**Pull request by grafitto**
fixes #4231 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### CODE REVIEW

**Comment by konzz** [comment]
### EntitySection component
Sometimes you want to display a section of the entity based on whether a given condition is met.

Examples:

You want to display the name of a person if they are older than 18 years old.
A simple implementation would look like this:
```html
<EntitySection show-if='{"metadata.age": { "$gt": 18 }}'>
  <EntityData label-of="name" />: <EntityData value-of="name" />
</Entitysection>
```
you want to only show the name of a person if the name exists in the entity metadata
```html
<EntitySection show-if='{"metadata.name": { "$exists": true }}'>
  <EntityData label-of="name" />: <EntityData value-of="name" />
</Entitysection>
```

You want to show the name of a person if it matches value `John` in the entity metadata
```html
<EntitySection show-if='{"metadata.name": "John"}'>
  <EntityData label-of="name" />: <EntityData value-of="name" />
</Entitysection>
```
Inherited properties:
```html
<EntitySection show-if='{"metadata.inherited_text": { "$in": ["something"] }}'>
  <EntityData label-of="name" />: <EntityData value-of="name" />
</Entitysection>
```
Note: `show-if` value is a JSON string and the query should follow query string specified by [sift](https://github.com/crcn/sift.js)

The metadata is unwrapped into a JSON object with keys as the property names and values as the property values.
```js
metadata: {
  name: "John",
  age: 18
}

// Note that if the metadata property name has spaces, they are replaced with underscores and uppercase letters are converted to lowercase letters
// Eg: a metadata property name of "First Name" will be converted to "first_name" therefore:
metadata: {
  "first_name": "John",
  age: 18
}

```
Some values takes the following format, this includes inherited values and values like multiselect:
```js
metadata: {
    inherited_text: ['text value'], //Can contain more than one values depending on entities inherited from
    inherited_multiselect: ['option 1', 'option 2'],  
    inherited_geolocation: [{ lat: 23, lon: 12}]
}
```

Note that some values like Date added and Date modified are replaced with `creationDate` and `editDate` respectively and also are presented as timestamps. So, to use them, you can do this:

```html
<EntitySection show-if='{"editDate": 12434564323 }'>
  <EntityData label-of="name" />: <EntityData value-of="name" /> // This only shows if the entity
</Entitysection>
```

txau:
@konzz the documentation looks neat. I would only add as a reference more complex example, ie. based on the value of an inheritance with multiple inherited values.

konzz:
Thanks, it was @grafitto  driving, he has the merit here 🥇

PR #4510, opened 2022-04-06
3039 - 404 when documents in different languages have a different amount of pages

### LINKED ISSUE #3039: 404 when documents in different languages have a different amount of pages (3 pts)
Documents vary in the amount of pages they have across translations. In Uwazi, you can switch from one language to another while being in the URL for a particular page, ie:
- en/entity/hcvrcury1oj?page=9

When switching to another language while in that URL, if the target document has less pages Uwazi will toss a 404.

I guess this can easily be detected and fixed.

Zasa-san:
fixes #3039

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/I18N/components/I18NMenu.js
+++ b/app/react/I18N/components/I18NMenu.js
@@ -17,6 +17,10 @@ class I18NMenu extends Component {
     const { languages, locale, location, i18nmode, toggleInlineEdit } = this.props;
 
     let path = location.pathname;
+    if (location.search.match(/page=/)) {
+      const cleanSearch = location.search.split(/page=\d+/).join();
+      location.search = cleanSearch;
+    }
     const regexp = new RegExp(`^/?${locale}/|^/?${locale}$`);
     path = path.replace(regexp, '/');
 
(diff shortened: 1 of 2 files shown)


app/react/I18N/components/I18NMenu.js
```diff
@@ -15,8 +15,12 @@ class I18NMenu extends Component {
 
   render() {
     const { languages, locale, location, i18nmode, toggleInlineEdit } = this.props;
-
     let path = location.pathname;
+
+    if (location.search.match(/page=/)) {
```
RafaPolit, line 20: Are you sure we don't use "page" in other type of navigation properties like pagination?  Please check that as this could really affect this approach.

Zasa-san: I understand this component I18NMenu is only used to construct the language navigation menu, so other uses of the page query should not be affected.

Edit: If you mean the pagination of, for example, the IX suggestions table, then no. That pagination does not use url parameters.

RafaPolit: Yes, pagination for any URL that could get affected by this regexp: 
- IX is one good example
- activity log
- library "load more"

This that I can "think of" don't use the Page, so it seems like it should work without affecting other places.  Maybe we can brainstorm other places that could use this?

PR #4535, opened 2022-04-14
🔥 [HOTFIX]🔥 Fixed display issue on multiple relationships

### LINKED ISSUE #4528: PR #4417 broke relationship display of more than one value in metadata
Pull Request #4417 broke the rendering of Relationship links for properties that have more than one value.

Previously, relationships went through the `default` in `Metadata.js`.  This process took into account that values should be rendered with `<ValueList>` with interposed a comma between values when there is more than one.

This is no longer used and values with more than one relationship are rendered next to each other with no separation.  This should have been an easy catch if the dev and QA tests would have been done with more than one value.

cc @grafitto

grafitto:
fixes #4528 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/App/scss/modules/_metadata-view.scss
+++ b/app/react/App/scss/modules/_metadata-view.scss
@@ -50,6 +50,19 @@
           content: ',';
         }
       }
+
+      button.link-button {
+        background: none !important;
+        border: none;
+        padding: 0 !important;
+        color: #2b56c1;
+        cursor: pointer;
+        background-color: transparent;
+        
+        &:hover {
+          text-decoration: underline;
+        }
+      }
     }
     a {
       color: $c-primary;
--- a/app/react/App/scss/modules/_metadata-view.scss
+++ b/app/react/App/scss/modules/_metadata-view.scss
@@ -55,7 +55,7 @@
         background: none !important;
         border: none;
         padding: 0 !important;
-        color: #2b56c1;
+        color: $c-primary;
         cursor: pointer;
         background-color: transparent;
         
--- a/app/react/Metadata/components/Metadata.js
+++ b/app/react/Metadata/components/Metadata.js
@@ -11,7 +11,6 @@ import ValueList from './ValueList';
 const renderRelationshipLinks = linksProp => {
   const formattedLinkValues = Array.isArray(linksProp.value) ? linksProp.value : [linksProp.value];
   const hydratedValues = formattedLinkValues.map(linkValue => ({
-    ...linkValue,
     value: <RelationshipLink propValue={linkValue} />,
   }));
   const hydratedProp = { ...linksProp, value: hydratedValues };
--- a/app/react/Metadata/components/Metadata.js
+++ b/app/react/Metadata/components/Metadata.js
@@ -1,3 +1,4 @@
+/* eslint-disable import/exports-last */
 import React from 'react';
 import PropTypes from 'prop-types';
 import { t } from 'app/I18N';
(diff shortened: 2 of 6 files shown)


app/react/Metadata/components/RelationshipLink.tsx
```diff
@@ -27,26 +27,25 @@ const connector = connect(mapStateToProps, mapDispatchToProps);
...
-      );
-    }
+const RelationshipLink = ({ propValue: propVal, selectConnection }: mappedProps) => {
+  if (propVal.relatedEntity) {
     return (
-      <I18NLink key={prop.url} to={prop.url}>
-        {prop.icon && <Icon className="item-icon" data={prop.icon} />}
-        {prop.value}
-      </I18NLink>
+      <a
```
RafaPolit: This is not correct.  `a` tags should be links.  If you are only looking for an onClick event, this should be a button.

app/react/Metadata/components/specs/__snapshots__/Metadata.spec.js.snap
```diff
@@ -225,15 +240,34 @@ exports[`Metadata should render links when the property has url 1`] = `
...
+                "5": "g",
+                "6": " ",
+                "7": "v",
+                "8": "a",
+                "9": "l",
+                "value": <Connect(RelationshipLink)
+                  propValue="string value"
+                />,
+              },
+            ],
```
RafaPolit: This cannot be correct.  Either you are passing the wrong value to the test, or this means that you are passing each letter of a string like the array values.  Please either fix the text or the implementation.

app/react/Metadata/components/specs/__snapshots__/Metadata.spec.js.snap
```diff
@@ -266,7 +301,25 @@ exports[`Metadata should render links with icons if propery has url and icon 1`]
...
+                "3": "i",
+                "4": "n",
+                "5": "g",
+                "6": " ",
+                "7": "v",
+                "8": "a",
+                "9": "l",
+                "value": <Connect(RelationshipLink)
+                  propValue="string value"
+                />,
```
RafaPolit: Same as above... why is the string being made into an array?

app/react/Metadata/components/Metadata.js
```diff
@@ -7,6 +8,16 @@ import GeolocationViewer from './GeolocationViewer';
 import { RelationshipLink } from './RelationshipLink';
 import ValueList from './ValueList';
 
+const renderRelationshipLinks = linksProp => {
+  const formattedLinkValues = Array.isArray(linksProp.value) ? linksProp.value : [linksProp.value];
```
RafaPolit, line 12: Can the value in relationships really be something that is not an array? I thought single values were coming as arrays nonetheless.  Can you confirm this please?

app/react/Metadata/components/RelationshipLink.tsx
grafitto: Replaced

app/react/Metadata/components/specs/__snapshots__/Metadata.spec.js.snap
grafitto: Fixed

app/react/Metadata/components/specs/__snapshots__/Metadata.spec.js.snap
grafitto: Fixed

app/react/Metadata/components/Metadata.js
grafitto: In some cases, the value is a string, based on the existing tests

app/react/App/scss/modules/_metadata-view.scss
```diff
@@ -50,6 +50,19 @@
           content: ',';
         }
       }
+
+      button.link-button {
+        background: none !important;
+        border: none;
+        padding: 0 !important;
+        color: #2b56c1;
```
RafaPolit: I believe we have some sort of variables for link colors?

app/react/App/scss/modules/_metadata-view.scss
```diff
@@ -50,6 +50,19 @@
...
+        cursor: pointer;
+        background-color: transparent;
+        
+        &:hover {
+          text-decoration: underline;
+        }
+      }
```
RafaPolit: Maybe we already have a "button zeroing out" class somewhere?

grafitto: I couldn't find any

app/react/App/scss/modules/_metadata-view.scss
grafitto: Found

PR #4606, opened 2022-05-11
4369 thesauri entry duplication

### LINKED ISSUE #4369: Import CSV thesauri duplicates values when importing multiple times.
- create a blank thesauri
- import a CSV to fill it
- go back to thesaurus list and edit the same thesauri
- import again the same file
- values get duplicated

I am not sure this needs to be fixed ? but we already having problems with some partners probably because the lack of feedback, they end up with multiple duplication of the same value and this can lead to assign different internal ids when relating this values to an entity.

- should we not allow duplicates on thesauri values, what is the use case to allow this ?
- should we provide feedback to the user that the values are going to be duplicated ?
- maybe on the particular case of importing, do not allow duplication at all ?

LaszloKecskes:
fixes #4369 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti (approved):

app/api/migrations/migrations/78-resolve_duplicate_thesauri_entries/specs/78-resolve_duplicate_thesauri_entries.spec.js
```diff
@@ -0,0 +1,115 @@
+import testingDB from 'api/utils/testing_db';
```
fnocetti, line 1: It may be worth fixing the related code climate issue here

PR #4625, opened 2022-05-17
Default restricted to true for logged in users

### LINKED ISSUE #4361: Set default view for logged in user to see restricted files
I'm finding this is the desired default and I find myself clicking on the "restricted" over and over again.

RafaPolit:
fixes #4361

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Users/specs/Login.spec.js
```diff
@@ -81,12 +82,17 @@ describe('Login', () => {
     });
 
     describe('on response success', () => {
-      it('should reload thesauris and go to home', async () => {
+      it('should reload thesauris, set filters to include "restricted", and go to home', async () => {
```
konzz, line 85: should this be in a separated unit test? even tho is the same action ?

RafaPolit: Since the title of the test is "on response success" and the current unit test was a "reload and goto home", I thought it would be enough to combine them.  If we split them, we should probably also remove the "go home" part, as it wouldn't make sense to have a compound "reload thesauris and go to home", and then an "individual" `set filters`. 
 Do you really think it merits a new test?

PR #4640, opened 2022-05-19
correct relationship filter

### LINKED ISSUE #4530: Relationship values cannot be used as filters when user has no permissions
Consider this scenario:

- a collection has entities of the type "country"
- it also has entities of the type "case"
- "case" has a relationship property pointing to "country"

If an entity of the type "case" has as value a "country" for which my user has no permissions, I won't even see the option available in filters. This is wrong since the filter aggregations should be calculated based on the values of "case" not on the access control of "country".

LaszloKecskes:
fixes #4530

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti (approved):
The strategy looks good

PR #4656, opened 2022-05-23
translate title when structureChange is false

### LINKED ISSUE #4462: Changing a template name makes it not translatable.
To replicate:
 * Update template with a new name
 * Navigate to library and activate inline translation
 * The new name is the only one that does not get highlighted
 * Navigate to setting -> translations for that template
 * The old name is in the list of translations but not the new one

Updating the name again to the old one "fixes" the problem but this makes changing template names basically not an option if you have multiple languages.

Probably a migration is required.

daneryl:
fixes #4462 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/migrations/migrations/80-fix_template_translations/index.js
+++ b/app/api/migrations/migrations/80-fix_template_translations/index.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-restricted-syntax */
 /* eslint-disable no-await-in-loop */
 export default {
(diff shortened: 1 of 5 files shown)


LaszloKecskes (approved):
Looks good.

PR #4768, opened 2022-06-21
4403 map geolocation inheritance

### LINKED ISSUE #4403: Map is not displaying inherited non-public geolocations
When inheriting a geolocation property from a non-public entity, the map fails to render the markers if you don't have permissions.

Access control over inherited geolocations is supposed to be based on the *inheriting* entity. In fact, the entities display a map with the geolocation even if you are logged out. Nevertheless, the library map is not showing the markers for these inherited geolocations from a non-public entity.

This also happens if you are logged in with a user that does not have permissions for that information.

LaszloKecskes:
fixes #4403 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/search/search.js
```diff
@@ -600,40 +512,28 @@ const entityHasGeolocation = entity =>
...
+            if (_v.inheritedType === propertyTypes.geolocation) {
+              newValue.inherit_geolocation = (_v.inheritedValue || []).filter(iv => iv.value);
+            }
+            return newValue;
+          }),
+        ])
+      ),
+    }))
+    .filter(r => r.metadata && Object.keys(r.metadata).length);
+  results.rows = results.rows.filter(r => entityHasGeolocation(r));
```
RafaPolit: In the previous implementation, there was one iteration over rows.  

Here there is an iteration over the entire array in line 517.  Then a map and filter on 518.  Then again a new filter in 534.

I'm wondering if this impacts performance for very large array of results?  Also, can't they be somehow combined?, for example, can't line 534 also include and `&& entityHasGeolocation(r)` instead of having to iterate again over the entire array?

LaszloKecskes: I guess there is, but it should be just by a constant. Nevertheless, there's no reason to not do it, so I'll put 535 to 534. For 517, it still should still come first. If we're very concerned about performance here, we could do it in one iteration with a forEach and pushing the filtered results to a new array.

PR #4798, opened 2022-06-28
Aggregated similar results on metadata values.

### LINKED ISSUE #4706: A way to present repeated inherited values in a consolidated format (5 pts)
I've got a partner collection that has two main types of entities: Documents and Paragraphs. We need to be able to filter Documents by topic, but this information currently lives on the Paragraph entities. Any easy enough solution is to add a relationship property to the Document and configure it to inherit the topic information from the Paragraph entities. 

BUT in this particular collection, there could be 50+ Paragraph entities that are associated with a Document entity and there is always a lot of overlap among the topics on each Paragraph entity. As a result, this creates a problem: when viewing the entity, a user is going to be confronted with an absurdly long list of individual inherited values repeating over and over again - doing harm to the meaningfulness of the information. A small example: 

image

We need a way to be able to present this information in a consolidated format in order to make the information here meaningful and not drown users in a sea of repeated values (not dissimilar to how filters are presented in the Library sidebar, like below).

image

RafaPolit:
fixes #4706

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Metadata/components/ValueList.js
```diff
@@ -2,14 +2,25 @@ import React from 'react';
...
+  }
+  return data.value;
+};
+
 const withIcon = v =>
   v.icon ? (
     <>
       <Icon className="item-icon item-icon-center" data={v.icon} />
-      {v.value}
+      {composeCount(v, 'withIcon')}
```
Zasa-san: This is throwing a warning message in the console about repeated keys.

PR #4803, opened 2022-06-29
4779 - Responsive pdf text mode for mobile

### LINKED ISSUE #4779: Responsive text mode for mobile (3pts)
When consuming PDF documents in a mobile device, the screen is to small to properly render the PDF. Our current text mode has formatted text, that is line breaks try to keep the original document layout, which renders ok for big screens but it is problematic for small screens.

Ideas:
- Have yet another copy of the extracted texts without any format. This poses some overhead on the database as this information is redundant. This is the version that will be offered if a smaller screen is detected.
- Change our text mode to un-formatted text, even for big screens. The down side is that it can greatly affect readability.
- Wait for a PDF to responsive HTML convertor. We can research available open source libraries for this purpose or wait for the ML team to create their own one (we are close to this objective at the moment).

Zasa-san:
fixes #4779

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/App/scss/modules/_document.scss
```diff
@@ -31,7 +31,15 @@
...
 
 .document-viewer .main-wrapper {
-  pre.force-rtl {
+  .raw-text {
+    white-space: pre;
+
+    @media (max-width: 1023px) {
+      white-space: pre-line;
+    }
+  }
```
RafaPolit: I know that Uwazi has a mix of max-width and min-width media queries.  Still, for new developments, lets try to do it "Mobile first".  So, the "normal" behavior is pre-line and at min-width 1024px it starts using pre.

PR #4827, opened 2022-07-05
Calculate suggestion using getIXSuggestionState in client

### LINKED ISSUE #4811: IX front end state update is sometimes inconsistent
When a suggestion is accepted, or a related label is changed, the UI updates the state without sending things from the back end. This update is inconsistent with the actual calculated states. The same applies as in #4810: instead of fixing it locally, we should try to use the same calculation function to make it automatically consistent, if the categorization again changes in the future. This means refactoring `app/api/suggestions/getSuggestionState.ts - getState` into shared, if necessary.

grafitto:
fixes #4811 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/suggestions/suggestions.ts
+++ b/app/api/suggestions/suggestions.ts
@@ -209,6 +209,7 @@ export const Suggestions = {
           state: 1,
           page: 1,
           date: 1,
+          error: 1,
           labeledValue: 1,
         },
       },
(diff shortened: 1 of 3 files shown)


LaszloKecskes (approved):
Looks good.

PR #4844, opened 2022-07-07
4910 IX refactor default values

### LINKED ISSUE #4810: IX blank state generates inconsistent states
IX blank state generation is inconsistent with the current state calculation. Note: do not only fix locally, make sure that it routes through the same state calculation function, so if the categories change, it will be automatically consistent.

fnocetti:
fixes #4810 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/suggestions/configurationManager.ts
+++ b/app/api/suggestions/configurationManager.ts
@@ -101,9 +101,10 @@ const fetchEntitiesBatch = async (query: any, limit: number = 100) =>
 
 const fetchEntitiesSharedIds = async (
   template: ISettingsTemplate['template'],
-  defaultLanguage: string
+  defaultLanguage: string,
+  batchSize = 2000
 ) => {
-  const BATCH_SIZE = 2000;
+  const BATCH_SIZE = batchSize;
   let query: any = {
     template,
     language: defaultLanguage,
@@ -127,10 +128,15 @@ const fetchEntitiesSharedIds = async (
 
 const createDefaultSuggestions = async (
   settingsTemplates: ISettingsTemplate[],
-  defaultLanguage: string
+  defaultLanguage: string,
+  batchSize?: number
 ) => {
   const templatesPromises = settingsTemplates.map(async template => {
-    const entitiesSharedIds = await fetchEntitiesSharedIds(template.template, defaultLanguage);
+    const entitiesSharedIds = await fetchEntitiesSharedIds(
+      template.template,
+      defaultLanguage,
+      batchSize
+    );
 
     const fetchedFiles = await files.get(
       { entity: { $in: entitiesSharedIds }, type: 'document' },
--- a/app/api/suggestions/configurationManager.ts
+++ b/app/api/suggestions/configurationManager.ts
@@ -1,4 +1,3 @@
-/* eslint-disable max-statements */
 import entities from 'api/entities';
 import entitiesModel from 'api/entities/entitiesModel';
 import { files } from 'api/files/files';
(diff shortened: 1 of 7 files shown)


app/api/suggestions/updateState.ts
```diff
@@ -13,6 +13,8 @@ import {
   getLabeledValueStage,
 } from './pipelineStages';
 
+type SuggestionsAggregationResult = SuggestionValues & { _id: any; propertyName: string };
```
LaszloKecskes, line 16: What was the reason for taking these out?

app/api/suggestions/configurationManager.ts
```diff
@@ -0,0 +1,190 @@
...
+import { files } from 'api/files/files';
+import settings from 'api/settings/settings';
+import languages from 'shared/languages';
+import { ObjectIdSchema } from 'shared/types/commonTypes';
+import { FileType } from 'shared/types/fileType';
+import { IXSuggestionType } from 'shared/types/suggestionType';
+import { IXSuggestionsModel } from './IXSuggestionsModel';
+import { Suggestions } from './suggestions';
+
+interface ISettingsTemplate {
```
LaszloKecskes, line 12: Maybe we could make this part of the settings type to be consistent, and import there?

app/api/suggestions/specs/configurationManager.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+      },
+    ];
+
+    await createDefaultSuggestions(settings, 'en', 2);
+    const newSuggestions = await IXSuggestionsModel.get({
+      propertyName: 'title',
+    });
+
+    expect(newSuggestions.length).toBe(2);
+    expect(newSuggestions[0].state).toBe(SuggestionState.valueEmpty);
```
LaszloKecskes: Maybe two cases for testing labelEmpty and emptyEmpty as well, as the other two possible blank states? I accept if you think we shouldn't, since technically the calculation is already tested when we are testing the state update.

app/api/suggestions/suggestions.ts
```diff
@@ -228,19 +136,19 @@ export const Suggestions = {
 
   saveMultiple: async (_suggestions: IXSuggestionType[]) => {
     const toSave: IXSuggestionType[] = [];
-    const update: IXSuggestionType[] = [];
+    let doStateUpdate = false;
```
LaszloKecskes: Let's talk about this, I don't completely see the reason.

LaszloKecskes (changes requested):
Aside from the few points commented for discussion, looks good.

app/api/suggestions/updateState.ts
fnocetti: This was originally misdefined but was not evident until I tried some things (that finally removed).
The thing is: `SuggestionVales` is defined as the type of the input needed but the `getState` function, or what is needed to calculate the state.
The `_id` and `propertyName` were there because they were returned in the mongo query, but made no sense from the point of view of the submodule defining the interface. What makes sense is that we segregate the interfaces, and this query uses an extended version for it's purpose.

app/api/suggestions/configurationManager.ts
fnocetti: Resolving as discussed

app/api/suggestions/specs/configurationManager.spec.ts
fnocetti: Working on it

app/api/suggestions/suggestions.ts
fnocetti: Will rework this after discussion

LaszloKecskes (approved):
Looks good!

PR #4855, opened 2022-07-08
4432 show training status in ix dashboard

### LINKED ISSUE #4432: IX: show training data dashboard
Display atop of the table, before the column headers a mini-dashboard with:
- Amount of training samples (label match + label mismatch + label empty)
- Amount of suggestion slots (number of documents in each language)
- Amount of non-obsolete suggestions

Keep it updated in real time as the users label data or the model sends suggestions.

fnocetti:
fixes #4432 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/MetadataExtraction/specs/GridChart.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+ */
+import { render, RenderResult, within } from '@testing-library/react';
+import React from 'react';
+import { act } from 'react-dom/test-utils';
+import { GridChart } from '../GridChart';
+import * as containerWidthHook from '../useContainerWidthHook';
+
+jest.spyOn(containerWidthHook, 'useContainerWidth').mockReturnValue(50);
+
+const formatForAssertion = (items: HTMLElement[]) => [
```
LaszloKecskes, line 12: Can you explain this one?

app/react/MetadataExtraction/TrainingHealthDashboard.tsx
```diff
@@ -0,0 +1,50 @@
...
+const mapStats = (data: SuggestionsStats['data']) => [
+  {
+    color: '#5073CF',
+    value: data.labeledMatching,
+    label: { text: <Translate>Training</Translate> },
+    overlaying: true,
+  },
+  {
+    color: '#CD4C4C',
+    value: data.labeled,
```
LaszloKecskes, line 20: While trying it out, this category doesn't show up for me in the legend (but I see the color on visualization). Missing label? Or is it intentional?

app/api/suggestions/stats.ts
```diff
@@ -0,0 +1,78 @@
...
+        },
+      },
+    ])
+    .then(([result]) => result);
+
+const getStats = async (propertyName: string): Promise<SuggestionsStats> => {
+  const groups = await getGroups(propertyName);
+  const labeledMatching = addCountsOf(groups, [SuggestionState.labelMatch]);
+  const labeled = addCountsOf(groups, [
+    SuggestionState.labelMatch,
```
LaszloKecskes, line 50: Label match is counted in two categories. The visualization would suggest that these categories are distinct. Is this intentional?

LaszloKecskes (approved):
In general, I do not quite agree with the categorization. For example, Match / Label is not the only category that is used for training (see the state explanation in the question mark), so I feel that this is misleading at the moment. Also, there is a possible missing legend (see comments).

However, the code itself looks good, the functionality works, and it is implemented in a flexible way, so changing categories later can be done with minimal pain, therefore approved for now, so we can begin functional testing on the server, and do small changes later in an agile way.

Maybe we can consider leaving the branch alive and the issue open, until the final version is approved. But this is an administration detail and up to you.

app/react/MetadataExtraction/TrainingHealthDashboard.tsx
fnocetti: Is intentional per designs, tho I would agree in that showing it would be a good idea :)

app/react/MetadataExtraction/specs/GridChart.spec.tsx
fnocetti: Maybe needs refactoring then :D It's just a helper for the test. Because the chart is actually a list of divs, that are draw from left to right (first-up, first-down, second-up, second-down, etc...), this helper turns a list of html elements into a matrix that is easier to work with in the test. For it it just takes the elements in odd positions to be one of the rows, and the ones in even positions to be the other one, and then it just keeps the substring of the text-form of the element that contains the value.

app/api/suggestions/stats.ts
fnocetti: Yes, cause labelMatch is a subset of label[Match/Mismatch/Empty]

fnocetti:
> In general, I do not quite agree with the categorization. For example, Match / Label is not the only category that is used for training (see the state explanation in the question mark), so I feel that this is misleading at the moment. Also, there is a possible missing legend (see comments).

Let's discuss it! It's already known that we need to sit and give it better thought

> Maybe we can consider leaving the branch alive and the issue open, until the final version is approved. But this is an administration detail and up to you.

I'm merging now and we can open a new branch to implement further changes, thanks!

PR #4859, opened 2022-07-08
Sending entity value instead of labeled value for training

### LINKED ISSUE #4820: IX: sending potentially wrong values to the service
Theoretically, the process should send the values from the entity and the segmentation boxes from the file for training the model. As seen [here](https://github.com/huridocs/uwazi/blob/f3db5dbb752f0d373843777c8abb989262bd6ee7/app/api/services/informationextraction/InformationExtraction.ts#L122) the process currently sends both from the file. It is possible, that we want to edit the value manually, but keep the boxes: for example, the click-to-fill might pick up and parse a wrong date (or any other value) that needs to be corrected. In these cases the current implementation sends wrong values to the service for training.

The services should pick up the current value from the entity, instead of the file.

konzz:
fixes #4820 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/services/informationextraction/InformationExtraction.ts
```diff
@@ -119,7 +119,7 @@ class InformationExtraction {
           data = {
             ...data,
             language_iso: languages.get(file.language!, 'ISO639_1') || defaultTrainingLanguage,
-            label_text: propertyLabeledData.selection?.text,
+            label_text: file.label_text,
```
LaszloKecskes: Minor semantic thing: could we name label_text something else, which shows that this is the value from the entity? entity_label_text, entity_value_text, etc.

app/api/services/informationextraction/getFiles.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
```
LaszloKecskes, line 1: Extra eslint ignore.

app/api/services/informationextraction/getFiles.ts
```diff
@@ -16,13 +17,15 @@ interface FileWithAggregation {
   entity: string;
   language: string;
   extractedMetadata: ExtractedMetadataSchema[];
+  label_text: string;
```
LaszloKecskes: If this is the value from the entity, this is not necessarily just a string, right? For example, for dates and numbers it is a number.

app/api/services/informationextraction/getFiles.ts
```diff
@@ -16,13 +17,15 @@ interface FileWithAggregation {
...
 }
 
 type FileEnforcedNotUndefined = {
   _id: ObjectIdSchema;
   filename: string;
   language: string;
   entity: string;
+  label_text?: string;
```
LaszloKecskes: Same as in line 20.

app/api/services/informationextraction/getFiles.ts
```diff
@@ -71,7 +75,17 @@ async function getFilesForTraining(templates: ObjectIdSchema[], property: string
...
 
-  return getFilesWithAggregations(files);
+  const filesWithEntityValue = files.map(file => {
+    const entity = entities.find(e => e.sharedId === file.entity);
+    if (!entity?.metadata || !entity?.metadata[property]?.length) {
+      return file;
+    }
+    const [{ value }] = entity.metadata[property] || [{}];
+
+    return { ...file, label_text: value };
```
LaszloKecskes: Due to the type mismatch between the entity values and what the service expect, we should probably transform this for dates into the actual string the service expects. If we do it here, immediately when getting it from the entity, we can leave label_text as string after all.

app/api/services/informationextraction/specs/InformationExtraction.spec.ts
```diff
@@ -121,13 +121,13 @@ describe('InformationExtraction', () => {
             width: 457,
             height: 15,
             page_number: 1,
-            text: 'something something',
+            text: 'something',
```
LaszloKecskes, line 126: Probably another test case with a date type would be nice, because of the already mentioned type reasons.

app/api/services/informationextraction/getFiles.ts
```diff
@@ -71,7 +75,17 @@ async function getFilesForTraining(templates: ObjectIdSchema[], property: string
     { limit: MAX_TRAINING_FILES_NUMBER }
   )) as (FileType & FileEnforcedNotUndefined)[];
 
-  return getFilesWithAggregations(files);
+  const filesWithEntityValue = files.map(file => {
+    const entity = entities.find(e => e.sharedId === file.entity);
```
LaszloKecskes: This goes through every entity and every related file, which can be a lot, so this is potentially slow (find inside map, running time scales with (number_of_entites*number_of_files). Consider pre-processing the entities, mapping the values, so here instead of the find it is just picking something from a map or object.
The indexObject function in shared/dataUtils is for these cases.

LaszloKecskes:
+1 without line: if I see correctly, this was branched from development, not from the feature branch. We should probably rebase this to the feature branch, since we were changing things a lot, and those changes are not yet in development.

LaszloKecskes (changes requested):
There are some necessary changes, especially the type issues.

Issue #4248, opened 2022-02-07
Document stuck in "processing..."


txau:
Find out if this is caused by a bug. The document seems to have been processed alright.

(CEJIL)
![image](https://user-images.githubusercontent.com/489831/152798941-e51e7c74-a27d-4350-aece-67bee7051c09.png)

txau:
Apparently this is a recurring issue that could not be replicated in the past: https://github.com/huridocs/uwazi/issues/3148

Zasa-san:
Could happen if the processing finishes while the users is not on the library. This seems to be related to the store not being properly updated. 
A simple way to reproduce is:
- Adding a delay to `app/api/files/processDocument.ts`
- Uploading a file
- Navigating to settings or an entity view
- Returning to the library

The document will display as ready, but the card and upload button will remain in `processing`.

Another scenario involves reloading or navigating when the file is being processed, an the processing of the file fails.

These statuses for the button and the card are being updated by a socket with the listeners are in the components, they might be unmounted when the emit happens. This means that the functions that update the store may not trigger as well, since the component missed the socket event.

txau:
@Zasa-san if I am understanding correctly, the server side process won't update the document status if the user does not receive the event via socket?

Zasa-san:
@txau Yes. For example:
https://github.com/huridocs/uwazi/blob/18db7a95e2766ab266244aafac6116b1c712dc0c/app/react/Metadata/components/UploadButton.js#L38-L40

Those actions update the store, but the component might not be mounted when the socket emits.

txau:
@Zasa-san my question is if the document remains as "Processing" in the database, because the problem here is not with the UI, the document appears as processing with a fresh session as well.

Zasa-san:
@txau That, I was not able to reproduce. I've not gotten a document stuck in `processing`in the DB.
The original issue https://github.com/huridocs/uwazi/issues/3148 state that a refresh fixes the problem, that's indicating that the document is not stuck in processing in the DB. Are there instances where even a refresh doesn't fix the problem?
https://github.com/huridocs/uwazi/pull/4884 Should fix the UI side of things that prevents the UI from updating.

txau:
here you have an example:  https://summa.cejil.org/en/library/?q=(allAggregations:!f,filters:(),from:0,includeUnpublished:!f,limit:30,order:desc,sort:creationDate,types:!(%275a4d294c79f3f44b101e2816%27),unpublished:!f)

[...]

PR #4884, opened 2022-07-12
4378 - Accept main document on entity creation

### LINKED ISSUE #4378: Accept main Document on entity creation
We should accept the main PDF on entity creation, just like we do with supporting files.

mfacar:
fixes #4378
Addresses some UI issues discovered in #4248

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/entities/managerFunctions.ts
```diff
@@ -0,0 +1,219 @@
...
+) => {
+  const deletedFiles = entityFiles.filter(
+    existingFile =>
+      existingFile._id &&
+      existingFile.type === type &&
+      !entity[type === 'attachment' ? 'attachments' : 'documents']?.find(
+        attachment => attachment._id?.toString() === existingFile._id.toString()
+      )
+  );
+  await Promise.all(
```
LaszloKecskes: filesApi.delete works with a query, consider building one query (out of the _id's, perhaps?) instead of a Promise.all, to have less db calls.

app/api/entities/managerFunctions.ts
```diff
@@ -0,0 +1,219 @@
...
+      } catch (e) {
+        errorLog.error(prettifyError(e));
+        saveResults.push(`Could not save file/s: ${file.originalname}`);
+      }
+    })
+  );
+
+  if (documentsToProcess.length) {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    Promise.allSettled(
```
LaszloKecskes, line 193: Since there is an eslint disable for it, I imagine that is the case, but let me double check it: this is intentionally not awaited? If yes, why? I'm asking since every other usage of the processDocument function I could find is awaited.

app/api/entities/managerFunctions.ts
```diff
@@ -0,0 +1,219 @@
...
+  const newUrls = entity.attachments?.filter(attachment => !attachment._id && attachment.url);
+
+  if (newAttachments.length) {
+    await Promise.all(
+      newAttachments.map(async (file: FileType) => {
+        const savedFile = await storeFile(attachmentsPath, file, true);
+        attachments.push({
+          ...savedFile,
+          entity: updatedEntity.sharedId,
+          type: 'attachment',
```
LaszloKecskes: In general, for this file: would it make sense to lift out the file type enum in `fileSchema.ts` to be importable separately, so we can use that in cases like this, instead of the hard-coded strings? If you agree, but you think this pr is big enough as it is (I think so), we can also add a tech debt issue for that.

app/api/entities/specs/entitySavingManager.spec.ts
```diff
@@ -329,13 +373,197 @@ describe('entitySavingManager', () => {
...
+            _id: entity3Id,
+            sharedId: 'shared3',
+            title: 'entity3',
+            template: template1Id,
+            attachments: [entity3textFile],
+          };
+
+          expectedDocuments = [];
+
+          ({ entity: savedEntity } = await saveEntity(entity, {
```
LaszloKecskes, line 475: Should this have the emitter as well, or is that redundant? Same question for the next text.

LaszloKecskes (commented):
I have only reviewed the API part and did some minimal testing. 
Looks good! There are some small questions, I consider the one about the "await" the only important one. Feel free to do what you think is best for the rest.

app/api/entities/managerFunctions.ts
mfacar: That's right @LaszloKecskes, we are intentionally leaving the promise floating because we don't want the request to wait until the documents are processed because that could increase significative the response time at 'saving entities' when there are several documents to save. We are updating the state of the documents through the socket when they are completed. We agreed that behavior in a standup meeting after the problem was discussed. The main difference with other places where we are waiting for processDocument function is that there we are processing only one document at a time.

app/api/entities/managerFunctions.ts
mfacar: Nice catch! I'll check it

app/api/entities/managerFunctions.ts
mfacar: It makes total sense 👍 , I'll do it here since it's not a big change

app/api/entities/specs/entitySavingManager.spec.ts
mfacar: We are using an empty function as the default value for the emitter, but yes, to be consistent in all the calls it should be the mocked function. 👍

app/api/entities/managerFunctions.ts
LaszloKecskes: Thanks for the explanation!

PR #4914, opened 2022-07-20
Webpack upgrade

### LINKED ISSUE #4096: Update Babel and Webpack
Currently, on the github repository we have several warnings about packages by dependabot. A large portion of those are part of the internal dependency chains of babel and webpack. In order to resolve the issues, we need to update. 
Before attempting the fix for the issue, some further discussion will be necessary about how to handle the changes; the initial view was that it would involve rewriting our babel and webpack configs, scripts, etc. from scratch, rather than attempting to update incrementally.

elreplicante:
fixes #4096 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


daneryl:
@daneryl check potential css problems in the login page, maybe the same issue that has been always there when you are in a blank state ? if that is the case, merge.

PR #4915, opened 2022-07-20
3046 csv import nested thesauri

### LINKED ISSUE #3046: Allow groupings on CSV Import of thesauri
Since we allow grouping or 'nesting' of thesaurus values, we should implement a way to achieve that structure via CSV Import.

fnocetti:
fixes #3046 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/csv/specs/csvLoaderThesauri.spec.js
```diff
@@ -118,5 +118,136 @@ describe('csvLoader thesauri', () => {
       ]);
       mockedFile.mockRestore();
     });
+
+    describe('nesting', () => {
```
LaszloKecskes, line 121: Please consider adding one test case, which checks the value of translations.

app/api/thesauri/thesauri.js
```diff
@@ -135,6 +135,22 @@ const update = async thesauri => {
...
 
+function recursivelyAppendValues(originalValues, newValues) {
+  const values = [...originalValues];
+  const existingLabels = new Set(values.map(v => v.label));
+
+  newValues.forEach(newValue => {
+    if (!existingLabels.has(newValue.label)) {
+      values.push(newValue);
+    } else if (newValue.values) {
+      const originalValue = values.find(v => v.label === newValue.label);
```
LaszloKecskes: Find inside a foreach. I don't know how big are our thesauri's, but I have been surprised the past. Consider mapping the original labels by value before the forEach.

app/api/csv/importThesauri.ts
```diff
@@ -0,0 +1,113 @@
...
+import { LanguageSchema } from 'shared/types/commonTypes';
+import { ThesaurusValueSchema } from 'shared/types/thesaurusType';
+import { Readable } from 'stream';
+import { CSVRow } from './csv';
+
+type ParsedValue = { nested: boolean; value: string };
+type ParsedRow = Record<string, ParsedValue>;
+
+const buildThesauri = (rows: ParsedRow[], languageLabel: string) =>
+  rows.reduce<ThesaurusValueSchema[]>((thesauriValues, row) => {
```
LaszloKecskes: Using reduce to build an array. The same concern as before about the thesaurus length we might have. Consider using a different approach. This seems complicated, so probably a loop.

Also, the same concern about using concat: it also copies every existing thing again, to return a new array. After resolving the loop, consider using an in-place modification, like push.

LaszloKecskes (changes requested):
The functionality works great, but there is some discussion needed about the implementation.

app/api/csv/importThesauri.ts
fnocetti: makes sense

app/api/csv/specs/csvLoaderThesauri.spec.js
fnocetti: sure

app/api/thesauri/thesauri.js
fnocetti: makes sense as well

app/api/csv/specs/csvLoaderThesauri.spec.js
fnocetti: done

app/api/thesauri/thesauri.js
fnocetti: done

app/api/csv/importThesauri.ts
fnocetti: done

LaszloKecskes (approved):
Every request is addressed, perfect. Thanks!

PR #4935, opened 2022-07-26
Delete obsolete mongo index entities - template_1

### LINKED ISSUE #4908: Deleting obsolete mongodb indexes
In #4276, the solution was to extend a mongodb entity index into a composite (template -> template_language_published). We do this through mongoose, and mongoose automatically adds the new index on database access. However, mongoose does not automatically remove obsolete indices. One possible justification is that one might have indices set up outside mongoose, and the existing sync function removes those (i.e. it syncs the indices to be exactly the same as they are on the schema).

Possible options to resolve this:
- if we do not (or do not want to) entirely handle indices through mongoose, write explicit migrations to remove indices, when we know we do not need them any more - in this case, in this issue the task is to complete #4276 by removing the old index
- if we are okay to handle things in mongoose, ensure that the sync function is called for each schema on startup (server.js) and for tests. This might be somewhat tricky to do.
- we can also decide to leave this to the back end rewrite, where we are planning to drop mongoose anyway, and live with an obsolete index for now - in this case, close this issue with a comment

LaszloKecskes:
fixes #4908 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti (approved):
LGTM and also added another index to the migration. Feel free to merge

PR #4968, opened 2022-08-04
3378 bulk relationship save

### LINKED ISSUE #3378: Bulk relationship save grows exponentially
When saving relationships in bulk, the amount of time needed to process the request grows exponentially as we add more relationships to the bulk. This is probably due to a blocking process.

LaszloKecskes:
fixes #3378 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/relationships/specs/relationships.spec.js
```diff
@@ -398,6 +399,182 @@ describe('relationships', () => {
...
+            entityData: entity3Expected,
+          },
+          {
+            _id: expect.anything(),
+            entity: 'entity4',
+            hub: expect.anything(),
+            entityData: entity4Expected,
+          },
+        ]);
+        expect(result[2].entityData.hub).toEqual(result[3].entityData.hub);
```
fnocetti, line 496: It feels wrong that we assume that connections sent outside a hub (group) are to be within the same hub.
I thought that connections outside a hub should always have a hub because in the bulk they represent an update to a hub (adding a new connection), isn't that true?

fnocetti (approved):
Added a comment regarding the definition of the API.
Anyway I think the development is correct and a very good job!

app/api/relationships/specs/relationships.spec.js
LaszloKecskes: This assumption only applies when you are creating a new hub. So, what we assume that in each subgroup:
- all elements have a hub or all elements do not have a hub
- if they all have a hub, they can belong to any hub, even mixed - I have changed the first test to show this option
- if they all do not have a hub, we assume that they are to belong to a new hub - you see this check in the test

Technically, you could mix things and say that the "non-hubbed" relationships will still go to the same new hub, but when we did the main relationship save optimization, we have not found any use case for than in Uwazi, and we have opted for the minimal solution that covers the existing cases, with an error message for safety.

fnocetti: Forgot to say it LGTM :D Thanks for the explanation

e2e/helpers/connection-actions.ts
```diff
@@ -0,0 +1,162 @@
+import { ElementHandle } from 'puppeteer';
```
RafaPolit: Please, rename this file to `relationship-actions.ts`, and all occurrences of the word `connection` to `relationship` .  We are deprecating the term `connections`.  And while it still exists in many places (most importantly the DB), let's try to introduce the newly "accepted" term to new development.

e2e/suite1/connections-actions.test.ts
```diff
@@ -0,0 +1,295 @@
...
+    await proxyMock();
+    await adminLogin();
+    await disableTransitions();
+  });
+
+  afterAll(async () => {
+    await logout();
+  });
+
+  it('should navigate to an entities connections view', async () => {
```
RafaPolit: this should be `entity's connections view`, not `entities`.

e2e/suite1/connections-actions.test.ts
```diff
@@ -0,0 +1,295 @@
...
+    await expect(page).toClick(
+      '#filtersForm > div:nth-child(2) > ul > li > ul > li:nth-child(6) > label > span.multiselectItem-name > span'
+    );
+    await expect(page).toClick('.item-name', {
+      text: 'Acevedo Buendia et al (Discharged and Retired Employees of the Office of the Comptroller)',
+    });
+    await expect(page).toClick(
+      '#app > div.content > div > div > div > aside.side-panel.metadata-sidepanel.is-active > div.sidepanel-footer > span > a'
```
RafaPolit: I thought we moved away from this type of selectors that can break easily with any sort of change to the code or even the class names?  Instead we now use selectors like "button with text `save`", or things like that?  Also, when using these type of selectors in the past we have created a lookup table and use more semantic names so that people reading the test have an idea of what it is that the test is supposed to be clicking or selecting.

e2e/suite1/connections-actions.test.ts
```diff
@@ -0,0 +1,295 @@
...
+  it('should add some people to "Firmantes"', async () => {
+    await connections.addNewConnection(2, 2, 'Máximo Cisneros Sánchez');
+    await connections.addNewConnection(2, 2, 'Margarette May Macaulay');
+  });
+
+  it('should save the relations', async () => {
+    await connections.clickSave();
+  });
+
+  it('should move the person between to countries to the correct group', async () => {
```
RafaPolit: Please fix this description.  Is it `two countries?`, is it `from countries to the correct group`? Thanks.

e2e/suite1/connections-actions.test.ts
```diff
@@ -0,0 +1,295 @@
...
+          Firmantes: ['Manuel Banchi Gundián', 'Marco Monroy Cabra'],
+          País: ['Ecuador', 'Guyana', 'Paraguay'],
+        },
+      ],
+      [
+        'Court',
+        {
+          País: ['Suriname', 'Uruguay'],
+        },
+      ],
```
RafaPolit: I really like this way of testing, it is immediately clear what the expected structure was.

RafaPolit (changes requested):

e2e/suite1/connections-actions.test.ts
```diff
@@ -0,0 +1,295 @@
+import * as connections from '../helpers/connection-actions';
```
RafaPolit: Please, rename this file to `relationships-actions.test.ts`

RafaPolit:
The flow appears to be complete and test the expected combination of actions between create, edit and delete and a good mix of all, as well as check the connections from the perspective of the connected elements.  Just a few minor language changes as well as some selectors, but nothing critical.

e2e/helpers/connection-actions.ts
LaszloKecskes: Done.

e2e/suite1/connections-actions.test.ts
LaszloKecskes: Done.

e2e/suite1/connections-actions.test.ts
LaszloKecskes: Fixed.

e2e/suite1/connections-actions.test.ts
LaszloKecskes: Credit goes to the original nightmare test, I've just mimicked it with puppeteer tools, since I've also thought it was a nice approach. :)

e2e/suite1/connections-actions.test.ts
LaszloKecskes: Done.

e2e/suite1/connections-actions.test.ts
LaszloKecskes: Simplified the selectors.

PR #5011, opened 2022-08-18
4949 entity metadata inconsistency fix

### LINKED ISSUE #4949: "Ghost" metadata creating validation error on multi-edit
**Describe the bug**
When a template has been deleted and some relationships (or inherited relationships) exists on other templates pointing to the template that was deleted, multi-editing entities that had PREVIOUSLY FILLED data will fail validation.

Single-editing them works as expected as the process is sanitizing the data, but multi-editing flow means that "invalid" metadata properties are being sent to the backend.

**To Reproduce**
Steps to reproduce the behavior (theoretical, not fully tested, may require extra steps or looking into the scenario with more detail):
1. Have templates A and B
2. Assign to template A a relationship property to template B (may require inherited relationship, not sure)
3. Create entities on template A pointing to some entities on template B
4. Delete template B (this may require to remove the properties from template A?)
5. Try to bulk-edit entities in template A

**Screenshots**
image

LaszloKecskes:
fixes #4949 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/migrations/migrations/92-remove_nonexisting_metadata/index.js
```diff
@@ -0,0 +1,66 @@
...
+      this.handleEntity(await entities.next(), propertyData);
+      // eslint-disable-next-line no-await-in-loop
+      if (this.actions.length >= this.batchSize) await this.perform(db);
+    }
+    if (this.actions.length) await this.perform(db);
+  },
+
+  async up(db) {
+    this.actions = [];
+    this.reindex = false;
```
fnocetti: I'm not sure if a reindex is not needed. I agree it won't break anything not to do it. But how long will it take for entities to get reindexed as side effect of other operations? Maybe let's ask the team. Whatever we decide I'm ok with it!

app/api/migrations/migrations/92-remove_nonexisting_metadata/index.js
```diff
@@ -0,0 +1,66 @@
...
+  reindex: false,
+
+  batchSize: 1000,
+
+  actions: [],
+
+  async getPropertyData(db) {
+    const templates = await db.collection('templates').find().toArray();
+    const indexed = Object.fromEntries(
+      templates.map(t => [t._id, t.properties ? new Set(t.properties.map(p => p.name)) : []])
```
fnocetti: I think the "else" case should be an empty set. I haven't properly got there but I saw some `.has` operations done, which array has not defined.

app/api/templates/specs/fixtures/fixtures.js
```diff
@@ -43,47 +62,62 @@ export default {
...
+          type: propertyTypes.select,
+          content: templateToBeDeleted,
+          label: 'selectToBeDeleted',
+          name: 'selectToBeDeleted',
+        },
       ],
-      commonProperties: [{ name: 'title', label: 'Title' }],
+      commonProperties: [{ name: 'title', label: 'Title', type: 'text' }],
     },
     {
```
fnocetti, line 106: Is this case even possible? How can it happen that you have a select property and it's content is a template and not a thesauri?

fnocetti (approved):
I added a couple of comments. Up to you to decide what to do.
Mostly LGTM and works fine.

app/api/migrations/migrations/92-remove_nonexisting_metadata/index.js
LaszloKecskes: I was also torn on this one, and I've decided to go with the reindex just to be safe. Let's ask the team.

app/api/migrations/migrations/92-remove_nonexisting_metadata/index.js
LaszloKecskes: You're right. Corrected. I guess this is why typescript would be useful.

app/api/templates/specs/fixtures/fixtures.js
LaszloKecskes: That is a good question. Since these were existing tests, I've just updated them assuming there were there for a reason. Let's also ask this from the team.

app/api/migrations/migrations/92-remove_nonexisting_metadata/index.js
LaszloKecskes: As discussed in the standup, a reindex is necessary, left as it is.

app/api/templates/specs/fixtures/fixtures.js
LaszloKecskes: Tests updated according to the discussion with @daneryl .

PR #5023, opened 2022-08-23
4961 - Create entity sidepanel closing button 

### LINKED ISSUE #4961: "Create Entity" sidepanel in relationships has a wrongly-implemented Close "X" button
**Describe the bug**
If you create an entity from the sidepanel in relationships assignment, the sidepanel's top X button does not close the entity creation form, but the underlying panel, which completely confuses the user.

**To Reproduce**
1. Go to create an entity in the relationships assign sidepanel
2. Try to close the sidepanel window with the top X

**Expected behavior**
The panel should close and return me to the previous "search" page, instead, it does nothing... upon clicking "Cancel" you can see that you actually closed the "underlying" sidepanel and now not even know how to bring it back.

mfacar:
fixes #4961 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


Zasa-san (commented):
The suggested solution in the issue is that the X should close the entity creation sidepanel, displaying the underlying search form below. Right now the X closes everything, including the search panel. Is the suggested solution achievable?

mfacar:
> The suggested solution in the issue is that the X should close the entity creation sidepanel, displaying the underlying search form below. Right now the X closes everything, including the search panel. Is the suggested solution achievable?

Indeed I'm using the same action as the cancel button, I think showing the search would be better but it's not possible now with a simple solution, and in case to do that we should replicate that to the cancel button.

PR #5026, opened 2022-08-23
Allowing version key on files

### LINKED ISSUE #4987: '/api/files' endpoint fails validation on files with __v property
According to the git history, we have turned off the version key feature of mongoose on the file model. This feature appends the __v property to the files. New files now created do not have this property, but there was no migration, so files already in the database since before the change still do have it. 

Now, we have set the ajv validation strict, and the `/api/files` endpoints fail on old files having this property, since it is not part of the file schema. This has not been detected, since the test have been kept up to date both in the unit tests and to nighmare e2e. Strictly speaking, the problem was the forgotten migration.

While discussing this, it was questioned whether removing the feature was a good idea in the first place. The task in this issue is:
1.  Research about this mongoose feature.
2. Decide whether we want to:
    - Keep it as it is, and write a migration to remove the __v property from the file objects.
    - Re-enable it, and make sure that this property does not get to the front end to later fail on validation through save.
    - Any better solution that comes up during the research.
3. Implement the change.

LaszloKecskes:
fixes #4987 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti:
LGTM

PR #5030, opened 2022-08-23
4921 & 4922 - Simplify date and select fields

### LINKED ISSUE #4921: Simplify date field type
Our template creator has too many field types, particularly date fields. 

Date fields could be simplified into a single field "Date/range" that includes some configuration options to:
- Enable multiple options
- Single date / range

### LINKED ISSUE #4922: Simplify select field type
In order the reduce the amount of available field types in templates we could simplify select and multi-select fields into a single field type.

The new field could be called "Thesaurus" since this field type works with thesaurus alone (relationship fields are different) and have a configuration option to indicate whether a single option or multiple options are allowed. 

Generally speaking, the behavior should that of a multi-select. TBD how the field will behave when it needs to act as a single option select (ie. selecting a new option deselects the previous one, it has some UI hint, etc) (cc @juanmnl )

Zasa-san:
fixes #4921 #4922

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


mfacar:
@Zasa-san LGTM! nothing to comment about the code, nice job!
Just a little doubt regarding the UI, in the designs the detail about other places using the same name of the property is showing a combination as type, like Select (Multiselect), I personally think it's better to keep it simple just with the final type but I'm mentioning that just in case.

Zasa-san:
> @Zasa-san LGTM! nothing to comment about the code, nice job! Just a little doubt regarding the UI, in the designs the detail about other places using the same name of the property is showing a combination as type, like Select (Multiselect), I personally think it's better to keep it simple just with the final type but I'm mentioning that just in case.

I think the way it looks right now it's sufficient as well. 
image

@juanmnl What are your thoughts?

Zasa-san:
@juanmnl @nwyu After some modifications this is the way it displays the properties now:
image

juanmnl:
@Zasa-san Looks good to me but I think you can remove the "this cannot be changed upon saving", @nwyu correct me if I'm wrong, but I think we decided at the end to not have those as we could allow a fake "downgrade" by selecting the first value. (my bad if I forgot to remove it on the designs, srry)

PR #5044, opened 2022-08-27
4872 import system ui translations from repository

### LINKED ISSUE #4872: Import System UI translations from a "predefined" remote repo
We are going to create an Uwazi-assets repo in github to contain UI translations, common Thesauri (like countries, or some HURIDCOS predefined ones), and, further down the line, common Entity CSVs for example, for extended Countries with flags, etc.

For the System Translations values, we want a small flow that will allow:
- From Translations system list select: Import UI translations (final name to be defined)
- use the GitHub API to fetch the current options available in the new Github repo (within a known path)
- select a "Download & Install" button to install it

Let's see the best path to do the actual download: ideally the client should fetch the data and send it to the backend so that we don't pollute the backend with extra options.  Alternatives: have a route in backend that downloads the data and sends it to the client and the client uses that data to create the translations (this has the advantage of not polluting the Translation routes, but it adds some backend request to the "outside world"). Finally we could expand the Translations routes to have a "from external repo" functionality.

Alternatively, we could import the assets repo into Uwazi and they could be simply loaded and used from within Uwazi.

This *could* include Backend work.

Related to #4871

### LINKED ISSUE #5013: Allow to delete "current" language (not "default")
**Describe the bug**
When you are currently on a particular language, you cannot delete that language.  It is completely obscured why the application would not allow you to delete that language: it is not a default, why does the system doesn't allow me to delete it?

**To Reproduce**
1. Add a second language
2. Set one to default, navigate the UI to the non-default language
3. Try to delete the non-default language
4. See the disabled button

**Expected behavior**
The system should allow you to delete any language except the DEFAULT.  If you are currently in the selected language, simply navigate the UI to the default language, or, maybe (?) add a modal saying "you will be directed to the X UI" Ok - Cancel

mfacar:
fixes #4872
fixes #5013

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Settings/components/LanguageList.tsx
```diff
@@ -0,0 +1,243 @@
...
+  onClick,
+  className,
+}: {
+  onClick: React.MouseEventHandler;
+  className?: string;
+}) => (
+  <button type="button" onClick={onClick} className={`btn btn-xs ${className}`}>
+    <Icon icon="sync" /> &nbsp;
+    <span>
+      <Translate>Reset default translation</Translate>
```
Zasa-san, line 57: I'm noticing that with languages that I've added before switching to this branch, the message to reset does not appear. I need to delete the language and add it again

app/react/Settings/components/LanguageList.tsx
```diff
@@ -0,0 +1,243 @@
...
+          }}
+          title={
+            <>
+              <Translate>Confirm delete </Translate>&nbsp;{deletingLanguage.label}
+            </>
+          }
+          message={
+            <>
+              <Translate>Are you sure you want to delete</Translate>&nbsp; {deletingLanguage.label}
+              <Translate> language?</Translate>
```
Zasa-san: The space before and after `language?`would be better handled with `&nbsp;`

app/react/Settings/components/LanguageList.tsx
```diff
@@ -0,0 +1,243 @@
...
+            setResettingLanguage(undefined);
+          }}
+          title={
+            <>
+              <Translate>Confirm reset translation</Translate>&nbsp;{resettingLanguage.label}
+            </>
+          }
+          message={
+            <>
+              <Translate>Are you sure you want to reset translation for</Translate>&nbsp;{' '}
```
Zasa-san: This can be removed `{' '}` and the space before `language?`should be `&nbsp;`

e2e/suite1/add-language.test.ts
```diff
@@ -24,9 +24,7 @@ describe('Add language', () => {
     await expect(page).toClick('span', { text: 'Show in cards' });
     await expect(page).toClick('span', { text: 'Save' });
     await expect(page).toClick('span', { text: 'Languages' });
-    await expect(page).toClick(
-      'ul.list-group.document-types > li:nth-child(1) > div > button.template-remove'
-    );
+    await expect(page).toClick('button', { text: 'Add language' });
```
Zasa-san, line 20: Generally this test is difficult to understand. Why does it matter if it's changing template values or not? 
If time permits and it's something useful, this test should be divided in more descriptive it's. And perhaps remove altogether the process of editing a template if it's inconsequential.

Zasa-san (commented):
Code is looking great so far!

app/react/Settings/components/LanguageList.tsx
mfacar: Nice catch, thanks, it's fixed

app/react/Settings/components/LanguageList.tsx
mfacar: absolutely

e2e/suite1/add-language.test.ts
mfacar: Just for the record, @Zasa-san and I updated the test to make it more related with languages

PR #5077, opened 2022-09-05
4804 - activity log is unstranslatable

### LINKED ISSUE #4804: Activity Log is untranslatable
**Describe the bug**
All of the contents of the Activity Log are unable to be translated, neither by the `Settings --> Translations --> User Interface` method nor the Live Translate feature. This includes the contents of the "Action" column (e.g. "Update", "Delete", "Create") and all the possible contents that can appear in the "Description" column. 

As well, the placeholder text in the search bar is not translatable either:

image 

As is the button that says "more" to display more logs in the Activity Log

**To Reproduce**
1. Go to `Settings --> Translations --> User Interface` and you'll notice that the list does not contain the strings for the contents of the Activity Log. It only contains strings for the "architecture" of the log, e.g. "Action" column header, "Description" column header, Search button, etc. 
    1.1. Even though there are entries here for "Delete", "Create" and "Update", modifying them appears to have no effect on the Action labels in the Activity Log
2. Go to `Settings --> Tools --> Activity Log` and turn on Live Translate. The same "architecture" things are translatable by Live Translate, but none of the contents of the Activity Log are. 

**Expected behavior**
If I have a non-English-language user interface configured, I would expect to be able to localize all parts of Uwazi, especially something as critical to the security and integrity of my database as the Activity Log.

mfacar:
fixes #4804

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/shared/types/activityLogEntrySchema.ts
```diff
@@ -0,0 +1,38 @@
+import { objectIdSchema } from 'shared/types/commonSchemas';
+
+const emitSchemaTypes = true;
+
+const activityLogSemanticSchema = {
```
fnocetti: This schema is not used anywhere, is it?. I understand it was created to be able to emit the types. But If the schema is not going to be used to validate user input then the types can (and imho should) be written manually.

fnocetti (approved):
Looks good to me. I left one comment but it's up to you to decide to do something about it or not.
Also, I noticed that the right side (the one in bold letters) of the description values for the activity logs is still not translatable. I particularly noted that on the logs for translation updates.
Again, just a comment. If you think this is good to go as it is, I would agree :)

app/shared/types/activityLogEntrySchema.ts
mfacar: Totally agree, I'll remove the schema

mfacar:
> 

Thanks so much for your review @fnocetti , I wrapped the description values for the activity logs to allow users to translate these values, but I didn't add any entry in UI system translations because the values are built from the body of the actions so there are lots of possible values for them.

PR #5086, opened 2022-09-07
4996 - load more instead endless scrolling in table view

### LINKED ISSUE #4996: Endless scroll in Table View is not the right approach
After several discussions, we mostly agree that Endless Scroll for the Table View was the wrong approach:
- Does not play well with "select all" features where you have little control over what is loaded
- Sharing links to a particular status could change things "here's a link, go to the previous to last item...", well, good luck!
- Detecting user nearing place where you want to load more is erratic, sometimes it loads more, sometimes you are stuck with your current results
- etc.

We are now mostly in agreement that we want a paginated view of Tables.

(Split from #3732 to work on the pagination first, and then tackle the "select all")

mfacar:
fixes #4996 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/App/scss/layout/_tableview.scss
```diff
@@ -110,7 +108,3 @@
     }
   }
 }
-
-.library-viewer.no-scroll-layout {
```
Zasa-san, line 114: I've attached a screenshot that has a visual bug that might be caused by removing this line. On mobile, it seems a side scroll bar will appear when the table has columns beyond the width of the device. (This is on Firefox)

Zasa-san: image

mfacar: Great catch, thank you @Zasa-san

PR #5087, opened 2022-09-08
float map suggestion_confidence, reindex migration

daneryl:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


database/elastic_mapping/elasticMapFactory.ts
```diff
@@ -31,10 +31,14 @@ export default {
             (property.type === 'select' || property.type === 'multiselect')
           ) {
             map.properties.suggestedMetadata.properties[property.name] = {
-              properties: propertyMappings[property.type](),
+              properties: {
+                ...propertyMappings[property.type](),
+                suggestion_confidence: {
```
konzz, line 36: Not a big problem, but if we add suggestion_confidence here, we are adding that field to every property, but we only support text, numeric, and date.
- Pros: It's prepared for every new property we support in the future.
- Cons: We are adding fields that may never be used, and I know we reach the limit in the past.

daneryl: This feature is going away, so i prefer to only add it  to the ones we support as you suggest, but this mapping is inside a condition to only be added for 'select' and 'multiselect' which is the ones that support topic classification right ?

konzz: You are right, didn't  see that is inside of a conditional for only select and multiselect

PR #5088, opened 2022-09-08
5078 endpoint available languages

### LINKED ISSUE #5078: Endpoint that returns language list with flag indicating if it has predefined system translations

daneryl:
fixes [#5078](https://github.com/huridocs/uwazi/issues/5078)

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/i18n/contentsClient.ts
+++ b/app/api/i18n/contentsClient.ts
@@ -14,6 +14,13 @@ export class GithubAuthenticationError extends Error {
   }
 }
 
+export class GithubFileNotFound extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = 'GithubFileNotFound';
+  }
+}
+
 export class ContentsClient {
   private readonly GITHUB_API_URL = 'https://api.github.com/repos/huridocs/uwazi-contents/contents';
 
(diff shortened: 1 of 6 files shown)


app/api/i18n/contentsClient.ts
```diff
@@ -25,13 +32,20 @@ export class ContentsClient {
...
       throw new GithubAuthenticationError('Github authentication failed');
     }
 
     return (await response.text()) || '';
   }
 
+  async retrieveAvailablePredefinedLanguages() {
+    const url = `${this.GITHUB_API_URL}/ui-translations/`;
+    const response = await fetch(url, { headers: this.headers() });
+    return ((await response.json()) as { name: string }[]).map(v => v.name.replace('.csv', ''));
```
mfacar, line 40: The response status should also be checked here because when the status is other than 200, it is reporting `(intermediate value).map is not a function ` instead of the real reason as error

app/api/i18n/specs/routes.spec.ts
```diff
@@ -105,9 +106,32 @@ describe('i18n translations routes', () => {
 
     describe('api/languages', () => {
       it('should return the available languages', async () => {
+        const githubResponse = [{ name: 'ar.csv' }, { name: 'es.csv' }, { name: 'fr.csv' }];
```
mfacar, line 112: It seems to be testing the same like translation.spec.ts, given this is a kind of integration test, is it needed the other one?

app/shared/languagesList.ts
```diff
@@ -88,7 +88,7 @@ const availableLanguages: LanguageSchema[] = [
     rtl: true,
     ISO639_3: 'ara',
     localized_label: 'العربية',
-    translationAvailable: true,
+    translationAvailable: false,
```
mfacar, line 91: I think it is fine if we remove the property from all the items, although it is not a big difference in size, still could be good given this file is part of the bundle, what do you think?

mfacar (approved):
LGTM!, it is great to keep the language list as a contingency 👍. I left a couple of comments about minor things.

app/shared/languagesList.ts
daneryl: We talked about it and decided to keep it to be consistent and always have the property, the mid term idea with this is that is always requested to the server, same as templates for instance instead of having it as part of the bundle.

app/api/i18n/contentsClient.ts
daneryl: Fixed

app/api/i18n/specs/routes.spec.ts
daneryl: removed the translations.spec test, both are now on the endpoint spec.

PR #5090, opened 2022-09-08
Exclude No Label in default filters

### LINKED ISSUE #5033: Exclude non related entities from the default filter
I've a property which is mutual between 4 templates,  I have defined this property as a default filter, but in the library main page this "default filter" consider the non related entities from the other templates as "no-label" entities however the property is not a part of those templates, is there a way to exclude those non related templates from the default filter ? 

image

Zasa-san:
fixes #5033 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Library/components/FiltersForm.js
```diff
@@ -62,9 +62,14 @@ export class FiltersForm extends Component {
       documentTypes.get(0) || (templates.get(0) || fromJS({})).get('_id') || 'System';
     const allFields = this.props.fields.toJS();
     const showNoValueOnFilters = allFields.size;
-    const fields = libraryHelper
+    let fields = libraryHelper
```
mfacar, line 65: Given ```fields``` could be reassigned according the size of documentTypes, parseWithAggregations doesn't need to be called all the time. Couldn't it be a ternary assignment?

app/react/Library/helpers/libraryFilters.js
```diff
@@ -87,14 +87,24 @@ export function parseWithAggregations(filters, aggregations, showNoValue = true)
...
 
+export const prepareDefaultFilters = fields =>
+  fields.map(field => {
+    if (!field.options) {
+      return field;
+    }
+    if (!field.defaultfilter) {
+      return field;
+    }
+    const filteredOptions = field.options.filter(option => option.id !== 'missing');
```
mfacar: I think this could be a single condition

mfacar (approved):
LGTM! I left a pair minor comments about code styling, nothing important so feel free to merge the PR without changes

app/react/Library/helpers/libraryFilters.js
Zasa-san: Yes, I totally forgot to refactor that!

app/react/Library/components/FiltersForm.js
Zasa-san: The reassignment doesn't change what `parseWithAggregations` does, that helper needs to run every time the component renders since it does quite a bit. `prepareDefaultFilters` just adds on top when necessary.

PR #5092, opened 2022-09-12
5039 - table view click events

### LINKED ISSUE #5039: Table View - click row, and click checkbox need to be fixed
Current row / checkbox behavior in table view is not what users expect

- Clicking on a row should show the sidepanel with the entity data, as clicking on the card view in library
- Clicking on checkbox should start a multi-selection with a single entity checked (the display should be the "stacked" view of multiselect on the sidepanel
- If there are checkboxes checked (one or more) and you click on a row, it should put the entity metadata on top of the multiselection, but keeping the checkboxes checked
- This could present a conflict if the user clicks on "edit" on a single entity in the sidepanel while having multiple checkboxes checked, this may need some thought over the process

mfacar:
fixes #5039

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Library/components/specs/TableRow.spec.tsx
```diff
@@ -0,0 +1,203 @@
...
+  describe('row format', () => {
+    it('should namespace the row with the template id', () => {
+      render();
+      const row = screen.getByRole('row');
+      expect(row.className).toContain('template-idTemplate1');
+    });
+  });
+
+  describe('columns format', () => {
+    it('should render a column with the expected columns', () => {
```
Zasa-san: I think this should be `should render a column with the expected rows`

Zasa-san (commented):
Code is looking good so far!

PR #5103, opened 2022-09-13
Upgrade to es2022

Zasa-san:


Zasa-san:
This is a simple upgrade on tsconfig that should not have an impact on how the application compiles. Whenever features of newer ECMA script versions become necessary, we will do the upgrade.

PR #5119, opened 2022-09-19
4893 - untranslatable labels in settings translations

### LINKED ISSUE #4893: Labels for things to translate in Settings --> Translations are not translatable
**Describe the bug**
In Uwazi `Settings --> Translations`, a list appears of things that a user can translate. These things are appended with a grey label that describes their nature, e.g. "entity". However, it appears that these labels are not translatable, neither by the `Settings --> Translations --> User Interface` nor by the Live Translate feature. 

- "Connection" - Doesn't exist in `Settings --> Translations --> User Interface`, not translatable via Live Translate. I would suggest changing this to "Relationship Type" because "connection" is an obsolete term that we no longer use
- "Entity" - "Entity" does exist in `Settings --> Translations --> User Interface`, but it doesn't appear to be connected to this label because changing it doesn't affect how it appears. It is not translatable via Live Translate either.
- "Thesaurus" - Doesn't exist in  `Settings --> Translations --> User Interface`, and it is not translatable via Live Translate.
- "Uwazi UI" - "Uwazi UI" does exist in `Settings --> Translations --> User Interface`, but it doesn't appear to be connected to this label because changing it doesn't affect how it appears. Neither is it translatable via Live Translate. I would suggest changing this to "System" to be more understandable to wider audiences; UI as an acronym isn't as accessible. 
- "Document" - this appears in a staging instance, but I suspect it may be now obsolete as I can't get it to appear in my own instance. 

Furthermore, it seems that the hyperlinked things that fall into the label of "Uwazi UI" don't appear translatable, neither by Live Translate nor by `Settings --> Translations --> User Interface`:
- "Filters" - There is a "Filters" term that exists in `Settings --> Translations --> User Interface`; however, it doesn't appear to be connected to this instance of the term because changing it doesn't affect how it appears here 
- "Menu" - There is a "menu" term that exists in `Settings --> Translations --> User Interface`; however, it doesn't appear to be connected to this instance of the term because changing it doesn't affect how it appears here
- "User Interface" - Doesn't exist at all in `Settings --> Translations`

**Expected behavior**
I would expect to be able to translate any area of an Uwazi instance, particularly the labels that are explicitly related to translation like these are.

**Screenshots**
List with labels: 
imagen

The hyperlinked items appended with the label "Uwazi UI":
imagen

mfacar:
fixes #4893

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


elreplicante (approved):
LGTM!

PR #5121, opened 2022-09-20
4978 - Filter untranslated terms

### LINKED ISSUE #4978: Filter untranslated entries in translations menu.
When looking at the list of terms in `Translations` it's difficult to figure out what terms I have already translated. This is especially true for `User Interface` but can also happen with long thesauri. It would be really helpful to be able to filter terms that I have already translated and only see those that might need one.

**Additional context**
An idea for the implementation is to filter keys by value. If, for a given key, there are one or more values that match the default language value, assume that it is untranslated. This way users will only see translations that are likely to be new. This should work regardless of the number of languages the user has, or the default language.

Zasa-san:
fixes #4978 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


mfacar (approved):
LGTM!

PR #5159, opened 2022-10-04
3605 text changes on tooltips in settings collection

### LINKED ISSUE #3605: text change on tooltips in Settings/Collection
Replacing text in some of the tooltips in Settings/Collection:

See design link: https://www.figma.com/file/xjsz7r65epN4B0b96zilac/0-uwazi-redesigns?node-id=2546%3A0

### LINKED ISSUE #4776: Obsolete system key translations
There are many terms in the list of translations in `Settings -> Translations -> User Interface` that are obsolete. These terms belong to strings that are no longer used in the UI. This is making the already big list more unmanageable and at times can be confusing for users.

There is a simple way of doing an initial scan for obsolete strings by using `scripts/checkTranslations.mjs`. A function similar to the one being used to detect new translatable strings can be used to this end.

The problem is that this will return many false positives for obsolete translations. An example of false positives is:
https://github.com/huridocs/uwazi/blob/0dadcceffbc1787a6bc590f59b7f1ffea5f92d53/app/react/Settings/components/AccountSettings.js#L77-L92

When the script parses the code the instantiations of this function will not be picked up as translatable strings. This means that keys on the database that are used in functions like this will be incorrectly considered obsolete, even though at runtime the function returns fully translatable text. (This also means that the script will not check if the values passed to this function are included in the translations collection)

Initially implementing the check for obsolete translations might require some manual work identifying which keys are actually obsolete. But including this scan on the `checkTranslations` script might help keep a cleaner list of translatable terms in the future.

We could also look up for a way of better scanning the code to pick up on these functions that return `<Translate />`.

mfacar:
fixes #3605
fixes #4776

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


grafitto:
LGTM!

PR #5222, opened 2022-10-17
5196 - SEO/Accessibility tweaks 1 - Language menu

### LINKED ISSUE #5196: SEO/Accessibility tweaks - 1
Navigating Uwazi with a text only browser (such as CLI links2 browser) reveals a serie of accessibility problems that also have a great impact on SEO, particularly when it comes to content discoverability.

The server side rendering of some of  our components need to be tweaked so it tosses valid, semantic html.

1. Language navigation
image

problem: The language selector is rendered as a text input that doesn't allow proper language selection.
solution: A proper list with links to the languages should be rendered instead, with spacing between the elements. Ideally we also have some title along the lines of "Select your language". It should be enclosed in a <nav> tag.

2. Sorting
image

Same problem/solution as language navigation.

Zasa-san:
fixes part 1 of #5196

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


mfacar (approved):
LGTM! Great job moving to typescript!

PR #5259, opened 2022-10-25
Relationships v2

fnocetti:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/entities/v2_support.ts
```diff
@@ -0,0 +1,124 @@
...
+  const transactionManager = new MongoTransactionManager(getClient());
+  if (!(await DefaultSettingsDataSource(transactionManager).readNewRelationshipsAllowed())) {
+    return;
+  }
+
+  const templateIdToProperties = objectIndex(
+    await templates.get({ _id: entities.map(e => e.template) }),
+    t => t._id.toString(),
+    t => t.properties.filter((prop: any) => prop.type === 'newRelationship')
+  );
```
daneryl: templateIdToProperties is not properly typed as { [k: string]: any }, is there any way to infer this type properly ? worst case we can just  type the variable inline to avoid the "any" types from this point on.

app/api/entities/v2_support.ts
```diff
@@ -0,0 +1,124 @@
...
+  const entitiesDataSource = DefaultEntitiesDataSource(transactionManager);
+
+  await Promise.all(
```
daneryl: Since the amount of entities received is unknown, this is a potential performance bottleneck, should we do it sequentially ? maybe in batches ?

app/api/entities/v2_support.ts
```diff
@@ -0,0 +1,124 @@
...
+      if (queryedEntity) {
+        relProperties.forEach((relProperty: any) => {
+          entity.metadata[relProperty.name] = queryedEntity.metadata[relProperty.name];
+        });
+      }
+    })
+  );
+  entities.forEach(entity => {
+    entity.obsoleteMetadata = [];
+  });
```
daneryl: Should we try avoiding mutating the entities passed by parameter to the function ?

app/api/entities/entities.js
```diff
@@ -427,12 +448,16 @@ export default {
   },
 
   async getWithoutDocuments(query, select, options = {}) {
-    return model.getUnrestricted(query, select, options);
+    const entities = await model.getUnrestricted(query, select, options);
+    await performNewRelationshipQueries(entities);
```
daneryl: Since this function is modifying the entities passed (mutating them or not), we think its clearer if the function returns the entities, (ideally without mutating the input).

app/api/entities/v2_support.ts
fnocetti: Makes sense. I will try to type the function properly. That should be rather quick. The possible problem I foresee is the rest of the usages already in the code. Hopefully they don't break :D

app/api/entities/v2_support.ts
fnocetti: Making it sequentially will elongate to the worst case the time needed by mongo to complete the operation. The only think that I can think that *may* benefit from serialization is the database by not having a potentially big number of active queries.  Let's discuss this one.

app/api/entities/v2_support.ts
fnocetti: I will take a look. Ideally yes. The inconvenience of immutability is that we don't have a mechanism to properly clone the entity. If we agree that modifying a composed object (think about the metadata and other nested structures) should either mutate it or deeply copy it (nothing in the middle), then we erred on the side of ease for this one :)

app/api/entities/entities.js
fnocetti: Inline with the previous comment: I think we should try to return the object only if the function is pure and doesn't mutate the parameters. Returning a the just-mutated parameter might be misleading by making you think that the function is pure. Will take a look as well.

app/api/entities/v2_support.ts
fnocetti: We'll keep this one as is for now.

app/api/entities/entities.js
fnocetti: We'll also keep this one as is for now.

app/api/entities/v2_support.ts
fnocetti: Implemented the serialized processing.

app/api/entities/entities.js
fnocetti: @daneryl actually I forgot we had decided during a meeting that a good step to fix the original concerns about this particular function was to find a better name. I have committed those changes.

daneryl: @fnocetti i think during the review we commented this but forgot to comment here :(, i think the new name is much better and makes things clearer :+1:

PR #5293, opened 2022-11-08
5129 - Fix settings navigation on mobile

### LINKED ISSUE #5129: Fix settings navigation on mobile
When clicking the items within the settings navigation on mobile, the form displays below all the items. It is misleading to a user who expects to see something after their click selection, but instead must scroll down to find the content that they are expecting to see.

This could be solved with a caret right icon. Link to the [Figma design](https://www.figma.com/file/kpGAA005yHdL6Ch1rncaB2/SaaS-Improvements?node-id=791%3A7875).

Zasa-san:
fixes #5129

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Pages/components/PagesList.js
```diff
@@ -27,7 +27,16 @@ class PagesList extends Component {
...
       <div className="panel panel-default">
-        <div className="panel-heading">{t('System', 'Pages')}</div>
+        <div className="panel-heading">
+          {' '}
+          <I18NLink to="settings/" className="only-mobile">
+            <Icon icon="arrow-left" directionAware />
+            <span className="btn-label">
+              <Translate>Back</Translate>
+            </span>
+          </I18NLink>
```
mfacar: This seems to be a good candidate to be extracted as a component because it's used in at least four other places, what do you think?

mfacar (changes requested):
LGTM, I just have one observation, please check the comment in the code. 
Everything works as expected, good job, although I'm not very convinced about the size of the label and the back button, of course, this is how it is in the design so don't mind about it 😄 .

app/react/Pages/components/PagesList.js
Zasa-san: Yes, that makes sense.

mfacar (approved):

PR #5298, opened 2022-11-10
5286 - sidepanel action buttons position

### LINKED ISSUE #5286: Sidepanel action buttons could be hiding content if translations change layout
**Describe the bug**
The sidepanel actions assume a specific amount of rows and, according to that, determine the sidepanel height, bottom margin and bottom padding to have all its contents visible.

The problem is that certain translations could change this row layout and create a higher sidepanel actions pane.  If that happens, certain scenarios could be hiding content.  One particular such case is library sidepanel upon clicking on an entity while being logged in.

If you make the size of the text of some buttons larger, it will create a taller actions container which hides the last supporting file.

Maybe, for the time being, is enough to add an extra padding to the sidepanel content... this will produce "extra white space" at the bottom of the panel on standard sized text, but it could be a quick fix.

mfacar:
fixes #5286

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


mfacar (commented):
.

PR #5337, opened 2022-11-18
Resolved #5172 via null check

### LINKED ISSUE #5172: Error when deleting an empty thesauri value
**Describe the bug**
An error is thrown when you delete the last (empty) thesauri value in settings.

**To Reproduce**
1. Go to Settings
2. Create a new thesauri
3. Add a group and values value
4. Delete the last empty value field that appears outside the group
5. See error in developer tools control

**Expected behavior**
Idealy nothing should happen. Should the delete button even the there?

**Screenshots**
image

image

 - [x] Desktop
 - [x] Mobile

**Browser**
All

enigmatic-bacon:
fixes #5172 

This is my first time doing a PR for open source and I'd love some guidance. 

- The bug has to do with the last group item possibly being a null item. I've added a null-check.
- I've run the tests in `app\react\Thesauri\components\specs` successfully (all tests pass before and after changes)
- I've verified that the bug is fixed by reproducing the steps in #5172

**Consideration:**
It's possible that the bug should be addressed higher up the stack trace. Maybe `this.groups` should never hold a null item. What do you think?

**Questions:**

- I want to create a regression test to ensure the bug doesn't come back. I feel that `app\react\Thesauri\components\specs\ThesauriFormField.spec.js` would be a good place for that. Does that sound right? Is there any documentation available to help me write my first test case?
- When I wanted to run a smaller test suite, I ran `node --max-http-header-size 20000 ./node_modules/.bin/jest app/react/Thesauri/components/specs/  --watch --verbose --runInBand"`. Does that sound right? Running just those tests took nearly 10 minutes, is that normal?
- Are there any other steps I should perform in the future before making a PR?
- Issue #5172 states we may want to entirely remove the trash icon from the right side of the last non-group item. Would that be a separate issue at this point?

Thanks!
Alex

PR checklist:
- [ x ] Update READ.me ?
- [ x ] Update API documentation ?

QA checklist:
- [ ] Smoke test the functionality described in the issue
- [x] Test for side effects
- [x] UI responsiveness
- [x] Cross browser testing
- [x] Code review


konzz:
@enigmatic-bacon Hey thanks for this, if you could rebase this branch would be great, if not I will keep it updated, and we will review the changes soon.

enigmatic-bacon:
Sure thing. Happy to as soon as I get home tonight. Could I be added as an assignee on this task as well?

enigmatic-bacon:
Would a maintainer be willing to approve running workflows? First-time contributors can't run them themselves.

PR #5357, opened 2022-11-24
Feature/relationships text references

### LINKED ISSUE #5341: Text references for relationships v2
A clear and concise description of what the problem is. Ex. I want to create entities in a batch [...]

**Have you considered an alternative?**

**Additional context**

fnocetti:
fixes #5341

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/entities/v2_support.ts
```diff
@@ -65,23 +64,16 @@ const deleteRelatedNewRelationships = async (sharedId: string) => {
...
-    const candidates = await service.getCandidateEntitiesForEntity(sharedId, language);
-
-    await entitiesDataSource.markMetadataAsChanged(candidates);
-
-    if (index && candidates.length > 0) {
-      await search.indexEntities(
-        { sharedId: { $in: candidates.map(c => c.sharedId) } },
-        '+fullText'
-      );
+    const denormalizationService = DenormalizationService(transactionManager);
```
LaszloKecskes, line 73: Let's discuss this part. What I don't like is that the default DenormalizationService receives a transactionManager (and a mongo one, even), but I see that the reason maybe that you need to run that executeOnCommitHandlers. So let's see if we can figure something out.

app/api/entities/v2_support.ts
```diff
@@ -96,7 +88,7 @@ const denormalizeAfterEntityUpdate = async ({
   const transactionManager = new MongoTransactionManager(getClient());
   if (await DefaultSettingsDataSource(transactionManager).readNewRelationshipsAllowed()) {
     const denormalizationService = DenormalizationService(transactionManager);
```
LaszloKecskes, line 90: Same here as above, just marking.

LaszloKecskes, line 90: Same as above.

app/api/files.v2/database/schemas/relationshipTypes.d.ts
```diff
@@ -0,0 +1,13 @@
+import { ObjectId } from 'mongodb';
```
LaszloKecskes: Is the filename correct? These are not relationshipTypes, but rather fileTypes?
Also, as we have discussed, can we remove the 'd' if it was created by hand? I'd vote for meaningful distinction, rather than consistency.

app/api/relationships.v2/model/Relationship.ts
```diff
@@ -1,25 +1,92 @@
...
+      throw new Error('Text references must have at least 1 selection');
+    }
+    super(entity);
+    this.file = file;
+    this.selections = selections;
+    this.text = text;
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  equals(_pointer: TextReferencePointer): boolean {
```
LaszloKecskes: Continuing the discussion: can we check how much this depends on user movement? It might be that the library that gives these selections calculates the bounding box on the selected text. If that is the case, then the same piece of text will always have the same selections, and it might be worth trying to equate the file and selections, even if we miss minute differences (like few pixels due an extra space at the end, etc.).

Issue #4963, opened 2022-08-03
Ability to use curly and straight quotes interchangeably


fnocetti:
**Is your feature request related to a problem? Please describe.**
Some syntaxes for search queries, like those for exact match and matching particular fields imply the use of straight double quotes. e.g:
`attachments.mimetype:"image/png"`.
From #4816 (and some conversations regarding that issue) it looks like for some users the default quote when using Uwazi is the curly one and it's not as straightforward to detect the mistake, or even type a straight quote instead of a curly one.

**Describe the solution you'd like**
We can consider replacing the curly quotes with straight ones on the fly, since curly ones do not seem to participate in any particular syntax.

iJustErikk:
I'll be taking this on, any problems? I also think users should be able to interchange within a single query eg. 
`attachments.mimetype:“image/png"`. This would be helpful in a partial copy-paste, when one does not get the last quote, but types a regular after.

iJustErikk:
Hi, how should I go about adding a test case for this? Should this go into the spec tests for search or E2E? Right now I am trying to create an entity with a jpg attachment (batman) and search for it. I am not sure if this is the right way to go about testing this. 

Update: I am being silly. I ended up writing a mock for this

iJustErikk:
Could someone add me to this ticket and review my PR please? I am ready to take this to the next level 🚀

fnocetti:
Hi @iJustErikk , I went through your PR and added a couple comments.
@RafaPolit I think this should include an escaping mechanism. What do you think?
Per https://lucene.apache.org/core/2_9_4/queryparsersyntax.html the Lucene escape char is `\`, we should use the same.

RafaPolit:
I have two questions:
- Are curly quotes actually reaching the ES query or are they being wrongly parsed by the body / query parser?  So, bottom line, is it arriving to the query as curly quotes or as an escaped URL code?
- I'm concerned about the smell this introduces: since we introduce a global replacement, then we need to create a escaping mechanism. This sounds like the wrong approach?

Maybe this transformation can be done only in client to prevent the API from being polluted with such problems and allow direct API calls that actually can discern between the two?

RafaPolit:
@iJustErikk Maybe hold on on the actual implementation and move forward with the e2e.  The e2e will still prove useful whatever approach we decide to implement.

LaszloKecskes:
[...]

PR #5369, opened 2022-11-30
feature: replace curly with straight quotes when building query

iJustErikk:
fixes #[4963](https://github.com/huridocs/uwazi/issues/4963). This allows curly and straight quotes to be used interchangeably (either in same query or a query can use curly quotes in lieu of straight ones).

PR checklist:
- [x] Update READ.me ?
- [x] Update API documentation ?

QA checklist:
- [x] Test for side effects
- [x] UI responsiveness
- [x] Cross browser testing
- [x] Code review


README.md
```diff
@@ -99,7 +99,7 @@ $ ./run start
 $ yarn hot
 ```
 
-This will launch a webpack server and nodemon app server for hot reloading any changes you make.
+This will launch a webpack server and nodemon app server for hot reloading any changes you make. NOTE: You may need to migrate and reindex. PITFALL: Make sure to run the blank-slate script prior to migrate and reindex.
```
fnocetti: This section's parent section, that is about how to set up the development environment already states that you need to run the blank-state script during the process. I think this new statement may be misleading since blank-state runs are only needing if you are starting a fresh set-up. In other case you may be clearing useful debugging/testing data.
Regarding the note about migrate and reindex, Uwazi will usually write a message in the console if migrate and reindex is needed. 
IMHO we should not add this notes in here but, @RafaPolit ,maybe we should start a troubleshooting section at the end of the README with this kind of hints?

app/api/search/search.js
```diff
@@ -557,13 +557,15 @@ async function searchTypeFromSearchTermValidity(searchTerm) {
 const buildQuery = async (query, language, user, resources, includeReviewAggregations) => {
   const [templates, dictionaries] = resources;
   const textFieldsToSearch = _getTextFields(query, templates);
-  const searchTextType = query.searchTerm
-    ? await searchTypeFromSearchTermValidity(query.searchTerm)
+  // GitHub: Ability to use curly and straight quotes interchangeably #4963
```
fnocetti: As part of our coding style, we don't introduce comments in the code unless absolutely necessary. This particular piece of information is provided by the git history. The commit messages usually hold this information and GitHub will add the PR number to the squashed commit during the merge. From there anyone can find the related issue on GitHub. Let's remove it :)

run
```diff
@@ -3,15 +3,15 @@
 set -eo pipefail
 
 function start {
-  docker-compose up --build -d "${@}"
+  docker compose up --build -d "${@}"
```
fnocetti, line 6: @daneryl could you take a look at this? It looks like it's fine to change cause compose as part of docker cli is the next version.

app/api/search/specs/search.spec.js
```diff
@@ -1160,4 +1160,25 @@ describe('search', () => {
...
+      const valueInObjectDeep = (obj, value) =>
+        Object.entries(obj).some(([, v]) => {
+          if (v === value) return true;
+          if (typeof v === 'object') {
+            const f = valueInObjectDeep(v, value);
+            if (f) return true;
+          }
+        });
+      userFactory.mock(undefined);
+      spyOn(elastic, 'search').and.callFake(async () => Promise.resolve(result));
```
fnocetti, line 1177: I think it will be really valuable and hopefully not that complex to actually go E2E with this instead of mocking. Let me check on Monday what would be the approach and I'll get back to you.

fnocetti (changes requested):
Hi @iJustErikk , thank you for your PR. I added a couple comments, please take a look.

app/api/search/search.js
iJustErikk: I'll remove it! Thank you

README.md
iJustErikk: I'll begin the troubleshooting section at the bottom

run
fnocetti: We agreed this is a good change. Resolving this comment.

app/api/search/specs/search.spec.js
iJustErikk: following up on this :)

README.md
iJustErikk: does the new troubleshooting section look better?

fnocetti:
We decided to follow a different approach on this issue, so we are closing this PR now. Thank you for the time you invested working on this!

PR #5386, opened 2022-12-05
Add numbers types on property configurations in IX

### LINKED ISSUE #5375: Support for numeric in IX
**Describe the bug**
Information Extraction should show numeric properties in the configuration pop up but they don't show.

popup

grafitto:
fixes #5375

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/MetadataExtraction/specs/PropertyConfigurationModal.spec.tsx
```diff
@@ -0,0 +1,98 @@
...
+ */
+import React from 'react';
+import { fireEvent, screen } from '@testing-library/react';
+import { defaultState, renderConnectedContainer } from 'app/utils/test/renderConnected';
+import {
+  PropertyConfigurationModal,
+  PropertyConfigurationModalProps,
+} from '../PropertyConfigurationModal';
+
+describe('Attachments Modal', () => {
```
Zasa-san: Attachments modal? :thinking:

Zasa-san (commented):
When accepting suggestions for numeric properties it's asking if they should be applied to all languages or not. Numeric properties are the same across languages. 

Also, If the user doesn’t apply to all languages, then numeric properties are not saving in the entity across all languages. This is a different behavior from when you edit an entity, where changing a numeric property affects all languages.
Looking at the code it seems that date properties would also have this behavior.

Is this expected? Otherwise maybe open an issue about it.

app/react/MetadataExtraction/specs/PropertyConfigurationModal.spec.tsx
```diff
@@ -0,0 +1,98 @@
...
+      expect(items).toEqual(['Date', 'Numeric', 'Text', 'Title']);
+    });
+
+    it('should call onAccept when the accept button is clicked', () => {
+      const acceptButton = screen.getByRole('button', { name: 'Save' });
+      fireEvent.click(acceptButton);
+      expect(props.onAccept).toHaveBeenCalledWith(props.currentProperties);
+    });
+
+    it('should call onAccept when the accept button is clicked', () => {
```
Zasa-san: This test should be "when the cancel button is clicked". :grimacing:

PR #5419, opened 2023-01-05
Start replicaset for development when docker compose up

### LINKED ISSUE #5396: docker compose for development mongodb starts replica set with wrong primary
Even with manually executing rs.initiate its not working.

daneryl:
fixes #5396 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,6 +25,14 @@ services:
     ports:
       - '27017:27017'
 
+  mongoreplicaset_start_script:
+    image: "mongo:4.2"
+    restart: "no"
+    container_name: mongoreplicaset_start_script
+    depends_on: 
+      - mongo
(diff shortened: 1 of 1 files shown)


LaszloKecskes (approved):
After deleting the containers and volumes, it works without extra steps.

One thing I don't like is that it leaves an extra container in the mix that executes the command every time we do compose up. The only other idea I can think of is having our own script in the package.json for yarn, which does a compose up, then tries to execute this from the host shell.
But, that is an overkill from the other direction, and people accustomed to just "docker compose up" will probably miss it. So I think it is good for now until we have a better idea.

PR #5566, opened 2023-02-23
5507 - Relationships sidepanel fixes

### LINKED ISSUE #5389: New entity side panel behaves inconsistently.
**Describe the bug**
Following https://github.com/huridocs/uwazi/issues/5363 there are inconsistencies with the behavior of the side panel when creating a new entity.

If an entity is selected, and the user is on the relationships tab on the side panel, creating a new entity renders an empty side panel. This is because the new entity side panel is still on the relationships tab.

This also happens when no entity is selected, there’s a tab for relationships. 

Likewise, selecting entities while the create entity side panel is open causes the relationships count to change, and if the selected entity has a different template than the new one an error is displayed on the template selector.

**To Reproduce**
1. Go to the library
2. Create a new entity
3. Select other entities and play around with the tabs on the side panel

**Expected behavior**
The new entity side-panel should not have tabs at all, what should happen when selecting an entity while creating a new one is unclear. 

**Screenshots**
image
image
image
image

 - Desktop
 - Mobile

**Browser**
Firerox

**Additional context**
It’s likely that editing an entity has the same problems and should be addressed. It’s also important to check what happens when multiselecting, and on the table view.

### LINKED ISSUE #5507: Viewing an entity’s relationships while switching between entities for some scenarios doesn't show relationship cards.
**Describe the bug**
View an entity’s relationships while switching between entities.  

**To Reproduce**
1. Login into an instance with at least three entities, one of them with relationships and the other two without relationships.
2. Select one entity with the RelationShip tab opened on the side panel.
3. Switch selection between the three entities, at some point the relationships are not shown.
No error appears, Uwazi merely does not show the relationships with other entities. 
Clicking the buttons Collapse all or Expand all do not allow viewing the relationships either.

**Expected behavior**
Uwazi shows the existing relationships, and the action buttons allow a user to complete necessary actions. 

**Screenshots**
Screenshot 2023-02-03 155245

Clicking View and navigating back to the page to view the relationships panel results in having the relationship count show zero. 

**Additional context**
It was reported by @serena-huridocs as part of testing for #5331

Zasa-san:
fixes #5507 #5389

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


grafitto (approved):
LGTM!

PR #5592, opened 2023-03-06
Csv import thesauri id duplication hotfix

### LINKED ISSUE #5585: Importing entity through CSV breaks thesauri.
**Describe the bug**
Importing through CSV an entity that includes a thesauri values to be created on the flight breaks the thesauri in such a way that for each group in the thesauri, it creates a value witin the group that shares label and id with the parent.

**To Reproduce**
1. Create a thesauri with a group `A` that contains at least a value `A1`.
2. Create a template with a select/multiselect property that uses the thesauri from 1. as content.
3. Import through CSV an entity for the themplate created in 2. and use `B` as value for the select property. Note the value doesn't exist in the thesauri and should be created on the flight by the import process.

**Current behavior**
The entity is created but the thesauri looks like follows:
```
- A
  - A1
  - A
- B
```

where both `A`s also share the ids.

**Expected behavior**
The entity is created and the thesauri looks like follows:
```
- A
  - A1
- B
```

**Additional context**
Note that because it created those children values, you are able to use the parent thesauri as values in the import, which will end up assigned to the freshly created child values.

LaszloKecskes:
fixes #5585

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/csv/arrangeThesauri.ts
```diff
@@ -59,23 +59,33 @@ type ThesauriValueData = {
...
   const thesauriIdToNewValues = new Map();
   const thesauriIdToNormalizedNewValues = new Map();
+  const thesauriIdToGroups = new Map();
 
   allRelatedThesauri.forEach(t => {
     const id = t._id.toString();
     const a = flatThesaurusValues(t, true);
     const thesaurusValues = a.map(v => normalizeThesaurusLabel(v.label));
+    const thesaurusGroups =
+      t.values?.filter(v => v.values)?.map(v => normalizeThesaurusLabel(v.label)) || [];
```
fnocetti: I think the second optional chaining is not needed. `filter` will at least return an empty array.

LaszloKecskes: Thanks, corrected.

fnocetti (approved):
LGTM!

Issue #5596, opened 2023-03-07
Entity creation with relationship metadata is not checking for the existence of the relationship type


fnocetti:
**Describe the bug**
When creating an entity for a template that has a relationship property, it is not checking that the relationship type associated to the property actually exists. This causes problems later in the listing and reading of entities.

**To Reproduce**
1. Have templates `A` and `B`.
2. Create a relationship type `relType1`.
3. Add to `A` a relationship property that points to template `B` using the type `relType1`.
4. Create an entity `B1` of type `B`
5. Start to create an entity of type `A` and assign `B1` in the relationship field, **but don't save it yet**.
6. In other tab, go to the settings and delete `relType1`
7. Back to the original tab, save the entity that was being created.

**Current behavior**
The entity is created but you are showed a server error. In the logs it will show errors of the following shape
```
TypeError: Cannot read properties of undefined (reading 'name')
    at getGroupData (uwazi/app/api/relationships/groupByRelationships.js:9:84)
    at uwazi/app/api/relationships/groupByRelationships.js:26:23
    at Array.forEach (<anonymous>)
    at groupRelationships (uwazi/app/api/relationships/groupByRelationships.js:25:17)
    at app/api/relationships/relationships.js:211:78
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```
Because of this, the library and the entity view won't work properly.

**Expected behavior**
You should be presented with a validation error and not being allowed to create the entity.

**Additional context**
This also happens with the public forms, but the sender is not presented with any errors.

Related to #5595

fnocetti:
We agreed that fixing #5595 would make this bug unreproducible, so the PR for that issue would fix this one as well.

daneryl:
> We agreed that fixing #5595 would make this bug unreproducible, so the PR for that issue would fix this one as well.

I do not understand how #5601 fixes this issue ? it is only validating on delete, can you explain please ? the issue describes how creation does not validate that the relation type exists. the PR only validates that you can not delete a relation type that is being used, i understand this fixes some cases, but not validation on entity creation ?

fnocetti:
This is fixed in relationships v2 and should be closed after the deprecation and migration.

PR #5601, opened 2023-03-09
Check that relationship type is not used in relationship property bef…

### LINKED ISSUE #5595: Deleting a relationship type doesn't check if type is being used in relationship property
**Describe the bug**
When a relationship type is being used in a relationship property and the user tries to delete it, the system will check for connections using the relationship type but not for uses in properties. I won't neither delete the property.
It results in an inconsistent state where relationship properties exist that doesn't have a valid relationship type.
Note that this isn't triggered if connections for the type exist because the other check will prevent it.

**To Reproduce**
1. Create a relationship type `relType1`
2. Have 2 templates: `A` and `B`
3. Create in `A` a relationship property that points to entities of template `B` and uses `relType1`
4. Delete `relType1`

**Current behavior**
The relationship type is deleted and the property persists on the template but with no, or invalid, relationship type.

**Expected behavior**
The model is left in a consistent state. Either you're not able to delete the relationship type or the property gets deleted as well.

fnocetti:
…ore deleting it

fixes #5595 #5596 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/templates/templates.ts
```diff
@@ -304,5 +304,9 @@ export default {
     return model.count({ 'properties.content': thesauriId });
   },
 
+  async countByRelationType(relationTypeId: string) {
+    return model.count({ 'properties.relationType': relationTypeId });
```
LaszloKecskes: It might be a bit much, but since we have collections with multiple templates that can have over a 100 properties, maybe adding an index for this on the model might be worth it, since it is not a big effort.

LaszloKecskes:
We have discussed about the front end facing part of this issue. This is currently throwing an unhandled backend error. Will that be enough?

LaszloKecskes (changes requested):
Code looks good, and functionally works, let's discuss the comments.

app/api/templates/templates.ts
fnocetti: 100 is not, IMO, a big number. Also, the deletes of relationship types shouldn't be frequent. I would say it doesn't justify the penalty of maintaining the index.

fnocetti:
> We have discussed about the front end facing part of this issue. This is currently throwing an unhandled backend error. Will that be enough?

Let's talk about this. Pragmatically speaking, it is preventing an undesired situation the cheapest way possible. It is more or less what we agreed.

EDIT: let's check if we can make it more meaningful without touching the frontend.

LaszloKecskes:
This was more of a double-checking question. If it was discussed, then it works.

LaszloKecskes (approved):
Questions discussed, looks good.

LaszloKecskes (approved):
Looks good.

LaszloKecskes (approved):
Everything as discussed.

PR #5628, opened 2023-03-22
[HOTFIX] Fix PDFview page canonical when page is undefined

daneryl:
PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Viewer/PDFView.js
+++ b/app/react/Viewer/PDFView.js
@@ -105,7 +105,7 @@ class PDFViewComponent extends Component {
 
     return (
       <>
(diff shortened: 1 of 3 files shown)


Zasa-san (approved):
LGTM. I can't think how to test if this would trigger notifications on the errors channel, but I don't see a problem with this change.

daneryl:
> LGTM. I can't think how to test if this would trigger notifications on the errors channel, but I don't see a problem with this change.

in this particular case there is no way we can know in advance on the channel, only bots produced the error, which only happens in production.

PR #5644, opened 2023-03-28
5594 csv import multilanguage thesauri fix

### LINKED ISSUE #5594: CSV Import is corrupting multi-language collections
NOTE: More critical part of this bug described on comment below!

**Describe the bug**
If you import the following CSV:

| Title\_\_en | Title\_\_ko | Multilingual thesauri test\_\_en | Multilingual thesauri test\_\_ko |
| --- | --- | --- | --- |
| Bina | 비나 | North Chungcheong | 충북 |
| Bena | 베나 | South Chungcheong | 충남 |
| Bana | 바나 | Seoul | 서울 |

Where `Multiligual thesauri test` is a `Select` or `Multiselect` property, the system will do some unexpected processes:
- It "accepts" and understands that `Multilingual thesauri test__en` and `Multilingual thesauri test__ko` should populate the same property, even though it's not a property that is susceptible of being translated via CSV
- It will create values for both the english and korean values inside the thesauri (half-expected, though it's weird that it even understands the `__en`  and `__ko` as the same property)
- It will create the entity in EN with the value in English, and the entity in Korean with the Korean value.  This is really, really wrong as those type of properties cannot have different values on entities of different languages

As soon as you edit the entity, it will correctly "sync" all versions of the entity to a single value.

We need to decide what to do.  We can either:
- Simply reject those properties that cannot be translated if they are namespaced
- Ignore the non-default language values for those entites
- Actually allow the creation of translated thesauri via this method

### LINKED ISSUE #5635: Possible missing value on multi_language csv import
Assume an Uwazi set up with 3 languages: English (default), Spanish, and Hungarian. A template is set up with the text property `csv_text`. Import the following with an English UI language:
```
title,csv_text__hu,csv_text__es
csv_lang_test,hello,hola
```
This import will create 3 entities, but the English one will be missing the metadata. This impacts all languages missing from the csv.

Currently, the csv import assumes the UI language as the fall-back source. If the UI language exists, missing columns will use the value of that column. For example, if the above is imported with a Spanish UI language, the English entity will be present and have the value 'hola'. However, if the UI language column is missing, an entity will have missing metadata.

The tasks for this issue:
1, Review whether the above behavior is correct. To me, it feels counter-intuitive that the result of the csv import depends on the UI language, or that we might end up with missing values.
2, If the above behavior is not correct, decide and implement a fix (validating column structure, falling back to default language or first valid value, etc.).

+1: Additionally, I did not find any mention of multi-language csv imports (ie. the '__en' mechanics) in the manual at uwazi.io. If it indeed isn't there, it should be added.

LaszloKecskes:
fixes #5594, #5635

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/csv/validateColumns.ts
```diff
@@ -0,0 +1,117 @@
...
+  propertiesByName.title = { label: 'title', name: 'title', type: 'text' };
+  const headers = await peekHeaders(await file.readStream());
+  const safeNameHeaders = headers.map(h => templateUtils.safeName(h, newNameGeneration));
+  const doesNotHaveLanguage = notTranslated(availableLanguages);
+  const [headersWithoutLanguage, headersWithLanguage] = _.partition(safeNameHeaders, h =>
+    doesNotHaveLanguage(h)
+  );
+  const languagesPerHeader: Record<string, Set<string>> = {};
+  headersWithLanguage.forEach(h => {
+    const [name, l] = h.split('__');
```
fnocetti, line 47: Is this safe to do? Could it be possible that the name generation leads to something like `word__other__en`?

LaszloKecskes: That's a good question, probably yes, it can. This is done similarly in other places in the csv import (for example, entityRow.ts in extractEntity), so I wanted to keep it consistent, but that means this one is an older problem. Would it make sense to create a separate, smaller issue focusing on this?

fnocetti: Then, I agree with you and keeping it consistent is the best thing to do now. It may make sense to create a lower priority issue, but also, we haven't received an incident in that regard. So I leave it to you :)

app/api/csv/specs/csvLoaderSelects.spec.js
```diff
@@ -86,52 +90,112 @@ describe('loader', () => {
...
-      actualEntities.reduce(
-        (result, e) => ({
-          ...result,
-          ...(e.metadata[propertyName]?.length
-            ? { [e.title]: e.metadata[propertyName].map(v => v.label).join('|') }
-            : {}),
-        }),
-        {}
-      );
+  it('should share metada values (thesauri value pointers) across languages', async () => {
```
fnocetti: There's a typo here in `metadata`

fnocetti (approved):
Proceeding to functional testing

fnocetti:
The development works as expected

app/api/csv/validateColumns.ts
LaszloKecskes: Issue at: #5656

PR #5708, opened 2023-04-13
HOTFIX Fix relationship property not detecting existing relationships

### LINKED ISSUE #5645: Edit an entity with relationship metadata causes the addition of relations in the relationship hub
- Have templates A and B.
- Create a relationship type relType1.
- Add to A a relationship property that points to template B using the type relType1.
- Create an entity B1 of type B
- Create an entity A1 of type A and assign B1 in the relationship field and save.
- Edit entity A1, do not make any changes and save.

Current behavior
A new relation between A1 and B1 is automatically added each time A1 is edited. No need to perform any change to A1. The simple fact to click on edit then save leads to the addition of one relation in the relationship hub.

Please see below a screencast illustrating the issue: 

https://user-images.githubusercontent.com/82155887/228432368-e71831c8-8d74-44c3-8bc6-2127ddd755a2.mov

fnocetti:
fixes #5645 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/migrations/migrations/128-remove-duplicated-relationships/index.js
+++ b/app/api/migrations/migrations/128-remove-duplicated-relationships/index.js
@@ -1,0 +1,114 @@
+/* eslint-disable no-await-in-loop */
+
+import { ObjectId } from 'mongodb';
+
+async function findRelationshipPropertyHub(db, sharedId, relationType) {
+  const result = await db
+    .collection('connections')
+    .aggregate([
+      {
+        $match: {
+          entity: sharedId,
+        },
+      },
+      {
+        $lookup: {
+          from: 'connections',
+          localField: 'hub',
+          foreignField: 'hub',
+          as: 'rightSide',
+        },
+      },
+      {
+        $unwind: '$rightSide',
+      },
+      {
+        $match: {
+          'rightSide.entity': { $ne: sharedId },
+        },
+      },
+      {
+        $group: {
+          _id: '$rightSide.hub',
+          templates: { $addToSet: '$rightSide.template' },
+        },
+      },
+      {
+        $match: {
+          $and: [{ 'templates.0': relationType }, { 'templates.1': { $exists: false } }],
+        },
+      },
+    ])
+    .toArray();
+
+  if (!result[0]) {
+    return null;
+  }
+
+  return result[0]._id;
+}
+
+async function deduplicateHub(db, hubForProperty) {
+  const relationshipsInHub = await db
+    .collection('connections')
+    .find({ hub: hubForProperty })
+    .toArray();
+  const toDelete = [];
+  const unique = [];
+  relationshipsInHub.forEach(relationship => {
+    if (
+      unique.find(
+        u =>
+          u.template && u.template.equals(relationship.template) && u.entity === relationship.entity
+      )
+    ) {
+      toDelete.push(relationship);
+    } else {
+      unique.push(relationship);
+    }
+  });
+  if (toDelete.length) {
+    await db
+      .collection('connections')
+      .bulkWrite(toDelete.map(d => ({ deleteOne: { filter: { _id: d._id } } })));
+  }
+}
+
+export default {
+  delta: 128,
+
+  name: 'remove-duplicated-relationships',
+
+  description: 'Removes relationships duplicated by issue #5645',
+
+  reindex: false,
+
+  async up(db) {
+    const templates = await db.collection('templates').find({}).toArray();
+
+    await templates.reduce(async (previousTemplate, template) => {
+      await previousTemplate;
+
+      const relationshipProperties = template.properties.filter(p => p.type === 'relationship');
+      if (relationshipProperties.length) {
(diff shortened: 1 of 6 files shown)


app/api/migrations/migrations/128-remove-duplicated-relationships/specs/128-remove-duplicated-relationships.spec.js
```diff
@@ -0,0 +1,54 @@
+import testingDB from 'api/utils/testing_db';
+import migration from '../index.js';
+import { fixtures } from './fixtures.js';
+
+describe('migration remove-duplicated-relationships', () => {
+  beforeEach(async () => {
+    // jest.spyOn(process.stdout, 'write').mockImplementation(() => {});
```
LaszloKecskes: You may remember, but just to be safe: don't forget to uncomment this. I often do. :)

app/api/migrations/migrations/128-remove-duplicated-relationships/specs/128-remove-duplicated-relationships.spec.js
```diff
@@ -0,0 +1,54 @@
...
+      // fixtures.connections[12],
+      // fixtures.connections[13],
+      // fixtures.connections[14],
+      fixtures.connections[15],
+      fixtures.connections[16],
+      fixtures.connections[17],
+      fixtures.connections[18],
+      fixtures.connections[19],
+      // fixtures.connections[20],
+      fixtures.connections[21],
```
LaszloKecskes: I know we strive for as few comments as possible, but I think in this case it would be worth it to leave a small comment here (or in the fixtures next to withRepetitionNoProp) explaining that hubs without relationship properties are intentionally left out. I think it's not a trivial decision, and if we will need review this any time in the future, it might be helpful.

app/api/migrations/migrations/128-remove-duplicated-relationships/index.js
```diff
@@ -0,0 +1,114 @@
...
+}
+
+export default {
+  delta: 128,
+
+  name: 'remove-duplicated-relationships',
+
+  description: 'Removes relationships duplicated by issue #5645',
+
+  reindex: false,
```
LaszloKecskes, line 84: Just to double-check: this doesn't need indexing, since we are not changing entity metadata, and we're not indexing the connections in elastic?

LaszloKecskes:
This is not related to the issue, but the failure on the api unit test might be a systemic flakiness. The failing lines are the following (taken from the CI run):
 ```
FAIL Server app/shared/specs/isSameDate.spec.ts (964 MB heap size)
  ● should only consider year, month and day

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

       5 |   const todayUtc = moment().utc();
       6 |   const todayInAnHourUtc = todayUtc.clone().add(1, 'hour');
    >  7 |   expect(isSameDate(todayUtc.unix(), todayInAnHourUtc.unix())).toBe(true);
         |                                                                ^
       8 |
       9 |   const tomorrowUtc = todayUtc.clone().add(1, 'day');
      10 |   expect(isSameDate(todayUtc.unix(), tomorrowUtc.unix())).toBe(false);

      at Object.toBe (shared/specs/isSameDate.spec.ts:7:64)
 ```
 
 If I read it correctly, that's going to fail if run after 23:00 utc. We're probably not seeing this often, because that's outside working time for all of our developers (even in South America), but if you agree, I think this would merit an issue.

LaszloKecskes (changes requested):
There are some minor comments (with probably one change only), but otherwise looks good and works. If the comments are closed, it is good to go!

app/api/migrations/migrations/128-remove-duplicated-relationships/specs/128-remove-duplicated-relationships.spec.js
fnocetti: Yes, that's a good one :D

app/api/migrations/migrations/128-remove-duplicated-relationships/specs/128-remove-duplicated-relationships.spec.js
fnocetti: Yeah, I'm in the process of adding more granularity to the tests. I think that's what will introduce this information without needing comments.

app/api/migrations/migrations/128-remove-duplicated-relationships/index.js
fnocetti: Correct

fnocetti:
> This is not related to the issue, but the failure on the api unit test might be a systemic flakiness. The failing lines are the following (taken from the CI run):
> 
> ```
> FAIL Server app/shared/specs/isSameDate.spec.ts (964 MB heap size)
>  ● should only consider year, month and day

Yes, I think I've seen that in the past. Will create an issue.

PR #5760, opened 2023-04-28
5570 quick new relationships UI crud

### LINKED ISSUE #5570: New relationship CRUD through frontend (v2)
Currently, new relationship properties are edited manually in the database, or through direct API calls to the endpoints. The goal is to provide a quick solution for interacting with new relationship objects through the front end. This is necessary for the feature to work, and it will speed up our own testing and development process as well.

LaszloKecskes:
fixes #5570 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/relationships.v2/model/Relationship.ts
```diff
@@ -71,6 +71,15 @@ class Relationship {
     this.to = to;
     this.type = type;
   }
+
+  static getSharedIds(relationships: Relationship[]): Set<string> {
```
fnocetti, line 74: I can't seem to find a use of this method. Please check.

app/api/relationships.v2/services/GetRelationshipService.ts
```diff
@@ -0,0 +1,51 @@
...
+    this.authService = authService;
+    this.entitiesDS = entitiesDS;
+  }
+
+  async getByEntity(
+    sharedId: string
+  ): Promise<{ relationships: Relationship[]; titleMap: Record<string, string> }> {
+    const relationships = await this.relationshipsDS.getByEntities([sharedId]).all();
+
+    const involvedSharedIds: Set<string> = new Set(
```
fnocetti: Maybe this is were you meant to use the static method in Relationships.ts?

fnocetti (changes requested):
Nice work!
Please, check the comments and the failing tests.
Also, the functionality seems to work as expected. I think what needs just an extra bit of work is that the list of relationships shows in a misleading way. When you create a relationships from entity A to B, if you then visit entity B, the list will shows the arrow pointing in the wrong direction, and the current entity in both ends (once as sharedId, and once as a title). I think it could be useful to just show both sides for each relationships, even if the current entity at view shows many times.

LaszloKecskes:
> 

That's a nice catch. :)

app/api/relationships.v2/model/Relationship.ts
LaszloKecskes: You've found it.

app/api/relationships.v2/services/GetRelationshipService.ts
LaszloKecskes: Precisely.

fnocetti (approved):
LGTM!

PR #5769, opened 2023-05-03
5768 - [HOTFIX] Media properties with external URLs produce CORS errors.

### LINKED ISSUE #5768: Media with a resource added from URL produces a CORS error and doesn't show the media.
**Describe the bug**
When a media property points to an external URL this is not displayed because it is producing a CORS error while fetching.

**To Reproduce**
1. Create an entity that has a media property
2. Select an external video URL, ie. from youtube.com
3. Save the entity
4. The video is not rendered.

**Expected behavior**
The video should be displayed and to be reproducible.

**Screenshots**
Screenshot 2023-05-02 215430

**Additional context**
The bug was introduced in #5582. It doesn't affect uploaded videos.

mfacar:
fixes #5768

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


mfacar:
While working on issue #5768, other problems introduced by #5604, were identified:
- Timelinks were broken when adding at creating entities.
- The media and images fields in PublicForm were behaving wrongly

Additionally, one problem of rendering was visible with that PR, the cards in TilesView had as key the index for the item instead a unique identifier, which was causing the media values were taking the values of the entities in the order of the list.

These problems are also fixed in this PR.

@grafitto , @Zasa-san , @RafaPolit

PR #5784, opened 2023-05-06
Disable fa-flip-horizontal for some icons like fa-check

EzzalddeenAli:
fixes #
add function  to disable fa-flip-horizontal for some icons like fa-check


RafaPolit:
Thank you for this contribution.  It is hard to assess the extent of this PR since every new code introduced is failing our ESLINT check rules.  Can you please ensure the PR complies with our ESLINT configuration so that we can evaluate the PR.

Thanks.

PR #5809, opened 2023-05-11
5729 - Settings on-screen translations

### LINKED ISSUE #5729: Settings on-screen translations should not appear in default language (if they match default value)
**Describe the bug**
We recently introduced a useful display of translations for the template property names and the thesauri values in parenthesis in order to signal the user the values of the translations, but also the fact that you don't translate those by changing languages.

Seeing this in the non-default language is pretty clear: Country (País).  Makes a lot of sense.

However, when watching in the default language, it actually is more confusing: Country (Country).

So, we should only show the value on two scenarios:
- When you are NOT in the default language (even if the value is the same in default language and its translation - so that we can see they are untranslated)
- When you are in default language but the translation is not the same as the default language value.  This is important to know that a particular property name or thesaurus value is actually translated differently, even in default language.

It should NOT show if you are in the default language and the translation and original value is the same.

Also, in the templates section, there is a space missing between the original name and the translation.

Zasa-san:
fixes #5729

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


mfacar (approved):
LGTM!

PR #5859, opened 2023-05-24
Translations v2

### LINKED ISSUE #3344: Refactor translations
Change our translations collection to a flat collection so we can handle better atomic translation operations.

daneryl:
fixes #3344 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/i18n.v2/model/Translation.ts
```diff
@@ -0,0 +1,29 @@
+export class Translation {
+  readonly key: string;
+
+  readonly value: string;
+
+  readonly language: string;
+
+  readonly context: {
+    type?: 'Entity' | 'Relationship Type' | 'Uwazi UI' | 'Thesaurus';
```
LaszloKecskes: Can a context not have type?

fnocetti: Also, maybe its a good idea to extract the type to it's own type definition.

daneryl: No, i do not think type should be optional, going to change that and extract the type also :+1:

app/api/i18n.v2/schemas/TranslationDBO.ts
```diff
@@ -0,0 +1,10 @@
+export interface TranslationDBO {
+  language: string; // should be an enum ?
```
LaszloKecskes: Good question. If we have a list of supported languages somewhere with their keys, probably we should make use of that, where we define languages.

app/api/i18n.v2/services/CreateTranslationsService.ts
```diff
@@ -0,0 +1,72 @@
+import { TransactionManager } from 'api/common.v2/contracts/TransactionManager';
+import { SettingsDataSource } from 'api/settings.v2/contracts/SettingsDataSource';
+import { TranslationsDataSource } from '../contracts/TranslationsDataSource';
+import { LanguageDoesNotExist, TranslationMissingLanguages } from '../errors/translationErrors';
+import { Translation } from '../model/Translation';
+
+export interface CreateTranslationsData {
+  language: string; // should be an enum ?
```
LaszloKecskes: Same as in the DBO, probably yes.

app/api/i18n.v2/database/MongoTranslationsDataSource.ts
```diff
@@ -0,0 +1,195 @@
...
+        )
+      )
+    );
+    await stream.flush();
+
+    return translations;
+  }
+
+  async deleteByContextId(contextId: string) {
+    return this.getCollection().deleteMany({ 'context.id': contextId });
```
LaszloKecskes: For this, add an index on the collection just for context.id

app/api/i18n.v2/database/MongoTranslationsDataSource.ts
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  async deleteByLanguage(language: string) {
+    return this.getCollection().deleteMany({ language });
```
LaszloKecskes: Same: an index for language.

app/api/i18n.v2/database/MongoTranslationsDataSource.ts
LaszloKecskes: Let's probably make it a compound with the order (context.id, key), since later on there are queries that use only these two.

app/api/i18n.v2/database/MongoTranslationsDataSource.ts
```diff
@@ -0,0 +1,195 @@
...
+          'context.id': contextId,
+        });
+
+        const translationsByKey = objectIndex(
+          contextTranslations,
+          t => t.key,
+          t => t
+        );
+
+        for await (const dbt of dbtrans) {
```
LaszloKecskes: With Fede we usually use 
```
while(await cursor.hasNext()) {
    const item = await cursor.next();
```
It's a lot less pretty, but avoids the `/* eslint-disable no-restricted-syntax */` problem. On the downside, it also adds a lot of no-await-in-loops. Maybe we could discuss disabling that one on the tech call, since we seem to use it a lot.

app/api/i18n.v2/migrations/index.js
```diff
@@ -0,0 +1,107 @@
+/* eslint-disable no-await-in-loop */
+const flattenTranslations = (translation, languagesByKeyContext) => {
+  if (translation.contexts?.length) {
+    return translation.contexts.reduce(
+      (flatTranslations, context) =>
+        flatTranslations.concat(
```
LaszloKecskes: I think concat copies both arrays, so inside a reduce this can be slow. Alternatives can be pushing one by one inside reduce and always pass the reference to the array, or I think this one might be solved by a douple map, then flat().

app/api/i18n.v2/database/MongoTranslationsDataSource.ts
LaszloKecskes: Scratch that, the current one covers this.

app/api/i18n.v2/services/specs/DeleteTranslationsService.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+    new MongoTranslationsDataSource(
+      getConnection(),
+      new MongoSettingsDataSource(getConnection(), new MongoTransactionManager(getClient())),
+      new MongoTransactionManager(getClient())
+    ),
+    new MongoTransactionManager(getClient())
+  );
+
+const fixtures = {
+  translations_v2: [
```
LaszloKecskes, line 20: This is a common comment for all tests: in common.v2/testing we've started with experimenting how we could use the fixture factory for v2, so if you'd like, you could add to that next time. For now, I don't think it is necessary.

app/api/i18n/specs/translations_v2_support.spec.ts
```diff
@@ -0,0 +1,793 @@
...
+            ],
+          });
+        });
+      });
+    });
+  });
+
+  describe('updateContext', () => {
+    describe('when feature flag is on', () => {
+      it('should properly change context name, key names, values for the keys changed and deleteProperties, and create new values as new translations (OMG !)', async () => {
```
LaszloKecskes: Left an extra comment at the end of the text description.

app/api/i18n/v2_support.ts
```diff
@@ -0,0 +1,217 @@
...
+  } catch (e) {
+    if (e.message !== 'ns not found') {
+      throw e;
+    }
+  }
+};
+
+const flattenTranslations = (translation: TranslationType): CreateTranslationsData[] => {
+  if (translation.contexts?.length) {
+    return translation.contexts.reduce(
```
LaszloKecskes: Same comment as in the migration: try to go without the concat in reduce.

app/api/i18n/v2_support.ts
```diff
@@ -0,0 +1,217 @@
...
+  });
+
+  return Object.values(resultMap).map(translation => {
+    // eslint-disable-next-line no-param-reassign
+    translation.contexts = Object.values(contexts[translation.locale]);
+    return translation;
+  }) as EnforcedWithId<TranslationType>[];
+};
+
+export const createTranslationsV2 = async (translation: TranslationType) => {
```
LaszloKecskes, line 86: For the next couple, where the dependency tree's are initialized: we usually write out the initialization separately, so they are easier to use. For data sources, see files named data_source_defaults, for services, see files named service_factories. Relationships.v2 is the best example, since that is a lot.

app/api/i18n/translations.ts
```diff
@@ -175,44 +186,64 @@ const update = async (translation: TranslationType | IndexedTranslations) => {
...
-        if (!context) {
-          return Promise.resolve();
-        }
-        context.values = context.values || [];
-        context.values.push({ key, value: defaultValue });
-        return this.save(translation);
-      })
-    );
-    return 'ok';
+  async save(translation: TranslationType | IndexedTranslations) {
```
LaszloKecskes, line 243: So, the idea is that we save to both until we think we are ready to drop the v1?

LaszloKecskes (changes requested):
There are a couple of local things, but overall now everything is there I think. The online migration is impressive. :)

It's a pity we can't run the entire api unite test suite with the feature flag on to see what happens.

app/api/i18n.v2/migrations/index.js
daneryl: done :+1:

app/api/i18n/v2_support.ts
daneryl: done :+1:

app/api/i18n/translations.ts
daneryl: Yes, this gives us the ability to deactivate if something goes wrong to use the old code, while we work on a fix.

daneryl:
> There are a couple of local things, but overall now everything is there I think. The online migration is impressive. :)
> 
> It's a pity we can't run the entire api unite test suite with the feature flag on to see what happens.

Thanks for the review, working on some changes, i did run it locally with the flag activated and everything was green btw, but its a good point maybe we should be able to somehow run it on ci ?

app/api/i18n.v2/services/specs/DeleteTranslationsService.spec.ts
daneryl: Im going to have to change all fixtures on the final iteration, i will add this change to that development :+1:

app/api/i18n/translations.ts
LaszloKecskes: Nice!

app/api/i18n.v2/database/MongoTranslationsDataSource.ts
daneryl: changed to a while :+1:

app/api/i18n.v2/schemas/TranslationDBO.ts
daneryl: Added as a String literal union type on the original SettingsSchema

app/api/i18n.v2/database/MongoTranslationsDataSource.ts
daneryl: done :+1:

app/api/i18n/specs/translations_v2_support.spec.ts
daneryl: removed it :)

app/api/i18n/v2_support.ts
daneryl: done :+1:  thank you, now all transactions.run are proper transactions!

app/api/i18n.v2/contracts/TranslationsDataSource.ts
```diff
@@ -0,0 +1,25 @@
...
+  deleteByContextId(contextId: string): Promise<DeleteResult>;
+  deleteByLanguage(language: string): Promise<DeleteResult>;
+  deleteKeysByContext(contextId: string, keysToDelete: string[]): Promise<DeleteResult>;
+
+  updateContextLabel(contextId: string, contextLabel: string): Promise<UpdateResult<Translation>>;
+  updateKeysByContext(contextId: string, keyChanges: { [k: string]: string }): Promise<void>;
+
+  updateValue(key: string, contextId: string, language: string, value: string): Promise<void>;
+
+  calculateUnexistentKeys(keys: string[]): Promise<string[]>;
```
LaszloKecskes: Minimal english nitpick, but according to my dictionary it is either: unexisting, nonexistent, or inexistent.

app/api/i18n.v2/database/MongoTranslationsDataSource.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+    return translations;
+  }
+
+  async upsert(translations: Translation[]): Promise<Translation[]> {
+    const items = translations.map(translation => TranslationMappers.toDBO(translation));
+    const stream = this.createBulkStream();
+
+    await Promise.all(
+      items.map(async item =>
```
LaszloKecskes: I did not catch this last time, but maybe it's not a good idea to use the stream functions parallel, since the underlying structure holding the information is just a regular array. So while it is most likely extremely rare, there might be some race conditions involved (with multiple things triggering a flush at the same time, etc.).

LaszloKecskes: ...and actually, later in the same file (updateKeysByContext) it's synchronized through a promise based reduce. I think something like that would be the good approach.

app/api/i18n.v2/migrations/index.js
```diff
@@ -0,0 +1,107 @@
+/* eslint-disable no-await-in-loop */
+const flattenTranslations = (translation, languagesByKeyContext) => {
```
LaszloKecskes, line 2: I don't remember exactly, but should we move this to the normal migrations?

LaszloKecskes (changes requested):
Aside from those two small things, looks good.

app/api/i18n.v2/migrations/index.js
daneryl: Will move it on the next iteration, for now this migration runs when activating a feature flag

app/api/i18n.v2/contracts/TranslationsDataSource.ts
daneryl: :+1: changed to calculateNonexistentKeys

app/api/i18n.v2/database/MongoTranslationsDataSource.ts
daneryl: :+1: changed to a reduce

LaszloKecskes (approved):
Looks good!

PR #5860, opened 2023-05-24
5344 async denormalizations

### LINKED ISSUE #5344: Implement asynchronous relationships v2 property denormalization
A clear and concise description of what the problem is. Ex. I want to create entities in a batch [...]

**Have you considered an alternative?**

**Additional context**

fnocetti:
fixes #5344 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/common.v2/database/MongoDataSource.ts
```diff
@@ -22,7 +22,7 @@ export abstract class MongoDataSource<CollectionSchema extends Document = any> {
     return this.transactionManager.getSession();
   }
 
-  protected createBulkStream() {
+  createBulkStream() {
```
LaszloKecskes: Why set this public?

app/shared/types/settingsSchema.ts
```diff
@@ -280,7 +280,17 @@ const settingsSchema = {
...
         },
-        newRelationships: { type: 'boolean' },
+        newRelationships: {
+          oneOf: [
+            { type: 'boolean' },
+            {
+              type: 'object',
+              additionalProperties: false,
+              required: ['updateStrategy'],
+              properties: { updateStrategy: { type: 'string' } },
```
LaszloKecskes: Can we restrict this to an enum or literal type? I know the user will be able to set anything in the db anyway, but let's have the proper options in the code at least.

app/api/queue.v2/infrastructure/specs/ApplicationRedisClient.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+      done();
+    })
+    .catch(done);
+
+  clientMock.emit('ready');
+});
+
+it('should throw an error if the creation throws', done => {
+  const promise = ApplicationRedisClient.getInstance(createClientMock);
+  promise
```
LaszloKecskes, line 41: Double check/understanding question: this is not flaky, because we would need to emit a ready from the mock to be complete?

app/api/queue.v2/infrastructure/ApplicationRedisClient.ts
```diff
@@ -0,0 +1,37 @@
...
+      clientPromise = null;
+      return true;
+    }
+
+    return false;
+  },
+};
+
+// eslint-disable-next-line camelcase
+export const TESTING_clean = () => {
```
LaszloKecskes, line 35: What is the reason for this particular naming (capitalization, underscore)?

package.json
```diff
@@ -28,7 +28,8 @@
     "run-production": "NODE_ENV=production node server.js",
     "dev-server": "./node_modules/.bin/nodemon ./server.js --ignore 'app/dist/*' --watch 'app/api' --watch 'app/shared' --ext 'ts js tsx'",
     "webpack-server": "node ./webpack/webpack.server.js",
-    "dev-worker": "node run.js ./app/worker.ts",
+    "dev-worker": "node ./scripts/run.js ../app/worker.ts",
```
LaszloKecskes, line 31: So we have not added things to the old worker, but started a new one. They will be a different process, but sharing the redis. Will this be okay from a production standpoint (i.e. extra process)? Of course, it's going to be some time until we get there, so by then this question might be irrelevant.

app/api/queue.v2/infrastructure/StringJobSerializer.ts
```diff
@@ -0,0 +1,69 @@
...
+  async deserialize(id: string, serialized: string, definitions: Record<string, Definition<any>>) {
+    const data = JSON.parse(serialized);
+
+    const definition = definitions[data.name];
+
+    if (!definition) {
+      throw new Error(`Unregistered job ${data.name}`);
+    }
+
+    const propertyDescriptors = Object.keys(data.data).reduce(
```
LaszloKecskes: I know it's an annoying pet peeve of mine, but please no reduce for building objects and arrays unless it is absolutely necessary. Functional tools like Object.entries, map, Object.fromEach. Or if we want to use reduce, at least work on a reference rather than passing entire objects every iteration.

app/api/queue.v2/infrastructure/StringJobSerializer.ts
```diff
@@ -0,0 +1,69 @@
...
+          value: data.data[property],
+        },
+      }),
+      {}
+    );
+
+    const builtDependencies = await definition.dependenciesFactory?.(data.namespace);
+
+    const dependenciesDescriptors = builtDependencies
+      ? Object.keys(builtDependencies).reduce(
```
LaszloKecskes: Same here.

app/api/queue.v2/application/JobSerializer.ts
```diff
@@ -0,0 +1,17 @@
+import { Job } from '../contracts/Job';
+
+interface Constructor<T extends Job> {
```
LaszloKecskes: This is repeated in at least one other place (Queue.ts). I think it would make sense to put it into the job contracts, since it's the type for that constructor.

app/api/queue.v2/application/Queue.ts
```diff
@@ -0,0 +1,110 @@
+import { Job } from '../contracts/Job';
+import { JobsDispatcher } from '../contracts/JobsDispatcher';
+import { QueueAdapter } from '../contracts/QueueAdapter';
+import { JobSerializer } from './JobSerializer';
+
+interface JobConstructor<T extends Job> {
+  new (...args: any[]): T;
+}
+
+interface Definition<T extends Job> {
```
LaszloKecskes: This is repeated in JobSerializer. I think it should be imported from there.

app/api/relationships.v2/infrastructure/registerUpdateRelationshipPropertiesJob.ts
```diff
@@ -0,0 +1,37 @@
...
+import { UpdateRelationshipPropertiesJob } from '../services/propertyUpdateStrategies/UpdateRelationshipPropertiesJob';
+
+export function registerUpdateRelationshipPropertiesJob(queue: Queue) {
+  queue.register(
+    UpdateRelationshipPropertiesJob,
+    async namespace =>
+      new Promise((resolve, reject) => {
+        tenants
+          .run(async () => {
+            const transactionManager = DefaultTransactionManager();
```
LaszloKecskes, line 15: can we wrap this part into a default service factory, like the other services? This is the first service for entities, that's why it did not have the file yet (see relationships.v2 - service_factories.ts).

app/api/queue.v2/application/Queue.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  async peek() {
+    await this.ensureQueue();
+    const message = await this.adapter.receiveMessageAsync({ qname: this.queueName });
+    if ('id' in message) {
+      return this.deserializeJob(message.id, message.message);
+    }
+
+    return null;
```
LaszloKecskes, line 75: Are we sure about the null return here? If the message does not have that property, it means other messages are also in the queue which theoretically should not happen.

app/api/queue.v2/contracts/QueueAdapter.ts
```diff
@@ -0,0 +1,12 @@
+export interface QueueMessage {
+  id: string;
+  message: string;
+}
+
+export interface QueueAdapter {
```
LaszloKecskes, line 6: Does this mostly come from RedisSMQ, since if I see the Queue initialization correctly, it immediately satisfies this without any extra steps? Then, if that is the case, when we later break of from Redis, will we still keep to this?

app/api/queue.v2/application/Queue.ts
```diff
@@ -0,0 +1,110 @@
...
+  async progress(job: Job) {
+    await this.ensureQueue();
+    await this.adapter.changeMessageVisibilityAsync({
+      qname: this.queueName,
+      id: job.id!,
+      vt: job.lockWindow ?? this.options.lockWindow,
+    });
+  }
+
+  register<T extends Job>(
```
LaszloKecskes, line 92: This is a bigger question, but i'll put it here. I'll confess and say that I don't quite see the need for this registration/definition mechanism. Could not the dependency factories and all the other baggage be simply part of the implemented job class as statics?
Then, if we want still want something like that, maybe to ensure that each queue instance can serve specific jobs, then we can still have a registration, but that would be considerably simpler, just the class, no external definitions and such.

app/api/queue.v2/infrastructure/StringJobSerializer.ts
```diff
@@ -0,0 +1,69 @@
+import { Definition, JobSerializer } from '../application/JobSerializer';
+import { Job } from '../contracts/Job';
+
+const DEFAULT_DESCRIPTOR = {
```
LaszloKecskes, line 4: What's this coming from?

app/api/queue.v2/application/Queue.ts
LaszloKecskes: Looking at other parts: so the id is autogenerated by the queue?

app/api/queue.v2/infrastructure/MemoryQueueAdapter.ts
```diff
@@ -0,0 +1,66 @@
+/* eslint-disable max-classes-per-file */
```
LaszloKecskes, line 1: If I see correctly, this is only used in testing now. Do we see this ever used for any practical purpose? If not, I think we should indicate it with a comment, or move this to some test related folder. If yes, we should make sure that all the finds and filters have the relevant supporting data structures over the messages that allow for less-than linear time operations (e.g. some hash for the id find, some ordered searchable tree for the times), and the additions should make sure to update them.

app/api/settings.v2/database/MongoSettingsDataSource.ts
```diff
@@ -35,4 +35,13 @@ export class MongoSettingsDataSource
     const settings = await this.readSettings();
     return !!settings?.features?.newRelationships;
   }
+
+  async getFeatureConfiguration(featureKey: string) {
```
LaszloKecskes: If we don't use it anywhere else, maybe we could write this specifically for this one feature, but in return type it properly instead of a string-any record.

app/api/entities.v2/model/Entity.ts
```diff
@@ -1,4 +1,14 @@
-type MetadataValue = any;
+type BaseResultValue = {
+  value: string;
+  label: string;
+};
+
+type InheritedResultValue = BaseResultValue & {
+  inheritedValue: MetadataValue[];
```
LaszloKecskes: isn't this just BaseResultValue[]? We don't have full recursion, right?

app/api/relationships.v2/services/propertyUpdateStrategies/UpdateRelationshipPropertiesJob.ts
```diff
@@ -0,0 +1,31 @@
...
+import { EntityRelationshipsUpdateService } from 'api/entities.v2/services/EntityRelationshipsUpdateService';
+import { Job } from 'api/queue.v2/contracts/Job';
+
+interface IndexEntitiesCallback {
+  (sharedId: string): Promise<void>;
+}
+export class UpdateRelationshipPropertiesJob extends Job {
+  private entityId: string;
+
+  updater?: EntityRelationshipsUpdateService;
```
LaszloKecskes, line 11: Things are optional, but then we expect them to be there with '!'. I assume these are supplied by the environment somehow through that registration/definition process. If we keep that as it is, even then I would not just bypass typescript using the exclamation mark, but rather implement proper checks that throw errors when some of this is still undefined when trying to use. I think we did some things that needed initialization after the constructor and before use, and that's how we did it.

app/api/relationships.v2/services/DenormalizationService.ts
```diff
@@ -23,20 +24,24 @@ export class DenormalizationService {
...
+  private updateStrategy: RelationshipPropertyUpdateStrategy;
+
   constructor(
     relationshipsDS: RelationshipsDataSource,
     entitiesDS: EntitiesDataSource,
     templatesDS: TemplatesDataSource,
     settingsDS: SettingsDataSource,
     transactionManager: TransactionManager,
-    indexEntitiesCallback: IndexEntitiesCallback
+    indexEntitiesCallback: IndexEntitiesCallback,
```
LaszloKecskes, line 35: Why are we using a callback parameter (here and multiple places) for indexing?

LaszloKecskes:
Plus, I think the code climate warning about the import order should be fixed, that's easy enough.

LaszloKecskes (changes requested):
Comments, and some further discussions in calls.

app/api/common.v2/database/MongoDataSource.ts
fnocetti: Good catch. It shouldn't. I must have made it public to do some quick testing and forgot. BTW, the getCollection is also public, but no from this PR. I checked and there's no need for that. Also shouldn't. I'm adjusting both.

app/shared/types/settingsSchema.ts
fnocetti: Sounds good

app/api/queue.v2/infrastructure/specs/ApplicationRedisClient.spec.ts
fnocetti: Even if we emitted the ready event, this is not flaky because we are not it's not a rogue promise, it's properly then-ed and catch-ed.
But yes, in any case, that would continue until the event is emitted. That means that if the code is broken, it may be scenarios in which we see it failing by timeout.

app/api/queue.v2/infrastructure/ApplicationRedisClient.ts
fnocetti: I expect that method not to be used anywhere else that is not the testing for this file. So I chose this because it is obvious to the eye, but also will give you trouble with eslint when you try to use it :) Of course is arbitrary, and we can decide for something else.
To give context, the problem is that function should't be needed, but because there's no injection in the part of the application that is going to use that code, this is a global singleton, and global singletons are difficult to test.

package.json
fnocetti: This command line utility is not for production but for development. From that perspective, the difference is, because we didn't add anything to the current workers, you need to separately run the worker in a different terminal.
From a production point of view this is not only ok but ideal. The philosophy behind the workers is that you are able to reduce the load of the user-facing processes, plus the ability to elastically scale up according to general or eventual workloads. Having the workers run in the most granular way posible allows the infra team to individually scale up based on queues. Said other way, queues and queue workers are a tool, together with stateless api-only server processes, to scale up vertically.
This was validated with the infrastructure team, btw.

app/api/queue.v2/infrastructure/StringJobSerializer.ts
fnocetti: I'll take the disclaimer :P I'll use a forEach

app/api/queue.v2/application/JobSerializer.ts
fnocetti: I've discussed in the past that not all interfaces (and sometimes code) need to be deduplicated. But this is a pretty utility-like type. I'll try to do something about it.

app/api/queue.v2/application/Queue.ts
fnocetti: I'll take a look

app/api/relationships.v2/infrastructure/registerUpdateRelationshipPropertiesJob.ts
fnocetti: Good catch

app/api/queue.v2/application/Queue.ts
fnocetti: The id is autogenerated by the adapter, yes, it's part of it's specification.
The `in` here is working as a typeguard. The adapter will return `null` if there's no message to peek.

app/api/queue.v2/contracts/QueueAdapter.ts
fnocetti: Yes, we will. The semantics defined by the API of RedisSMQ are pretty good, and also adjusted to the kind of semantics that we need as an application.
When we move away from Redis, we will want to implement an adapter that has the same semantics. The names of the methods are not pretty, also the parameter passing strategy is far from our conventions, but just coping its names prevented the need to write an adapter for the RedisSMQ provider. But, to recap, leave names alone, the specification is what we need anyway. Having an interface in the middle IS what is going to allow us to switch implementations in the future.
If the naming of the methods, that they are a bit not "us", looks problematic I can anyway adapt them and write an adapter for RedisSMQ. What do you think?

app/api/queue.v2/application/Queue.ts
fnocetti: This is indeed a bigger question that merits a bigger answer, but I'll still leave something here for the record :)
First things first, having the dependency factories in the class, in any form, is a violation of the dependency rule and a violation to the dependency injection principle. The Jobs are part of the application layer, they implement application logic, and depend on abstractions: for example, on a `RelationshipsDataSource`, not a `MongoRelationshipsDataSource`. Deciding what implementation to use is not part of its responsibilities, it happens in the outermost layer. This is not new, that is why we have factories. They are the composition root of the dependency injection pattern. This implementation is not different to that. The only difference is that it uses a different strategy for the injection: instead of constructor injection, it uses attribute injection, because we use the constructor for a different purpose in the jobs.
Why is the registration mechanism necessary? Is not to ensure that a certain queue only processes a certain type of job. That is not needed, and should be handled by not pushing jobs to the wrong queue. The job is instantiated by a uwazi server instance and put in a queue. You can't simply put a class instance into Redis, so we serialize it. When we deserialize it and try to reconstitute the instance (to have the data and the behavior), how would the worker (running on a separate process) know which class to instantiate? I don't mean the name, the name is actually serialized, but how to find the class code. That's what the registration is for: if you are looking for this job, this is the class and this is how to build it's dependencies.
Maybe you already noticed, but going back to "the worker runs on a separate process", that's why we need to tell it how to build dependencies through the registration: they are only going to be built when the job is going to be processed.

app/api/queue.v2/infrastructure/StringJobSerializer.ts
fnocetti: That's arbitrary. The `Object.create` method uses property descriptors. Unless specified, it will make, between other things, the attributes not enumerable, which implies they are not shown in the console logs. So I just made them enumerable.
I'm anyway considering moving away from this. Options are: 
- Serialize the whole property descriptor. Which would allow to properly replicate them.
- Do not serialize the whole class but only a `data` property.
I'm leaning towards the later because this deserialization does not do deep deserialization. Meaning it won't handle nested class instances, and will just return a plain object. Following the later approach we can say that is a restriction that the `data` needs to be a POJO, otherwise the deserialization might not result as expected.

app/api/queue.v2/infrastructure/MemoryQueueAdapter.ts
fnocetti: I guess you meant "production", because it in fact has very practical purposes :P It is used for tests and as a specification for the semantics of the interface it implements. I didn't put it into the tests because I was afraid that it was modified out of the semantics to fulfill some testing need in the future. But I guess all of that can be mitigated with a comment.

app/api/settings.v2/database/MongoSettingsDataSource.ts
fnocetti: Let's do that.

app/api/entities.v2/model/Entity.ts
fnocetti: Very, very good catch :D

app/api/relationships.v2/services/propertyUpdateStrategies/UpdateRelationshipPropertiesJob.ts
fnocetti: I will surely implement checks, but they will only add a prettier message. This would also throw an exception in runtime :)

app/api/relationships.v2/services/DenormalizationService.ts
fnocetti: We decided to use a callback many months ago because we decided not to implement an indexation service yet. For the mongo side we can just inject an scoped DB connection, but for the indexation you need to inject a context-ed search instance. Also we don't want to import v1 code from v2 code whenever possible. IndexEntitiesCallback is just a contract, and we write one-liners implementing it in the functionality in the composition roots (factories)

app/api/queue.v2/application/JobSerializer.ts
fnocetti: Ok, update: I actually agree with you, this type should live within the JobSerializer module. As usual, good catch!

app/api/queue.v2/contracts/QueueAdapter.ts
LaszloKecskes: I'm not one to hung up on names that much, and I think these are fine. I was more concerned about the implied functionality, but if we've thought about it and we think it's good, that's also nice. If I would go into nitpicking territory, the only thing I don't like is the naming of the 'vt' parameter, which is the traditional undergrad math student type unreadable shorthand naming (which might be fine in locally defined anonymous functions or lambdas, but not in interfaces like this). :)
But there's of course no need to do anything about that for now.

app/api/queue.v2/application/Queue.ts
LaszloKecskes: Let's have a call about this. ...when we can, depending the other things that came up.

app/api/queue.v2/infrastructure/MemoryQueueAdapter.ts
LaszloKecskes: I'm fine with a comment.

app/api/relationships.v2/services/propertyUpdateStrategies/UpdateRelationshipPropertiesJob.ts
LaszloKecskes: I agree. I was more trying to be consistent, rather than fault the usage, out of context it's fine. If we used things (optional-then-exclamation mark) like this habitually, I would not have commented.
And actually, if we think that these kind of errors do not need prettier messages, we can start using this pattern here and leave it as is. I'll leave the decision to you., I'm fine either way.

app/api/relationships.v2/services/DenormalizationService.ts
LaszloKecskes: Ah, yes, I forgot about the context.

app/api/relationships.v2/services/propertyUpdateStrategies/UpdateRelationshipPropertiesJob.ts
fnocetti: I think maybe the only difference is that this is not user facing, so I was biased towards thinking that not having a better error was ok. But yeah, let's implement proper checks

LaszloKecskes (approved):
Everything as discussed.

PR #5862, opened 2023-05-25
5831 integrate relationship v2 migrations

### LINKED ISSUE #5831: Integrate relationship v2 migration into Uwazi
The proof-of-concept migrations to relationship v2 are currently ad-hoc scripts in the scripts folder. The feature is at a point, where it wold be beneficial for us to be able to perform the migration from Uwazi proper, and see the results in the UI. This would make iterating over the underlying methodology easier.

The task is:
- clean up the migration code, arrange the scripts into the api folders
- create a migration endpoint
- create a quick, just-for-testing UI for the migration

Note: this iteration will be untested, tests should be written after the methodology is final!

LaszloKecskes:
fixes #5831 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/relationships.v2/services/MigrationService.ts
```diff
@@ -0,0 +1,201 @@
...
+import { V1Connection, V1ConnectionDisplayed } from '../model/V1Connection';
+import { EntityPointer, Relationship } from '../model/Relationship';
+
+const HUB_BATCH_SIZE = 1000;
+
+class RelationshipMatcher {
+  readonly fieldLibrary: Record<string, Record<string, Set<string>>>;
+
+  constructor(v1RelationshipFields: V1RelationshipProperty[]) {
+    const groupbedByTemplate = objectIndexToArrays(
```
fnocetti: There's a typo in `groupbedByTemplate` :)

app/api/relationships.v2/contracts/HubDataSource.ts
```diff
@@ -0,0 +1,16 @@
...
+type HubType = {
+  _id: ObjectId;
+};
+
+export interface HubDataSource {
+  exists(): Promise<boolean>;
+  create(): Promise<void>;
+  drop(): Promise<void>;
+  insertIds(ids: string[]): Promise<void>;
+  all(): MongoResultSet<HubType, string>;
```
fnocetti: This needs to return an application layer ResultSet, MongoResultSet is leaking implementation.

app/api/common.v2/database/data_source_defaults.ts
```diff
@@ -1,5 +1,6 @@
+import { getClient } from 'api/common.v2/database/getConnectionForCurrentTenant';
```
fnocetti: Why was this change needed?

app/api/relationships.v2/contracts/V1ConnectionsDataSource.ts
```diff
@@ -0,0 +1,25 @@
+import { MongoResultSet } from 'api/common.v2/database/MongoResultSet';
```
fnocetti: Importing database code from the application layer is a violation to the dependency rule. Contracts need to use ResultSet

app/api/relationships.v2/database/MongoHubsDataSource.ts
```diff
@@ -0,0 +1,84 @@
...
+class TemporaryDataSourceError extends Error {}
+
+export class MongoHubsDataSource extends MongoDataSource<HubType> implements HubDataSource {
+  protected collectionName = 'temporary_hubs';
+
+  protected created = false;
+
+  protected dropped = false;
+
+  shouldBeCreated() {
```
fnocetti: I think all of these `should*` methods should be private

app/api/relationships.v2/database/MongoHubsDataSource.ts
```diff
@@ -0,0 +1,84 @@
+/* eslint-disable max-classes-per-file */
+import { Db } from 'mongodb';
+
+import { MongoIdHandler } from 'api/common.v2/database/MongoIdGenerator';
+import { MongoResultSet } from 'api/common.v2/database/MongoResultSet';
+import { MongoDataSource } from '../../common.v2/database/MongoDataSource';
+import { HubDataSource, HubType } from '../contracts/HubDataSource';
+
+const collectionExists = async (db: Db, name: string) => {
```
fnocetti: Why not make of this another private method?

app/api/relationships.v2/database/MongoHubsDataSource.ts
```diff
@@ -0,0 +1,84 @@
...
+      throw new TemporaryDataSourceError(`Collection ${this.collectionName} was already dropped`);
+    }
+  }
+
+  shouldBeReady() {
+    this.shouldBeCreated();
+    this.shouldNotBeDropped();
+  }
+
+  async create(): Promise<void> {
```
fnocetti, line 33: I don't get the idea behind the `shouldBeCreated` and `shouldNotBeDropped` logic. Why are those checks important? Wouldn't a "drop and re-create" flow be enough?

app/api/relationships.v2/database/MongoV1ConnectionsDataSource.ts
```diff
@@ -0,0 +1,83 @@
...
+    dbo.templateName
+  );
+
+export class MongoV1ConnectionsDataSource
+  extends MongoDataSource<V1ConnectionDBO>
+  implements V1ConnectionsDataSource
+{//eslint-disable-line
+  protected collectionName = 'connections';
+
+  allCursor(): MongoResultSet<V1ConnectionDBO, V1Connection> {
```
fnocetti: We don't usually sufix this method names with cursor. `all` should be enough.

app/api/relationships.v2/database/MongoV1ConnectionsDataSource.ts
```diff
@@ -0,0 +1,83 @@
...
+    const cursor = this.getCollection().find({}, { session: this.getSession() });
+    return new MongoResultSet<V1ConnectionDBO, V1Connection>(cursor, mapConnections);
+  }
+
+  getConnectedToHubs(
+    _hubIds: string[]
```
fnocetti: Prefixing with _ is usually a convention for unused parameters. If we can find a better name I would use the _ in the variable instead.

app/api/relationships.v2/database/MongoV1ConnectionsDataSource.ts
```diff
@@ -0,0 +1,83 @@
...
+        },
+      },
+      {
+        $set: {
+          pickedEntity: { $arrayElemAt: ['$entityInfo', 0] },
+          pickedRelType: { $arrayElemAt: ['$relTypeInfo', 0] },
+        },
+      },
+      {
+        $set: {
```
fnocetti, line 67: This `$set` + `$unset` combination can be performed with a single `$project`

app/api/common.v2/database/MongoResultSet.ts
```diff
@@ -36,6 +36,18 @@ export class MongoResultSet<T, U = T> implements ResultSet<U> {
     return null;
   }
 
+  async nextBatch(size: number): Promise<U[]> {
```
fnocetti, line 39: There's no added benefit from batching here because it is still fetching one by one. The only difference is in that this is blocking the process until it fetched `size` elements, which is not necessary good, but it's not saving any network traffic. I would just use the normal `all` and iterate over the results using the `forEach` method on the result set.

app/api/relationships.v2/services/MigrationService.ts
```diff
@@ -0,0 +1,201 @@
...
+  private async transformHubs(
+    matcher: RelationshipMatcher,
+    transform: boolean = false,
+    write: boolean = false
+  ) {
+    const hubCursor = this.hubsDS.all();
+    let total = 0;
+    let used = 0;
+    // eslint-disable-next-line no-await-in-loop
+    while (await hubCursor.hasNext()) {
```
fnocetti: I would have just used `all` + `forEach`. Check up the comment on `MongoResultSet`

app/api/relationships.v2/services/MigrationService.ts
```diff
@@ -0,0 +1,201 @@
...
+      // eslint-disable-next-line no-await-in-loop
+      const connections = await this.v1ConnectionsDS.getConnectedToHubs(hubIdBatch).all();
+      const connectionsGrouped = objectIndexToArrays(
+        connections,
+        connection => connection.hub,
+        connection => connection
+      );
+      const connectionGroups = Array.from(Object.values(connectionsGrouped));
+      const transformed: Relationship[] = [];
+      for (let i = 0; i < connectionGroups.length; i += 1) {
```
fnocetti, line 146: I personally find `forEach`s much more readable than plain `for`s

app/api/relationships.v2/services/MigrationService.ts
```diff
@@ -0,0 +1,201 @@
...
+        const {
+          total: groupTotal,
+          used: groupUsed,
+          transformed: groupTransformed,
+          // eslint-disable-next-line no-await-in-loop
+        } = await this.transformHub(connectionGroups[i], matcher, transform);
+        total += groupTotal;
```
fnocetti, line 152: Adding to previous comment, I think this is an almost perfect use case for a `reduce`

app/api/relationships.v2/services/MigrationService.ts
```diff
@@ -0,0 +1,201 @@
...
+  static transform(first: V1Connection, second: V1Connection, newId: string) {
+    const sourcePointer = new EntityPointer(first.entity);
+    const targetPointer = new EntityPointer(second.entity);
+    const relationshipType = second.template;
+    return new Relationship(newId, sourcePointer, targetPointer, relationshipType);
+  }
+}
+
+export class MigrationService {
+  private transactionManager: MongoTransactionManager;
```
fnocetti: I think this is not used

app/api/templates.v2/database/TemplateMappers.ts
```diff
@@ -12,17 +13,29 @@ type PropertyDBO = TemplateDBO['properties'][number];
...
+        templateId,
+        property.denormalizedProperty
+      );
+    case propertyTypes.relationship:
+      if (!property.relationType) throw new Error('Relation type is required');
+      return new V1RelationshipProperty(
+        propertyId,
+        property.name,
+        property.label,
+        property.content || V1RelationshipProperty.ALL_MARKER,
```
fnocetti: Maybe making it optional would have done the trick. The problem with this kind of markers is that, since it is defined as `' '`, typescript will infer the type as `' ' | string` which it decide its just `string` so it's a weaker typing compared to making it optional.

fnocetti (changes requested):
Hey, I added some comments. Please take a look. 
Other than that, I noticed there are no tests at all. Are we good with that?

LaszloKecskes:
> 

Yes. We are going to test this, when the methodology is final.

app/api/relationships.v2/services/MigrationService.ts
LaszloKecskes: ...I mean it almost makes sense. :)

app/api/relationships.v2/contracts/HubDataSource.ts
LaszloKecskes: To correct this, I needed to add lower level cursor declarations to the ResultSet (like hasNext and next), since I'm processing batches. I think that's okay, those are fairly common cursor behaviors, but take a look. If not, the alternative is that we can implement a batched forEach next to the current ordinary one.

LaszloKecskes: Scratch the above, I've went ahead and implemented the batchedForEach, since it sounded like a nice idea regardless of context.

app/api/common.v2/database/data_source_defaults.ts
LaszloKecskes: It wasn't, reverting it.

app/api/relationships.v2/contracts/V1ConnectionsDataSource.ts
LaszloKecskes: Seems to be a common thing I'm missing.

app/api/relationships.v2/database/MongoHubsDataSource.ts
LaszloKecskes: Done.

app/api/relationships.v2/database/MongoHubsDataSource.ts
LaszloKecskes: Or rather, just add the implementation to the exists method, there's no need for a general function.

The reason (or rather, history) it's here that I first experimented in creating a generic temporary datasource with autogenerated collection names and such, but I could not make it work without leaking mongo types into the services so I've scratched it. It was close, so if we need it, we can probably make it work (if anything, as an abstract class for sure). But we don't need it.

app/api/relationships.v2/database/MongoHubsDataSource.ts
LaszloKecskes: Maybe for the creation-drop, yes. But I'd like to have these checks for at least the insert functionality, and not have that auto create things and for that they are I think useful.

app/api/relationships.v2/database/MongoV1ConnectionsDataSource.ts
LaszloKecskes: Done.

app/api/relationships.v2/database/MongoV1ConnectionsDataSource.ts
LaszloKecskes: I think we do this all the time in uwazi. Regardless, specifically for this case I've just removed the variable, since we're not reusing it, we can do the map in the db call.

app/api/relationships.v2/database/MongoV1ConnectionsDataSource.ts
LaszloKecskes: I've tried, it gave me:
"MongoServerError: Bad projection specification, cannot exclude fields other than '_id' in an inclusion projection:"

Maybe I've tried the wrong way, if you'd like you can give it a try in the branch.

app/api/common.v2/database/MongoResultSet.ts
LaszloKecskes: The benefit comes not necessarily from the read side, but that one is able to do the processing more efficiently on the batch. For example, where this is used (in the migration service), callback is requesting additional data and performs write operations as well, so we save traffic there, since we're not doing that one-by-one for each hub.
Of course, at that point from the read perspective the batch is that some syntactic sugar for waiting and putting things in an array, but I think that's still worth it to make things a little bit easier.

app/api/relationships.v2/services/MigrationService.ts
LaszloKecskes: This has been changed to a batched for each, and I'm still standing by the batching. :)

app/api/relationships.v2/services/MigrationService.ts
LaszloKecskes: For this and the previous comment, this is definitely a question of preference. I could see the case for a forEach, but we can't do that here due to the async operation. Here, my preference is the opposite, I prefer loops for syncing promises (and in general, I don't prefer using reduce for anything else than the intended purpose in data processing, that is, providing aggregations).
We've talked about this with Joan on a couple of occasions (mostly in migration and csv import context), and we've agreed that in the end both are legible and correct, so we will allow both ways, depending on programmer preference.

app/api/relationships.v2/services/MigrationService.ts
LaszloKecskes: Removing it. I wanted to use it, but I'm concerned about the limits for transactions, and these operations can be quite big, and easily reversible anyway.

app/api/templates.v2/database/TemplateMappers.ts
LaszloKecskes: We should have a dedicated discussion about this, since this has been come up, and now I'm starting to feel that we should not have undefined mean a thing aside from it's original meaning in javascript.

For this specific one, I did it, since it matches how we store it in the database. So no marker for now.

app/api/relationships.v2/database/MongoV1ConnectionsDataSource.ts
fnocetti: Yeah, the way to do it is not to provide the include and excludes but only specify what you want yo see in the result (everything but what you want to exclude + the ones you put in the $set, that are calculated). Anyway, it's just a micro optimization

app/api/templates.v2/database/TemplateMappers.ts
fnocetti: It's indeed a good conversation to have :)

app/api/common.v2/database/MongoResultSet.ts
```diff
@@ -49,6 +61,13 @@ export class MongoResultSet<T, U = T> implements ResultSet<U> {
     }
   }
 
+  async batchedForEach(batchSize: number, callback: (items: U[]) => Promise<void> | void) {
```
fnocetti: I would suggest `forEachBatch` so the suggestions on the editor are more natural.

fnocetti (approved):
Other than the fussy comment about naming, this looks good!

PR #5895, opened 2023-06-05
5842 thesaurus import duplication error message

### LINKED ISSUE #5842: Better error message for thesaurus csv import for duplicated values
The thesaurus csv import does not allow for duplicated values. Currently, in case a file contains duplicated values, the error shows as a basic 'validation error' on the ui. The detailed ajv error only says that there was an error about duplicated values, but does not provide any clue about which ones exactly.

The task is:
1, Improve the error message to contain the duplicated values.
2, Make sure that the UI gets the error in a format so it can display the message.

LaszloKecskes:
fixes #5842

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


mfacar:
LGTM! Just a comment to define the next steps. At the moment the message in the front end seems to be redundant. 
duplicated1

Should we create a new issue or add that as part of this PR?

LaszloKecskes:
> 
That's strange, I don't see that.
image

LaszloKecskes:
Follow up: as per discussion on a call with @mfacar, we will remove the details from the ajv message in the backend, and let the front end build it itself.

PR #5951, opened 2023-06-17
Scope collections to transaction in MongoDataSource

fnocetti:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


daneryl:
LGTM

app/api/common.v2/database/MongoDataSource.ts
```diff
@@ -1,28 +1,132 @@
...
+    readPreference: null,
+    bsonOptions: null,
+    writeConcern: null,
+    hint: null,
+  } as const;
+
+  private appendSessionToOptions(args: any[], method: keyof MethodsOptionsArgPosition) {
+    const optionsPosition = MongoDataSource.scopedMethods[method]! - 1;
+    const minArgumentsLength = optionsPosition;
+    const missingArguments = minArgumentsLength - args.length;
```
LaszloKecskes: Absolute nitpick, and maybe it's because morning still, but can we rename missingArguments to missingArgumentsLength? :)

app/api/common.v2/database/MongoDataSource.ts
```diff
@@ -1,28 +1,132 @@
...
+
+    return paddedArgs;
+  }
+
+  private scopeCollectionToSession(collection: Collection<CollectionSchema>) {
+    if (this.collectionProxy) {
+      return this.collectionProxy;
+    }
+
+    const self = this;
```
LaszloKecskes, line 99: Just a question: this is an alternative solution to binding the 'this' to the 'get' inside the proxy?

app/api/common.v2/database/MongoDataSource.ts
```diff
@@ -1,28 +1,132 @@
...
+
+    paddedArgs[optionsPosition] = {
+      ...paddedArgs[optionsPosition],
+      session: this.transactionManager.getSession(),
+    };
+
+    return paddedArgs;
+  }
+
+  private scopeCollectionToSession(collection: Collection<CollectionSchema>) {
```
LaszloKecskes, line 94: Implementation detail question.

Since the reflection and other meta programming option in javascript are lacking, we've ended up needing to supply explicit data to the function that creates the proxy (i.e. the list of scoped methods). Whenever I have to do that, I always feel that i'm not doing real reflection anymore, since the solution is not completely based on normal function information, so I start to look for a "normal", non-reflective option.

Specifically, in this case: have you considered doing a wrapper around the collection object and explicitly pass down args to the collection as needed? The way I see it, the trade-off is that the proxy method is more elegant and way less typing, but the wrapper solution gets rid off meta-programming altogether,  thus it will be more readable in the future.

Also, to make sure: this is just a question for discussion, I'm not saying to change this here. :)

app/api/common.v2/database/MongoDataSource.ts
```diff
@@ -1,28 +1,132 @@
...
+    stats: 1,
+    findOneAndDelete: 2,
+    findOneAndReplace: 3,
+    findOneAndUpdate: 3,
+    aggregate: 2,
+    watch: 2,
+    initializeUnorderedBulkOp: 1,
+    initializeOrderedBulkOp: 1,
+    count: 2,
+    dbName: null,
```
LaszloKecskes, line 66: I understand that this null is a way to say that we're going to call the function unchanged. Why not just leave those out of the list?

app/api/common.v2/database/MongoDataSource.ts
```diff
@@ -1,28 +1,132 @@
...
   private transactionManager: MongoTransactionManager;
 
+  private collectionProxy?: Collection<CollectionSchema>;
+
   constructor(db: Db, transactionManager: MongoTransactionManager) {
     this.db = db;
     this.transactionManager = transactionManager;
   }
 
+  static scopedMethods: MethodsOptionsArgPosition = {
```
LaszloKecskes, line 31: Here we basically encode information about the mongo db functions. It might be too much, but I'm not sure I'm comfortable leaving this untested. If the subject was our own code as well, sure, but not if it is about external things.
I understand that this is potentially a lot of work, so I'm willing to stick my neck out and do it. Maybe it can be done with a clever it.each. :)

LaszloKecskes (changes requested):
The only comment with potential changes is the one about testing a scopedMethods, the rest is just for discussions.

app/api/common.v2/database/MongoDataSource.ts
fnocetti: Surely can :)

app/api/common.v2/database/MongoDataSource.ts
fnocetti: Yes... aliasing `this` is a bit more performant than binding, but that's not the reason whatsoever. Within the proxy, `this` is always set to the receiver (the object in which the property being proxied was called), so I didn't want to change it since I don't know what effects it could have. Probably none, but erred on the side of safety :)

app/api/common.v2/database/MongoDataSource.ts
fnocetti: I agree, and that was my first approach. The thing is that it's not easy to proxy the whole thing and respect types. Also, it's not only the methods, you need to proxy every member in the class with getters and setters.
That's why another approach I tried was to just have a reduced proxy, namely `ScopedCollection` that only proxies what's in use, and developers can add proxied methods on demand. But I discarded that for some reason. I can give it another try anyway.

app/api/common.v2/database/MongoDataSource.ts
fnocetti: Cause I couldn't figure out the typings. It doesn't seem to be a way to remove attributes from an interface based on the type of its value.
Typing them undefined instead of null won't make them optional, you'll still have to include them with undefined as value.
Typing them as never, will fail in all cases, cause you need to include them since they're not optional, but as you include it there's no value of type never, so it will also fail.
The apparent solution is to make them never and optional, but there's no way to only make some properties optional in a mapped type. So... :)

app/api/common.v2/database/MongoDataSource.ts
fnocetti: It's in fact a lot of work. I started it, and then dropped it because I wasn't sure it was valuable use of time.
The working assumption here is that the interface in mongo client is uniform, so I tested one or two methods, and should be good enough. But I agree it may seem a bit of a weak test strategy.

app/api/common.v2/database/MongoDataSource.ts
LaszloKecskes: No, "this" being taken is a good reason, I would've done the same.

app/api/common.v2/database/MongoDataSource.ts
LaszloKecskes: As I said, there's no need to change this, since it is already done. It is an internal thing we will most likely almost never change (sans bugs). I was just curious about the thought process.

app/api/common.v2/database/MongoDataSource.ts
LaszloKecskes: My worry was that if that Mongo changes, our tests should catch that. However, we are testing the data sources anyway, and if the parameter placement changes, through those we will notice the error, so yes, maybe it is not needed to explicitly test here.

LaszloKecskes (approved):
Looks good, thank you for the work on the tests!

PR #5963, opened 2023-06-21
Use error state in suggestions when segmentation is missing

### LINKED ISSUE #5962: IX: When more than 50 segmentations failed the metadata extractor stops providing suggestions
When more than 50 segmentations failed the metadata extractor stops providing suggestions

gabriel-piles:
fixes #5962 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


txau:
@gabriel-piles I just did a quick check in the code. I have one concern regarding using the suggestion state "error" assuming this error is always going to be a segmentation error. So just for clarification:
1. When the model finishes training, all suggestions will be marked as "Obsolete", including those that previously had an error.
2. Then the files that have a segmentation error will be marked as "error".
3. When processing the suggestions, it can happen that some of them give an error, ie. something happened in the service and they could not be processed.

I also have a doubt regarding what can happen if data with segmentation errors is sent for training.

I wonder if we should have an extra state specific for "segmentation error".

gabriel-piles:
@txau I also thought about adding an extra state regarding the lacking of segmentation. It is fine to me, I can do it, but a frontend developer should review the changes.

Regarding the concern with the training, the logic is a little bit different than the suggestions and all the training files are sent at the same time and it is filtered out the files that had no segmentation.

PR #5988, opened 2023-06-28
5928 v2 relationship migration text references

### LINKED ISSUE #5928: Handle text references in the v2 relationship migration
Currently the v2 relationship migration does not migrate text reference data, but the relationships themselves are already able to handle it. The task is to migrate text references as well.

Note: the early migration implementations are untested. Tests are going to be written, when the methodology is final.

LaszloKecskes:
fixes #5928 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/authorization.v2/database/schemas/permissionSchemas.ts
```diff
@@ -2,12 +2,22 @@ import { objectIdSchema } from 'api/common.v2/database/schemas/commonSchemas';
 
 const emitSchemaTypes = true;
 
+const legacyObjectIdSchema = {
```
fnocetti, line 5: We're not longer using schemas but this seems not to have been removed. Do you think it makes sense to remove within this PR? Boy scout rule :)

app/api/common.v2/contracts/ResultSet.ts
```diff
@@ -1,9 +1,14 @@
+type loopCallbackReturn = Promise<void | boolean> | void | boolean;
```
fnocetti: Type names should be capitalized. Also, I would use a more descriptive name. Something like BreakLoopSignal

app/api/common.v2/model/Tenant.ts
```diff
@@ -0,0 +1,15 @@
+class Tenant {
```
fnocetti: We need to discuss this. Tenants are not an application concept. Uwazi as an application doesn't know about tenants, everything "related" to tenants is hidden from the application in the form of injected dependencies that are already instantiated for the tenant. They're only relevant to the multitenancy strategy which is only important to the process serving Uwazi, which makes it an infrastructure concern.

app/api/common.v2/database/MongoResultSet.ts
```diff
@@ -54,52 +54,54 @@ export class MongoResultSet<T, U = T> implements ResultSet<U> {
     return mapped;
   }
 
-  async forEach(callback: (item: U) => Promise<void> | void) {
-    while (await this.hasNext()) {
+  async forEach(callback: (item: U) => loopCallbackReturn) {
+    let progress = true;
```
fnocetti: I think we need a more descriptive name for this. Progress is usually related to how advanced te process is. I would suggest something like `shouldContinue`

app/api/common.v2/contracts/ResultSet.ts
fnocetti: Also, you can just export type here.

app/api/relationships.v2/services/MigrationService.ts
```diff
@@ -65,29 +104,55 @@ export class MigrationService {
...
     this.v1ConnectionsDS = v1ConnectionsDS;
     this.templatesDS = templatesDS;
     this.relationshipsDS = relationshipsDS;
+    this.tenant = tenant;
   }
 
-  private async readV1RelationshipFields() {
+  private logNoRepair(first: V1Connection, second: V1Connection): void {
+    logLine(
+      `V2 Relationship Migration Error (tenant:${this.tenant.name})----------------------------------`
```
fnocetti: Instead of putting the tenant name here, this service should be injected with a tenant specific logger.

fnocetti (changes requested):
Left some comments. Nice job on the MongoResultSet!

app/api/common.v2/contracts/ResultSet.ts
LaszloKecskes: Done.

app/api/common.v2/database/MongoResultSet.ts
LaszloKecskes: Done.

app/api/common.v2/contracts/ResultSet.ts
LaszloKecskes: Did we not have an eslint rule for exporting things at the end?

Done.

app/api/authorization.v2/database/schemas/permissionSchemas.ts
LaszloKecskes: No, since I found a couple of others as well (unrelated to this development). Wrote an issue for it (#5991), and assigned to myself to be done right after this. It should not take long.

app/api/common.v2/model/Tenant.ts
LaszloKecskes: Will remove this in favor of the logger.

app/api/relationships.v2/services/MigrationService.ts
LaszloKecskes: As above; will do so.

app/api/log.v2/contracts/LoggerInterfaces.ts
```diff
@@ -0,0 +1,23 @@
+import { Tenant } from 'api/tenants/tenantContext';
+
+export interface LogLevelInterface {
```
fnocetti: `LogLevelInterface` and `LogEntryInterface` are internal to the implementation. I think they should be within the `infrastructure/Logger.ts` file. Nevertheless, it looks like an interface for those is not needed since, again, they are implementation details and it doesn't look like there's going to be a need for second implementations.

app/api/log.v2/contracts/LoggerInterfaces.ts
```diff
@@ -0,0 +1,23 @@
...
+export interface LogEntryInterface {
+  message: string;
+  timestamp: number;
+  level: LogLevelInterface;
+  tenant: Tenant;
+  toString(): string;
+  toJSONString(): string;
+}
+
+export interface LoggerInterface {
```
fnocetti: Please, let's kindly remove the interface postfix, it's redundant and we don't usually signal that in the class/interface names.

app/api/log.v2/infrastructure/Logger.ts
```diff
@@ -0,0 +1,103 @@
...
+  severity: number;
+
+  constructor(name: string, severity: number) {
+    this.name = name;
+    this.severity = severity;
+  }
+}
+
+const LogLevels: Record<string, LogLevel> = {
+  DEBUG: new LogLevel('DEBUG', 10),
```
fnocetti: If there's going to always be a single application-wise instance of each log level I think it makes more sense to just have plain objects, in which case an interface makes sense for typing purposes.

fnocetti (changes requested):
Mostly looks good to me! Only a few comments on the logger side :)

app/api/log.v2/contracts/LoggerInterfaces.ts
LaszloKecskes: Agree with the internal implementation, removed.

app/api/log.v2/contracts/LoggerInterfaces.ts
LaszloKecskes: Done. I've kept one place: since we now call everything Logger, one needs to disambiguate in the implementation, so:
`import { Logger as LoggerInterface } from '../contracts/Logger';`

app/api/log.v2/infrastructure/Logger.ts
LaszloKecskes: Locally, I think they are fairly equivalent (up to preference :)), since I think they are equally readable, and for this amount of objects the class overhead is negligible. The reason I would still keep it is that I expect that we will have some other functionality for LogLevels, where it will make sense to have a class for encapsulation. For example, one natural evolution of this is to use this frequently, and have the loglevel as a configuration when starting up the Uwazi server, so a comparison would then be necessary.

If you think we should wait to do a class until we need it, that's fine and I'll change it, but I think it's okay to "expect the future" this much, especially since it is no extra effort.

fnocetti: Sounds good!

fnocetti (approved):
LGTM!

PR #5994, opened 2023-06-30
5830 - IX new UI

### LINKED ISSUE #5830: IX Updated UI
```[tasklist]
### Tasks for PDF sidepanel
- [x] Date fields
- [x] Numeric fields
- [x] Clear selection button, Click to fill buttons, functionality to remove/create/update selections
- [x] Request entities on route loader, check that only needed fields are requested @grafitto
- [x] Redesign the button to accept a single suggestions
- [x] Solve visual problems on the sidepanel pdf rendering
- [x] When the PDF opens the view should scroll to the selection, if any exist.
- [x] Implement form for sidepanel to update files and selections. Disable save button if no changes and during save. Validate entity field.
```

### LINKED ISSUE #5954: You can click more than once the button to add new extractor
When adding a new extractor for IX, after clicking the button "Add", the dialog remains open while waiting for the extractor to be created. Users may click several times on "Add" while they wait as no proper feedback is provided.

Zasa-san:
related to #5830
fixes #5954 #6021

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


konzz:
There is a bug that happens whenever you open the create extractor modal and the first thing you do is press the "From all templates" button and then select a property, the state goes into a weird scenario. Not sure if this was introduced in this PR or already existed, but would be a good idea to disable that button until you select a property.

PR #6032, opened 2023-07-12
5991 remove obsolete schema from v2

### LINKED ISSUE #5991: Remove obsolete Schemas in v2 code
We have decided to change to JSON Type Definitions (JTD) on validators from the schemas, therefore there is no need to keep to the schema->generated type workflow anymore in v2. However, a few schema definitions still remain. The task is, for each remaining schema definition in v2 code:
- double check that a schema is not used anywhere, if it is, change the usage either to types or JTD, depending on the use case
- double check the generated types, copy and finalize them into manual type definitions, update usages
- remove old schemas and generated types

Note: this is for v2 only, v1 remains the same.

LaszloKecskes:
fixes #5991 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti (approved):
LGTM

PR #6037, opened 2023-07-13
Relationships v2 migration user choice

### LINKED ISSUE #6002: Allow user choice in relationships v2 migration
Currently, the relationships v2 migration chooses what to migrate based on the relationship properties defined in the collection. After reviewing real data, it has been determined that while this is a good baseline, it is necessary to allow users to add to and ignore things from this set.

LaszloKecskes:
fixes #6002 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/relationships.v2/database/MongoMigrationHubRecordDataSource.ts
```diff
@@ -0,0 +1,55 @@
...
+{
+  protected collectionName = 'migrationHubRecords';
+
+  async getPage(page: number, pageSize: number): Promise<MigrationHubRecord[]> {
+    const cursor = this.getCollection().find();
+    const resultset = new MongoResultSet<MigrationHubRecordDBO, MigrationHubRecord>(
+      cursor,
+      mapRecordToApp
+    );
+    return resultset.page(page, pageSize);
```
fnocetti: The page method in the ResultSet is there so we don't have to implement getPage in the data sources. The service gets a ResultSet and it should be up to it to consume it as a page.

app/api/relationships.v2/database/MongoRelationshipMigrationFieldsDataSource.ts
```diff
@@ -0,0 +1,63 @@
+import { MongoDataSource } from 'api/common.v2/database/MongoDataSource';
+import { MongoResultSet } from 'api/common.v2/database/MongoResultSet';
+import { RelationshipMigrationFieldsDataSource } from '../contracts/RelationshipMigrationFieldsDataSource';
+import {
+  RelationShipMigrationFieldUniqueId,
```
fnocetti: I think there's a type with the capital S in Relation*S*hip.

app/api/relationships.v2/database/MongoRelationshipMigrationFieldsDataSource.ts
```diff
@@ -0,0 +1,63 @@
...
+import { mapFieldIdToDBO, mapFieldToApp, mapFieldToDBO } from './RelationshipMigrationFieldMappers';
+import { RelationshipMigrationFieldDBO } from './schemas/relationshipMigrationFieldTypes';
+
+class MongoRelationshipMigrationFieldsDataSource
+  extends MongoDataSource<RelationshipMigrationFieldDBO>
+  implements RelationshipMigrationFieldsDataSource
+{
+  protected collectionName = 'relationshipMigrationFields';
+
+  async get(fieldId: RelationShipMigrationFieldUniqueId): Promise<RelationshipMigrationField> {
```
fnocetti: I think we use to call this methods getById, don't we?

app/api/relationships.v2/database/MongoRelationshipMigrationFieldsDataSource.ts
```diff
@@ -0,0 +1,63 @@
...
+    const dbo = await this.getCollection().findOne({ ...mapFieldIdToDBO(fieldId) });
+    if (!dbo) {
+      throw new Error(`Field with id ${fieldId} not found`);
```
fnocetti: I may be missing something, but by convention we return null/undefined and perform a check in the consumer. I don't think looking for a non-existing record is a data access error.

app/api/relationships.v2/routes/routes.ts
```diff
@@ -50,44 +65,127 @@ export default (app: Application) => {
...
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    async (req: Request, res: Response<GetRelationshipMigrationFieldsResponse>) => {
+      const service = GetRelationshipMigrationFieldsService();
+      const fields = await service.getAllCombinedWithInferred();
+      res.json(fields);
+    }
+  );
+
+  app.post(
+    '/api/v2/relationshipMigrationFields/create',
```
fnocetti: Why not use `POST relationshipMigrationFields` and `PUT relationshipMigrationFields` instead of `/create` and `/update`?

app/api/relationships.v2/services/CreateRelationshipMigrationFieldService.ts
```diff
@@ -0,0 +1,39 @@
...
+    targetTemplate?: string,
+    ignored: boolean = false
+  ) {
+    const field = new RelationshipMigrationField(
+      new RelationShipMigrationFieldUniqueId(sourceTemplate, relationType, targetTemplate),
+      ignored
+    );
+    const saved = await this.transactionManager.run(async () => {
+      await this.fieldDS.create(field);
+      return this.fieldDS.get(field.id);
```
fnocetti: I think we usually return the newly created object from the `create` method. While it is a probably imperceptible optimization, it also doesn't require another query since you already have the object in memory, and Mongo will provide you with the created id, in case you needed it.
In this particular case, I don't even think you need to get. If the save was successful, then the object you have in memory is the same thing the get will return, right?

app/api/relationships.v2/services/DeleteRelationshipMigrationFieldService.ts
```diff
@@ -0,0 +1,34 @@
...
+    this.fieldDS = fieldDS;
+  }
+
+  async delete(
+    sourceTemplate: string,
+    relationType: string,
+    targetTemplate?: string
+  ): Promise<void> {
+    await this.transactionManager.run(async () => {
+      const fieldId = new RelationShipMigrationFieldUniqueId(
```
fnocetti: Can we move this outside of the transaction? So if in the future there are some checks to fail, it will do so before even starting the transaction.

app/api/relationships.v2/services/GetRelationshipMigrationFieldService.ts
```diff
@@ -0,0 +1,66 @@
...
+    templatesDS: TemplatesDataSource
+  ) {
+    this.transactionManager = transactionManager;
+    this.fieldDS = fieldDS;
+    this.templatesDS = templatesDS;
+  }
+
+  async getAll(): Promise<RelationshipMigrationField[]> {
+    let fields: RelationshipMigrationField[] = [];
+    await this.transactionManager.run(async () => {
```
fnocetti: I don't think this needs a transaction

app/shared/data_utils/specs/stringUtils.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    {
+      list1: ['a', 'a', 'a'],
+      list2: ['a', 'a', 'b'],
+    },
+    {
+      list1: ['a', 'a', 'a'],
+      list2: ['a', 'b', 'a'],
+    },
+  ])(
+    'should compare lists based on alphabetical order as localeCompare does',
```
fnocetti, line 36: Isn't it called lexicographical order? :)

app/api/common.v2/contracts/Sink.ts
```diff
@@ -0,0 +1,6 @@
+interface Sink<T> {
```
fnocetti: I like the concept, I'm not sure about the name. I would attach to SaveStream, the same way you named the implementation, factory method and variables.

fnocetti (changes requested):
Overall looks good, please take a look at the comments.

app/api/relationships.v2/database/MongoMigrationHubRecordDataSource.ts
LaszloKecskes: Done.

app/api/relationships.v2/database/MongoRelationshipMigrationFieldsDataSource.ts
LaszloKecskes: Indeed.

app/api/relationships.v2/database/MongoRelationshipMigrationFieldsDataSource.ts
LaszloKecskes: Sure.

app/api/relationships.v2/services/GetRelationshipMigrationFieldService.ts
LaszloKecskes: Nice catch.

app/api/relationships.v2/services/CreateRelationshipMigrationFieldService.ts
LaszloKecskes: Done, same for upsert.

app/api/relationships.v2/database/MongoRelationshipMigrationFieldsDataSource.ts
LaszloKecskes: Due to changes in the create and upsert services, this is not used anymore, so I ended up removing entirely.

app/api/relationships.v2/routes/routes.ts
LaszloKecskes: Done.

app/api/relationships.v2/services/DeleteRelationshipMigrationFieldService.ts
LaszloKecskes: Of course.

app/api/common.v2/contracts/Sink.ts
LaszloKecskes: That was the original idea, then I've noticed that this is a more general interface for any kind of consumer. Also, I'm hesitant to use the word stream unless we're really starting to move towards proper stream processing. Nevertheless, I'll rename it for now, then we can revisit it later when we're using it for anything else.

fnocetti (approved):
LGTM!

app/api/common.v2/contracts/Sink.ts
fnocetti: Agree with you

PR #6052, opened 2023-07-19
Mongodatasource sync support + translations v2 sync support

daneryl:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/common.v2/database/MongoDataSource.ts
```diff
@@ -120,6 +189,44 @@ export abstract class MongoDataSource<CollectionSchema extends Document = any> {
...
+    if (conditions.length === 0) {
+      return;
+    }
+
+    const modifiedDocuments = this.getCollection().find(
+      { $or: conditions },
+      { projection: { _id: 1 } }
+    );
+
+    await new MongoResultSet(modifiedDocuments, d => d).forEach(async ({ _id }) => {
```
fnocetti: Very clever :D

app/api/i18n.v2/database/MongoTranslationsSyncDataSource.ts
```diff
@@ -0,0 +1,44 @@
...
+      await previous;
+      await stream.updateOne({ _id: translation._id }, { $set: translation }, true);
+    }, Promise.resolve());
+
+    await stream.flush();
+    return translations;
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  async get(_query: any, _select?: any, _options?: any): Promise<TranslationDBO[]> {
```
fnocetti, line 33: Is this unused?

app/api/i18n.v2/database/MongoTranslationsSyncDataSource.ts
```diff
@@ -0,0 +1,44 @@
...
+    return Promise.resolve([]);
+  }
+
+  async getById(translationId: string): Promise<TranslationDBO | null> {
+    return this.getCollection().findOne({ _id: new ObjectId(translationId) });
+  }
+
+  async delete(query: { _id: string }): Promise<DeleteResult> {
```
fnocetti, line 41: I think I would make this a `deleteById` method and just take the _id as a parameter. We have being trying to make the data access methods as semantic and single purposed as possible.

app/api/common.v2/database/MongoDataSource.ts
```diff
@@ -120,6 +189,44 @@ export abstract class MongoDataSource<CollectionSchema extends Document = any> {
...
+      await this.db.collection('updatelogs').updateMany(
```
LaszloKecskes: Maybe use a bulk write stream, to avoid doing it one by one.

app/api/common.v2/database/MongoDataSource.ts
```diff
@@ -109,7 +120,65 @@ export abstract class MongoDataSource<CollectionSchema extends Document = any> {
           const original = Reflect.get(target, property, receiver);
 
           return function proxiedFunction(...args: any[]) {
-            return original.apply(receiver, self.appendSessionToOptions(args, propertyName));
+            const originalMethod = () =>
+              original.apply(receiver, self.appendSessionToOptions(args, propertyName));
+
+            const condition = args[0];
+            if (property === 'insertMany') {
```
LaszloKecskes: Instead of if's, maybe we could consider have an object/dict/map with the extra steps/return values, like the scopedMethods for the session position (or make value of the scoped methods object into a `{ sessionPosition: number, extraSteps: <function> }`). 
I think that could serve as a good middle-ground between the theoretical collection wrapper we have talked about, and the current proxy-based implementation, even for the long term.

LaszloKecskes (changes requested):
Aside from a few discussion points, it looks good.

app/api/common.v2/database/MongoDataSource.ts
fnocetti: Let's discuss this tomorrow. Mongo changed it's API and the types we have built for the scopedMethods won't work after the bump, so time has come we need to migrate to the explicit wrapper. I wouldn't invest any effort in changing this now :p

app/api/i18n.v2/migrations/index.js
```diff
@@ -56,10 +66,13 @@ export default {
   },
 
   async up(db) {
+    timeoffset = 1;
     // process.stdout.write(`${this.name}...\r\n`);
```
fnocetti: Forgotten comment here :p

app/api/i18n/v2_support.ts
```diff
@@ -11,14 +13,17 @@ import { DeleteTranslationsService } from 'api/i18n.v2/services/DeleteTranslatio
...
 import { DefaultSettingsDataSource } from 'api/settings.v2/database/data_source_defaults';
 import { tenants } from 'api/tenants';
 import { Db } from 'mongodb';
 import { TranslationContext, TranslationType, TranslationValue } from 'shared/translationType';
 import { LanguageISO6391 } from 'shared/types/commonTypes';
-import { IndexedContextValues } from './translations';
 import migration from '../i18n.v2/migrations';
+import { IndexedContextValues } from './translations';
+
+models.translationsV2 = () =>
```
fnocetti, line 25: I leave this up to you, but this means that `translationsV2` being set is a side-effect of importing this file, which is probably a not-so-good practice. It can be more difficult to trace, and I don't thing import order is guaranteed. I would wrap this with a method you can explicitly execute when you need the thing set.

app/api/odm/model.ts
```diff
@@ -22,7 +22,15 @@ export type UwaziFilterQuery<T> = FilterQuery<T>;
 export type UwaziUpdateQuery<T> = UpdateQuery<DataType<T>>;
 export type UwaziQueryOptions = QueryOptions;
 
-export class OdmModel<T> {
+export interface SyncDBDataSource<T> {
```
fnocetti: Can we put this interface in the v2 side? So we minimize the V2-> V1 dependency.

fnocetti (changes requested):

app/api/i18n.v2/database/MongoTranslationsSyncDataSource.ts
daneryl: Throw an error instead of returnin an empty array.

app/api/common.v2/database/MongoDataSource.ts
daneryl: changed to a bulk write stream using the scoped collection without the updatelogs functionality, dont really love it since i had to change the type of the scope method to a generic and had to move the cache for the proxied collection away because the class now uses 2 different collections (it was always returning the original collection, never the new updatelogs one) which leads me to think that maybe the best option is for the "updatelogs" collection to have its own datasource instance ? is it a bad idea to have a concrete instance of itself in an abstract class ? @fnocetti @LaszloKecskes 

Edited: we decided to implement the full wrapper for mongo collection, i think this will be a non issue then.

app/api/i18n.v2/database/MongoTranslationsSyncDataSource.ts
daneryl: I agree its better but in this case we need to conform to the V1 interface for sync.

app/api/common.v2/database/MongoDataSource.ts
daneryl: solved by changing proxy implementation to a collection wrapper

app/api/common.v2/database/MongoDataSource.ts
daneryl: solved by changing proxy implementation to a collection wrapper

app/api/i18n/v2_support.ts
daneryl: solved by changing proxy implementation to a collection wrapper

daneryl: leaving it as a tech debt for now

app/api/odm/model.ts
daneryl: done !

app/api/common.v2/database/specs/MongoDataSource.spec.ts
```diff
@@ -260,32 +262,10 @@ describe('session scoped collection', () => {
     );
   });
 
-  it('should always return the same instance of the collection', () => {
+  xit('should always return the same instance of the collection', () => {
```
fnocetti: Excluded test

app/api/common.v2/database/MongoDataSource.ts
```diff
@@ -1,18 +1,15 @@
...
- */
-type MethodsOptionsArgPosition = {
-  [method in keyof Collection]: Collection[method] extends (...args: any) => any
-    ? MaxAmountOfParameters<Collection[method]>
-    : null;
-};
+import { SessionScopedCollection } from './SessionScopedCollection';
+import { SyncedScopedCollection } from './SyncedScopedCollection';
+
+export interface UpdateLog {
```
fnocetti: I think this doesn't belong here

app/api/common.v2/contracts/SyncDBDataSource.ts
```diff
@@ -0,0 +1,9 @@
+import { DeleteResult } from 'mongodb';
+
+export interface SyncDBDataSource<T> {
```
fnocetti: I think this should go to the database/infrastructure directory. It is only database related, and also knows about MongoDB :)

fnocetti (changes requested):
Other than a couple minor non-functional changes this LGTM. I think I'm going to do the changes myself so this doesn't stay alive for too long.

fnocetti:
After integrating the latest changes in production, that included a MongoDB upgrade that carried some API changes, I needed to change some type definitions in the wrappers. It didn't imply any functional changes.

app/api/common.v2/database/SyncedCollection.ts
```diff
@@ -0,0 +1,318 @@
...
+    db: Db
+  ) {
+    super(collection);
+    this.transactionManager = transactionManager;
+    this.db = db;
+  }
+
+  private async insertSyncLogs(mongoIds: ObjectId[]) {
+    if (mongoIds.length !== 0) {
+      await new SessionScopedCollection(
```
LaszloKecskes: Would it make sense to create this sessionsScopedCollection on init, then reuse?

app/api/common.v2/database/SyncedCollection.ts
```diff
@@ -0,0 +1,318 @@
...
+
+  private async upsertSyncLogs(conditions: any[], deleted: boolean = false) {
+    if (conditions.length === 0) {
+      return;
+    }
+
+    const modifiedDocuments = this.collection.find({ $or: conditions }, { projection: { _id: 1 } });
+
+    const stream = new BulkWriteStream(
+      new SessionScopedCollection(this.db.collection('updatelogs'), this.transactionManager)
```
LaszloKecskes: Same here as in line 62.

app/api/common.v2/database/SyncedCollection.ts
daneryl: done :+1:

app/api/common.v2/database/SyncedCollection.ts
daneryl: done :+1:

Issue #5477, opened 2023-01-27
IX: Suggestion text trimmed in a different way than the other columns


gabriel-piles:
**Describe the bug**
The suggestion text is trimmed in a different way than the other columns. All the columns should behave the same way. Could be all trimmed but with a higher character count.

 See attached image.

![Screenshot-20230127094314-1919x83](https://user-images.githubusercontent.com/22240564/215044624-2c6a501d-f819-4285-9c1d-e2c6c4c7f958.png)

txau:
@juanmnl @konzz @grafitto I think this is being addressed in the new UI for IX?

konzz:
Thanks for mentioning it, will check that this is fixed.

grafitto:
This should be fixed in the PR for new UI for IX #6056

Zasa-san:
This should be fixed now @gabriel-piles

PR #6056, opened 2023-07-26
IX suggestions UI Improvement

### LINKED ISSUE #5512: "Click to fill" can provide empty rectangles
**Describe the bug**
Using "Click to fill" in Firefox shows in most cases an alert. It happens in most cases using the mouse left button double click to select the text and pressing "Click to fill".

The alert says "Could not detect the area for the selected text, please try selecting again.

See the screenshot below.

Screenshot-20230206112743-1949x642

### LINKED ISSUE #5830: IX Updated UI

grafitto:
fixes #5830 #5512 #5477 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/V2/Routes/Settings/IX/components/FiltersSidepanel.tsx
```diff
@@ -0,0 +1,201 @@
...
+  };
+
+  let initialFilters: SuggestionCustomFilter = defaultFilter;
+
+  try {
+    if (searchParams.has('filter')) {
+      initialFilters = JSON.parse(searchParams.get('filter')!);
+    }
+  } catch (e) {}
+
```
Zasa-san, line 63: I've seen a couple of places where we have this try catch statements where nothing happens on the catch. Could they be removed?

app/react/V2/Routes/Settings/IX/IXSuggestions.tsx
```diff
@@ -0,0 +1,295 @@
...
+            revalidator.revalidate();
+          }
+        }
+      }
+    );
+
+    return () => {
+      socket.off('ix_model_status');
+    };
+  }, []);
```
Zasa-san: Should we add the recommended dependencies here?

app/react/V2/Routes/Settings/IX/components/FiltersSidepanel.tsx
grafitto: I purposely put that there to allow JSON.parse to fail silently. For example, if a user manually edited filter query string to an invalid JSON string, uwazi should ignore the filters in the query string and use the default filters

fnocetti:
On @Zasa-san 's request I reviewed the sorting-related changes on the backend and they LGTM!
One think to consider is that the way we agreed to implement the sort for this PR is not viable in the long term. We will need to use a denormalization of the entity title and current value in the suggestion to be able to index-sort. Let's discuss if we can live with this and a tech debt issue or we want to implement it right away.

PR #6060, opened 2023-07-31
Settings footer mobile bug

### LINKED ISSUE #5992: Settings footer visual errors
**Describe the bug**
Settings footer sometimes cover content with it's footer.

**To Reproduce**
This happens on any V2 settings

**Expected behavior**
We need to check the layout and how components stack to form the layout.

konzz:
fixes #5992

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/App/styles/globals.css
+++ b/app/react/App/styles/globals.css
@@ -2854,6 +2854,10 @@ input:checked + .toggle-bg {
   padding-bottom: 1.25rem;
 }
 
+.pb-14 {
+  padding-bottom: 3.5rem;
+}
+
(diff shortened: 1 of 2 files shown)


Zasa-san:
When instantiating the `<Settings.Footer />` element we should include a `flex-wrap` class so that the buttons will stack vertically on smaller screens, like this:
image

I don't know if its best to add it every time we use the footer, or add this class as a default in the definition of the component.

PR #6066, opened 2023-08-04
Async query recalculation schedules one single job

### LINKED ISSUE #6062: Improve job scheduling strategy
The relationships v2 functionality posts jobs to a jobs queue in order to process the query executions for the involved entities in an asynchronous way.
It posts a job per entity and we found that process to be a bottleneck in the flows of template updates.
We need to research better ways to handle that. The current idea is to post a single job that would then be in charge of creating new jobs for each entity or batches of entities. That way we don't reduce the global time, but we reduce the time the user needs to await after saving the template.

fnocetti:
fixes #6062 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti:
This was pair-reviewed with @daneryl. Merging.

PR #6093, opened 2023-08-25
5960 fix live translate on language addition

### LINKED ISSUE #5960: Adding a language  not working properly until page reload
I found this during my work on Translations v2 and testing on staging, i have tested this on the previous release to confirm it was happening before my changes.

**Describe the bug**
Adding a new language updates local redux store with the new language in the wrong format and the wrong values (in case of installing a language with predefined values), if you try to translate something you get a validation error from the server, fixed reloading the page after installing a language

**To Reproduce**
1. Create a blank state uwazi
2. Go to settings -> languages and add a new language
3. Try to live translate something
4. See Validation Error

**Expected behavior**
Should properly update the local languages and work without having to perform a page reload.

**Additional context**
As a bonus: the function that adds a language treats differently languages with and without predefined translations
https://github.com/huridocs/uwazi/blob/4f9e870274f07e0c6a3d4c87e1c7bc6d3edc70ed/app/api/i18n/routes.ts#L14-L27
Not 100% sure but this probably has to return the translations after applying the import predefined for the client to work properly without having to reload

LaszloKecskes:
fixes #5960 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/shared/asyncUtils.ts
+++ b/app/shared/asyncUtils.ts
@@ -8,7 +8,8 @@ const sequentialPromises = async <T>(
     index: number,
     originalArray: T[]
   ) => {
-    await previousPromise.then(async () => asyncCallback(element, index, originalArray));
+    await previousPromise;
+    await asyncCallback(element, index, originalArray);
   };
   await items.reduce(wrappedCallback, Promise.resolve());
 };
--- a/app/shared/asyncUtils.ts
+++ b/app/shared/asyncUtils.ts
@@ -1,0 +1,16 @@
+const sequentialPromises = async <T>(
+  items: T[],
+  asyncCallback: (element: T, index: number, originalArray: T[]) => Promise<void>
+): Promise<void> => {
+  const wrappedCallback = async (
+    previousPromise: Promise<void>,
+    element: T,
+    index: number,
+    originalArray: T[]
+  ) => {
+    await previousPromise.then(async () => asyncCallback(element, index, originalArray));
+  };
+  await items.reduce(wrappedCallback, Promise.resolve());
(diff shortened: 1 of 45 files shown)


app/shared/asyncUtils.ts
```diff
@@ -0,0 +1,16 @@
+const sequentialPromises = async <T>(
```
fnocetti, line 1: Since this takes a collection and an async function to be executed on each sequentially, I think this is actually an async version of a forEach. I would inspire the name from there.

app/shared/asyncUtils.ts
```diff
@@ -0,0 +1,16 @@
...
+  items: T[],
+  asyncCallback: (element: T, index: number, originalArray: T[]) => Promise<void>
+): Promise<void> => {
+  const wrappedCallback = async (
+    previousPromise: Promise<void>,
+    element: T,
+    index: number,
+    originalArray: T[]
+  ) => {
+    await previousPromise.then(async () => asyncCallback(element, index, originalArray));
```
fnocetti: The then callback can also be awaited:
```
await previousPromise;
await asyncCallback(element, index, originalArray);
```

fnocetti (changes requested):
Apart from minor comments, up to you to change, it LGTM and works as expected.

app/shared/asyncUtils.ts
LaszloKecskes: Done.

PR #6111, opened 2023-09-01
6097 update drag and drop library

### LINKED ISSUE #6097: Update Drag and Drop library
We are currently using an old version of the React Drag and Drop library which is preventing development of the Table ordering feature.  We need to upgrade it to the latest version.

There was a suggestion by @Zasa-san to create a wrapper for the old component so that we don't need to re-write everywhere this is used.

mfacar:
fixes #6097

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/V2/CustomHooks/useDnDContext.tsx
```diff
@@ -0,0 +1,180 @@
+import { useState } from 'react';
```
Zasa-san: Is this hook tested with an integration test? Does it merit to have a unit test for it?

app/react/componentWrappers.tsx
```diff
@@ -49,4 +49,35 @@ const withOutlet =
     const outlet = useOutlet(AppMainContext);
     return <Component {...props} outlet={outlet} />;
   };
-export { withRouter, withContext, withOutlet };
+
+const withLazy =
```
Zasa-san: Is this used? I can't find it referenced anywhere.

package.json
```diff
@@ -329,7 +332,7 @@
     "copy-webpack-plugin": "11.0.0",
     "css-loader": "^6.8.1",
     "css-minimizer-webpack-plugin": "^5.0.1",
-    "cypress": "12.17.4",
+    "cypress": "13.2.0",
```
Zasa-san: Is it neccesary to update cypress for this PR?

Zasa-san (changes requested):
So far so good! I’m still finishing code review. Some notes about manual ui testing:

Storybook:
- In the `Nested` story once an item is dragged into a list, it cannot be moved to another list. For example: drag available Item 4 as a subitem of the group Item 1, then try to move Item 4 to the group of Item 2, or Item 3, or as a new item group. Draggin available Item 4 from a sublist outside the boundary of Active Item throws an error. These behaviors should likely be tested.

Settings - Menu:
- There’s a problem when trying to double click the item name or the url. Usually double clicking paints the whole text, but now it does nothing. Same thing with click and drag to select a portion of text. Maybe the parent element is propagating some event that interrupts the normal behavior of the text field?

- The current design has lines separating items and item groups. 

- There is some odd behavior when adding items. It seems to update the names of the item based on the position? So if I have 3 items: Item A, Item B, and Group C, and I drag Group C to the second position, it gets renamed Item B

Settings - Filters:

- The delete buttons look off.
- The padding for groups is smaller on the right side, the delete button almost touches it.
- Trying to drag an item within a group throws an error.
- I cannot rename groups, they are always named as ‘Group one’

General:
- Perhaps we should check what should happen when an item is dragged from a list towards a space in the screen that is outside the boundaries of the drop area. Should it be considered a way of deleting draggable items? Could be a nice to have.

_(Tested in Firefox)_

app/react/V2/CustomHooks/useDnDContext.tsx
mfacar: It is tested through the component test DragAndDrop.cy.tsx, but for sure I can add some additional cases to cover more code.

app/react/componentWrappers.tsx
mfacar: It's not used now, I'll remove it. I was part of a previous implementation attempt 👍

package.json
mfacar: Not at all, was added during the problems to run cypress against chrome. I'll update to have the same as development.

mfacar:
After enabling two versions of React DnD to keep the compatibility for the current components that include draggable items, these files were reversed since they are projected to be migrated to Client V2

app/react/App.tsx
```diff
@@ -19,11 +21,13 @@ const recoilGlobalState = ({ set }: MutableSnapshot) => {
 
 const App = () => (
   <Provider store={store as any}>
-    <CustomProvider>
-      <RecoilRoot initializeState={recoilGlobalState}>
-        <RouterProvider router={router} fallbackElement={null} />
-      </RecoilRoot>
-    </CustomProvider>
+    <DndProvider backend={HTML5Backend}>
```
Zasa-san: Is it neccesary to wrapp the entire app? Could we not just use the provider when a route renders a table with dnd?

app/react/V2/Components/Layouts/DragAndDrop/Container.tsx
```diff
@@ -0,0 +1,72 @@
...
+interface ContainerProps<T> {
+  context: IDnDContext<T>;
+  itemComponent?: FC<IItemComponentProps<T>>;
+  iconHandle?: boolean;
+  className?: string;
+  parent?: IDraggable<T>;
+}
+
+// eslint-disable-next-line react/function-component-definition
+function Container<T>({
```
Zasa-san: If there's no impediment to write new components as arrow functions, could you refactor them?

app/react/V2/Components/Layouts/DragAndDrop/DropZone.tsx
```diff
@@ -0,0 +1,50 @@
...
+    }),
+  }));
+
+  const isActive = canDrop && isOver;
+  return (
+    <div
+      className={`flex flex-col items-center justify-center  text-gray-400 uppercase p-15 h-14 text-base m-5 ${
+        isActive
+          ? ' bg-bray-800 dark:bg-gray-700 bg-gray-100 dark:border-gray-600 dark:hover:border-gray-500 dark:hover:bg-gray-600'
+          : 'empty'
```
Zasa-san: I don't think `empty` is a valid tailwind class.

app/react/V2/CustomHooks/useDnDContext.tsx
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable max-statements */
```
Zasa-san: I don't believe this hook will ever be used outside of a drag and drop functionality. It would be ok in my opinion to move this inside the folder with the rest of the dnd components and functionality. That way you could import everything from `app/V2/Components/Layouts/DragAndDrop`.

app/react/componentWrappers.tsx
```diff
@@ -49,4 +49,34 @@ const withOutlet =
...
+  withLazy<T>(
+    Component,
+    async () => import('react-dnd'),
+    (module: any) => ({
+      useDrag: module.useDrag,
+      useDrop: module.useDrop,
+      useDragDropManager: module.useDragDropManager,
+    })
+  );
+export { withRouter, withContext, withOutlet, withDnD };
```
Zasa-san: `withLazy` and `withDnD` would be for V2, I would move it to the V2 folder.

cypress/support/component.ts
```diff
@@ -30,3 +30,15 @@ import { mount } from 'cypress/react18';
 Cypress.Commands.add('mount', mount);
 // Example use:
 // cy.mount(<MyComponent />)
+
+//@ts-ignore
+const check = components => {
```
Zasa-san, line 35: This could be typed as `ReactNode`, no? If not I rather have `any`

app/react/stories/dragAndDrop/DragAndDrop.stories.tsx
```diff
@@ -0,0 +1,255 @@
+/* eslint-disable react/no-multi-comp */
```
Zasa-san: I would move the story outside the folder, and maybe just use this folder for the components used in the story to simplify the file?

Zasa-san (changes requested):
Awesome work!

In the `Nested` story I’m noticing that if you delete an item with subitems (for example Item 1) you cannot reuse it. You have to clear the children first to avoid the bug.

app/react/V2/Components/Layouts/DragAndDrop/DropZone.tsx
mfacar: Definitely not, it's a few noise from the old implementation 😬

app/react/App.tsx
mfacar: I've moved the provider locally

cypress/support/component.ts
mfacar: The type certainly is ReactNode, but Cypress is throwing an Unexpected token error when the type is included. I'm leaving the exclusion until we can figure out how to handle the error.

PR #6116, opened 2023-09-01
6112 v2 authentication published flag

### LINKED ISSUE #6112: Incorrect published permission handling in v2 authentication
During  #6110 it was discovered that the v2 version of reading permissions from v1 entities handles the published flag incorrectly: it expects a "published" permission on the permissions array, instead of the published boolean on the entity body that is actually used. The task is:
 - write the missing tests
 - correct the usage
 - check that the places using permissions still work

LaszloKecskes:
fixes #6112

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/authorization.v2/database/schemas/permissionTypes.ts
```diff
@@ -3,20 +3,13 @@ import { ObjectId } from 'mongodb';
 export interface EntityPermissionsDBOType {
   _id?: ObjectId;
   sharedId: string;
+  published?: boolean;
```
fnocetti, line 6: Is it really optional? Does this come from the v1 schema or you actually found a case?

app/api/authorization.v2/database/schemas/permissionTypes.ts
```diff
@@ -3,20 +3,13 @@ import { ObjectId } from 'mongodb';
 export interface EntityPermissionsDBOType {
```
fnocetti: Do you think it is a good opportunity to adjust the naming to what we are doing on v2? This naming a legacy from using the types emitter, but we don't postfix `Type` anymore.
Also applies to the filename and the rest of the exported types.
Leave it to you to decide.

app/api/authorization.v2/model/EntityPermissions.ts
```diff
@@ -1,40 +1,37 @@
...
 
-  constructor(entity: string, permissions: Entry[]) {
+  readonly published: boolean;
+
+  constructor(entity: string, permissions: Entry[], published: boolean = false) {
     this.entity = entity;
     this.permissions = permissions;
-  }
-
-  allowsPublicReads() {
```
fnocetti, line 27: I think I would still like to have this method since is more semantic than just accessing `published`. Although it won't implement any complicated logic, I think it will go side by side with `allowUserTo`.
It would also allow us to change the "publish" implementation in a easier way in the future if we decide to.

app/api/relationships.v2/contracts/RelationshipsDataSource.ts
```diff
@@ -8,6 +8,7 @@ export interface RelationshipsDataSource {
   delete(_ids: string[]): Promise<void>;
   deleteAll(): Promise<void>;
   exists(_ids: string[]): Promise<boolean>;
+  getAll(): ResultSet<Relationship>;
```
fnocetti: I saw this is only used in a test. I think that until `getAll` is an application requirement it shouldn't be in the contract. Since that test is an integration test, and you are already instantiating a Mongo implementation, it's safe to say that you can simply use the method from the implementation (which in fact you are) and not include it here :)

app/api/users.v2/model/User.ts
```diff
@@ -17,3 +17,6 @@ export class User {
     return ['admin', 'editor'].includes(this.role);
   }
 }
+
+export type { UserRole };
```
fnocetti, line 21: Particularly in the case of the domain models, I think it's usually more readable and modular to instead of exporting internal types, just reference them relative to the model. In this case `User['role']`.
But at this point it's just an opinion :)

app/api/utils/fixturesFactory.ts
```diff
@@ -82,6 +85,19 @@ function getFixturesFactory() {
       commonProperties,
     }),
 
+    entityPermission: (
+      user?: string,
+      type?: PermissionSchema['type'],
+      level?: PermissionSchema['level']
+    ) => {
+      const isPublic = !user || user === 'public';
+      return {
```
fnocetti: Wasn't this what we were trying to remove? Surely you have good reasons for this, let's talk about it.

fnocetti (changes requested):
Left some comments on the code, overall it looks good. I'm proceeding to manual testing.

app/api/authorization.v2/database/schemas/permissionTypes.ts
LaszloKecskes: Yes, it is optional in the v1 schema.

app/api/authorization.v2/database/schemas/permissionTypes.ts
LaszloKecskes: Done.

app/api/authorization.v2/model/EntityPermissions.ts
LaszloKecskes: Done.

app/api/relationships.v2/contracts/RelationshipsDataSource.ts
LaszloKecskes: Done.

app/api/utils/fixturesFactory.ts
LaszloKecskes: No, actually this is just a remainder from yesterday, when I've still tried to have that, so I'll just need to remove this from here as well. Nice catch!

fnocetti (approved):
Looks good and works as expected

PR #6137, opened 2023-09-07
5469 numeric fields validation

### LINKED ISSUE #5469: Numeric fields are not properly validated in post api/entities
**Describe the bug**
If you send a string value on a metadata property that is configured as numeric, the endpoint will not give a validation error and save the string in the DB.

**To Reproduce**
1. Configure a template with a numeric value
2. Use curl or other tool to save an entity with the numeric property filled with a string
3. A string is now saved on a property that should be a number

fnocetti:
fixes #5469 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/migrations/migrations/142-parse-numeric-fields/index.ts
```diff
@@ -0,0 +1,53 @@
...
+      Object.keys(entity!.metadata || {}).forEach(propertyName => {
+        if (template.properties[propertyName].type === 'numeric') {
+          const candidateValue = entity!.metadata?.[propertyName]?.[0].value;
+          if (typeof candidateValue === 'string') {
+            updates[`metadata.${propertyName}`] = [{ value: parseFloat(candidateValue) }];
+          }
+        }
+      });
+
+      if (Object.keys(updates).length > 0) {
```
LaszloKecskes: Consider using bulk write with some batch size. In my experience that can speed up entity based migrations a lot.

LaszloKecskes (changes requested):
Aside from the one comment, it looks good and works!

app/api/migrations/migrations/142-parse-numeric-fields/index.ts
fnocetti: Nice catch, done!

LaszloKecskes (approved):

LaszloKecskes (approved):

Issue #6110, opened 2023-08-31
Write tests for relationships v2 migration


LaszloKecskes:
As mentioned in the previous relationships v2 migration issues (#5928, #6002), the migration was developed without tests to make the experimental phase faster. Now that the methodology is considered final, the tests should be written.

PR #6149, opened 2023-09-18
6110 add missing tests for commonv2

LaszloKecskes:
Part of, but does not completely solve #6110.

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/common.v2/database/specs/MongoIdGenerator.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+  describe('generate', () => {
+    it('should generate a string id', () => {
+      expect(MongoIdHandler.generate()).toEqual(expect.any(String));
+    });
+  });
+
+  describe('mapToDb', () => {
+    it('should map a string id to a Mongo ObjectId', () => {
+      const id = MongoIdHandler.generate();
+      expect(MongoIdHandler.mapToDb(id)).toEqual(expect.any(ObjectId));
```
fnocetti: This should test the implementation. That is: that the id is mapped to the correct string. 
Maybe manually create an ObjectId so you can beforehand know the string without having to reimplement the mapper for the sake of testing.

app/api/common.v2/database/specs/MongoIdGenerator.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    });
+  });
+
+  describe('mapToApp', () => {
+    it('should map a Mongo ObjectId to a string', () => {
+      const id = MongoIdHandler.generate();
+      expect(MongoIdHandler.mapToApp(MongoIdHandler.mapToDb(id))).toEqual(id);
```
fnocetti: Same as previous.

app/api/authorization.v2/database/specs/data_source_defaults.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+});
+
+afterAll(async () => {
+  await testingEnvironment.tearDown();
+});
+
+describe('data_source_defaults', () => {
+  describe('DefaultPermissionsDataSource', () => {
+    it('should return a MongoPermissionsDataSource', () => {
+      expect(DefaultPermissionsDataSource()).toBeInstanceOf(MongoPermissionsDataSource);
```
fnocetti, line 16: Now it's done, that's fine. But I don't know if this merits tests. This is exactly what types are for.

fnocetti (changes requested):
Overall looks good! Only a few comments.

app/api/common.v2/database/specs/MongoIdGenerator.spec.ts
LaszloKecskes: Good one, will do.

app/api/common.v2/database/specs/MongoIdGenerator.spec.ts
LaszloKecskes: Same as previous. :)

app/api/authorization.v2/database/specs/data_source_defaults.spec.ts
LaszloKecskes: I think, strictly speaking, on the type level the defaults return the interface, because the users (i.e. the services) need only know about that aspect. I think there's value in having tests that explicitly say in the code base that at this point in time, the expected default we have decided on is the Mongo version of whatever DS we look at. At least as a reminder if we ever have anything else and decide to change.

fnocetti (approved):

PR #6165, opened 2023-09-20
5965 limit root elements in aggregations

### LINKED ISSUE #5965: How to Display DKI Jakarta Province in the Lokasi Kejadian Entity
Dear Huridocs,

I want to ask a few things. Associated with the **Lokasi Kejadian** entity

https://kebebasanberkesenian.uwazi.io/in/library/?q=(includeUnpublished:!t,order:desc,sort:creationDate,treatAs:number,unpublished:!f)

- **Lokasi Kejadian** >>> DKI Jakarta does not appear in the filter. It only appears in Search menu, so it doesn't exist on the Chart either.

As shown in the following chart chart:

**Lokasi Kejadian**

https://kebebasanberkesenian.uwazi.io/in/page/e5ebwy36f6t/bar-chart-2

There are only 12 provinces. For **DKI Jakarta** province, none. Even if you search, there are 14 updates of **DKI Jakarta**.

Please help me regarding this issue. Thank You

ps : Thesaury: **Lokasi Kejadian1**

Best regard,

Andri

LaszloKecskes:
fixes #5965 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/shared/comonProperties.js
```diff
@@ -72,11 +72,14 @@ const allUniqueProperties = templates =>
     return filters;
   }, []);
 
-const allProperties = templates =>
-  templates.reduce((properties, template) => properties.concat(template.properties), []);
+const allProperties = templates => templates.map(template => template.properties || []).flat();
 
-const getInheritedProperty = (property, properties) =>
-  properties.find(p => property.inherit.property.toString() === p._id.toString());
+const getInheritedProperty = (property, properties, propertiesById) => {
```
fnocetti: I like the idea, but I think I would "overload" it instead, by making the `properties` parameter expect either an array or an object and treat it accordingly, so you don't have to pass both.

app/api/search/specs/search.spec.js
```diff
@@ -542,6 +557,52 @@ describe('search', () => {
...
+    });
+
+    it('should limit the number of root buckets only', async () => {
+      userFactory.mock(undefined);
+      const defaultResponse = await search.search(
+        { types: [ids.templateMetadata1, ids.templateMetadata2], allAggregations: true },
+        'en'
+      );
+      expect(defaultResponse.aggregations.all.groupedDictionary.buckets).toMatchObject([
+        { key: 'EuropeID', values: [{ key: 'GermanyID' }, { key: 'ItalyID' }] },
```
fnocetti: I would try to add a third value to the group so this also test that the limit only applies to the parents.

app/api/search/search.js
```diff
@@ -284,57 +293,154 @@ const _getAggregationDictionary = async (aggregation, language, property, dictio
...
+    }
+    dictIndex += 1;
   }
-  return Object.assign(aggregation, { buckets });
+
+  if (missingBucket) newBuckets.push(missingBucket);
+  return newBuckets;
+};
+
+const limitBuckets = (buckets, _limit) => {
```
fnocetti, line 340: I know we have this kind of uses of the `_` prefix in the codebase, but I would try to move away from it since it's also the convention for unused paramenters.

app/api/search/search.js
```diff
@@ -284,57 +293,154 @@ const _getAggregationDictionary = async (aggregation, language, property, dictio
...
 
-      const buckets = aggregations[key].buckets
-        .filter(bucket => role === UserRole.ADMIN || refIds.includes(bucket.key))
+      let buckets = limitBuckets(aggregations[key].buckets, limit);
+      buckets =
+        role === UserRole.ADMIN ? buckets : buckets.filter(bucket => refIds.includes(bucket.key));
+      buckets = buckets
         .map(bucket => {
-          const itemInfo = info.find(i => i.refId.toString() === bucket.key);
+          const itemInfo = bucket.key in infoByRefId ? infoByRefId[bucket.key] : null;
```
fnocetti: If I understand the purpose of this check correctly, it should have the same result as just `const itemInfo = infoByRefId?.[bucket.key]` since the possible `null` doesn't add to the algorithm.

app/api/search/search.js
```diff
@@ -346,13 +452,18 @@ const _denormalizeAggregations = async (aggregations, templates, dictionaries, l
         })
         .filter(b => b);
 
-      return Object.assign(denormaLizedAgregations, { [key]: { ...aggregations[key], buckets } });
+      denormalizedAggregations[key] = { ...aggregations[key], buckets };
+      return;
     }
 
-    let property = properties.find(prop => prop.name === key || `__${prop.name}` === key);
+    let property = key in propertiesByName ? propertiesByName[key] : null;
```
fnocetti: Same as above regarding the use of `in`

fnocetti (changes requested):
Left a couple comments, overall looks good. Proceeding to testing.

fnocetti:
I don't know if this was introduced by this PR or it was already there, but I found out that the filters for selects of grouped thesauri are not sorting by greatest values. In the other hand, when the thesauri is not grouped it does sort the buckets so the ones containing the more items (more relevant buckets) appear first.
This doesn't mean that the buckets **shown** in the filter are not sorted, they are. But looks like they're not being sorted before being limited, since buckets with big counts get lost if off limits.

Other than that, it works as expected.

app/shared/comonProperties.js
LaszloKecskes: Done.

app/api/search/specs/search.spec.js
LaszloKecskes: Done.

LaszloKecskes:
> I don't know if this was introduced by this PR or it was already there, but I found out that the filters for selects of grouped thesauri are not sorting by greatest values. In the other hand, when the thesauri is not grouped it does sort the buckets so the ones containing the more items (more relevant buckets) appear first. This doesn't mean that the buckets **shown** in the filter are not sorted, they are. But looks like they're not being sorted before being limited, since buckets with big counts get lost if off limits.

I think it was present. The partner noticed the original error of this pr, because the group that was cut was considered significant by them.

fnocetti (approved):
LGTM!

PR #6190, opened 2023-10-06
When syncing a default template, set the previous default to non-default

### LINKED ISSUE #5096: Sync can cause the target host to have 2 Default templates.
**Describe the bug**
Sync can cause the target  to have 2 Default templates when the target has already a default template that is not the same as the host.

**To Reproduce**
1. Create 2 tenants
2. Each tenant with a different Default template. Ensure that the default template is not the `Document` template that comes with the blank state, but a brand new one for each tenant.
3. Sync the default template from tenant host to target
4. You should see 2 default templates on the target tenant

**Expected behavior**
Only the already default template on the Target should remain as default.

**Screenshots**
Screenshot 2022-09-12 at 13 05 41

fnocetti:
fixes #5096 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/sync/routes.ts
+++ b/app/api/sync/routes.ts
@@ -9,6 +9,7 @@ import { Application, Request } from 'express';
 import { TranslationType } from 'shared/translationType';
 import { FileType } from 'shared/types/fileType';
 
+import { TemplateSchema } from 'shared/types/templateType';
 import { needsAuthorization } from '../auth';
 
 const diskStorage = multer.diskStorage({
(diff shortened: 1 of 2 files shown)


LaszloKecskes (approved):
Looks good, works.

fnocetti:
The cypress e2e are failing because of a known issue and no failing tests are related to the changes in this PR. Merging.

PR #6194, opened 2023-10-10
6000 csv import column name clash

### LINKED ISSUE #6000: Resolve 'language' name clash in entity csv import
Currently, at some point of the flow our csv import tries to arrange the base descriptors of an entity (title, language), and the imported metadata in one object:
```
{ title: 'A book', language: 'en', a_property: 'a value' }
```
If a template has a property called "Language", the core language descriptor overwrites the imported value. For example, even if you've tried to import the Language "English" through the csv, the system would overwrite that with the default 'en'. This can cause a couple of problems:
- for text fields, wrong values are imported
- for selects and multiselects, it tries to look up something from the thesauri that does not exist, and the process fails with an error

An easy workaround until the fix is to rename the affected properties.

Note: while this seems to be a very improbable edge case, we've noticed this, because it has been reported by one our PM's, so it has already happened at least once.

LaszloKecskes:
fixes #6000 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti (commented):
@LaszloKecskes the changes LGTM, but it makes me wonder: can this happen with the sharedId as well?
I will proceed to the manual testing now, I guess I'll find out :P

fnocetti (approved):
Works as expected. Just in case, I tested the sharedId and Date added columns and they seem not to be affected by this issue.

LaszloKecskes:
We are currently ignoring cypress tests on the CI, if they pass locally. Merging.

PR #6196, opened 2023-10-12
6195 improve transaction abort

### LINKED ISSUE #6195: Fix transaction aborted check in TransactionManager
On #5942 we have detected and fixed that the TransactionManager tried to abort and already aborted transaction. We do that by checking that the error code is 251, which we assumed was the "transaction aborted" code.
We have detected issues in production that seem related to the fact that the server returned  a retry-able transaction transient error, while also using 251 but not aborting the transaction. That implies that the transaction retry logic will try to start a new transaction, which is giving us a "Transaction already in progress" error.
After a lot of research, it looks like the correct way to check if we need to manually close the transaction is not checking the error code, but asking the session if it has a transaction in progress. i.e. `session.inTransaction()`.

Given these scenarios are hard to unit tests we have discussed that, even if these kinds of retryable errors are considered operational, we want to log them with some kind of severity to be hable to easily debug in the future.

fnocetti:
fixes #6195 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


LaszloKecskes:
I think switching to the default transaction manager in the higher level tests (services, v2_support, routes, etc) is good, since those should be infrastructure agnostic and work with the defaults at any time.

However, for the lower level DataSource implementations (Mongo<something>DataSource) I think it makes sense to pass the specific transaction manager implementation that specific DS needs. Unless we see ourselves ever implementing another Mongo transaction manager for any reason, and we want to make sure that all tests always use the same.

LaszloKecskes (approved):
Looks good, the comment is optional.

daneryl (approved):
LGTM !

PR #6214, opened 2023-10-19
6170 handle default translation csv errors

### LINKED ISSUE #6170: No error shown when csv of language translation is malformed
**Describe the bug**
When installing a language with default translations if the csv is malformed, the app will report the operation as successful, but what happens is that the language gets added but the default translations do not.

**To Reproduce**
1. Modify the header of `contents/ui-translations/ru.csv`, change the first word `Key` to `key` in lowercase
2. Run uwazi
3. Go to settings -> languages -> add Language
4. Add Russian
5. See a successfull feedback
6. Change uwazi to use Russian
7. You can now see that everything is in English

**Expected behavior**
Default Russian translations should be imported.

**Additional context**
related #6157, #6169

LaszloKecskes:
fixes #6170

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Code review
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing


app/api/i18n/specs/defaultTranslations.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+  });
+
+  it.each([{ key: 'en' }, { key: 'ar' }])(
+    'should expect the file to have 2 columns',
+    async ({ key }) => {
+      await expectThrow(
+        async () => translations.importPredefined(key),
+        ValidateFormatError,
+        'Expected 2 columns, but found 3.'
+      );
```
daneryl: There is a builtin solution for this, or maybe you tried and it does not work for you ? Something like this.
```typescript
await expect(async () => translations.importPredefined(key)).rejects.toEqual(
  new ValidateFormatError('Expected 2 columns, but found 3.')
);

```

daneryl (approved):
LGTM, only a comment on the expectThrow jest matcher

app/api/i18n/specs/defaultTranslations.spec.ts
LaszloKecskes: Yes, this will work. Thanks!

PR #6226, opened 2023-10-25
6217 any template relationship deletion fix

### LINKED ISSUE #6217: Entity remains in relationship fields after being deleted
**Describe the bug**
When an entity related to others gets deleted the values in those other entities keep the reference to the deleted one. This value is present in the edition of the entity with the inconsistency, if the Save button is clicked the action doesn't get completed and the buttons remain disabled. Bulk edition actions like the edition of the template fix the problem.

**To Reproduce**
1. Create an entity with any template <A>
2. Create a template <T> with a Relationship field <R>
3. Create an entity with the template T with the field R pointing to A
4. Delete entity A
5. View entity B and see the value of R, A is still in field R

**Expected behavior**
All the references to A should be removed with its deletion.

**Screenshots**

_Before deleting entity_
Screenshot 2023-10-20 095505
_After deleting entity A_
Screenshot 2023-10-20 095617

LaszloKecskes:
fixes #6217 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/entities/entities.js
```diff
@@ -732,10 +732,11 @@ export default {
     const properties = allProperties.filter(p => propTypes.includes(p.type));
     const query = { $or: [] };
     const changes = {};
+    const contentMatches = p =>
+      (p.content && p.content.toString() === propertyContent.toString()) ||
+      (!p.content && p.content === '');
```
fnocetti: The property schema says that the content is optional. Why are we allowing falsy values only when they're empty strings?

fnocetti (changes requested):
It doesn't work for me. In line 730 it is trying to load the templates that have a property with that content configured, which is not the case, since the only property I have is the one described in the issue with no content.

LaszloKecskes:
> It doesn't work for me. In line 730 it is trying to load the templates that have a property with that content configured, which is not the case, since the only property I have is the one described in the issue with no content.

And I did not notice this, because the test case in the fixtures had two relationships, and while doing the manual testing, I did the same. Embarrassing. Thanks!

app/api/entities/entities.js
LaszloKecskes: As usual, the content is optional, because we're handling every property in one schema, so everything is optional aside from the shared base (label, name, type). When we're creating the relationship to any template, an empty string gets written to the db, so a not-existing content means a property where content does not have a meaning, rather than the any template case.
However, empty strings are falsy anyway so you are right, this is redundant.

fnocetti:
It looks good and it works as expected. I was just thinking that this may need a migration. Let's discuss it with the team.

LaszloKecskes:
> It looks good and it works as expected. I was just thinking that this may need a migration. Let's discuss it with the team.

After the discussion, we have decided to add the migration.

app/api/migrations/migrations/149-remove_inconsistent_relationships_metadata/specs/fixtures.ts
```diff
@@ -0,0 +1,243 @@
...
+
+const untypedRelationshipProperty = {
+  name: 'relationship_untyped_target',
+  label: 'relationship_untyped_target',
+  type: 'relationship' as 'relationship',
+  relationType: '',
+  content: ids.sourceTemplate.toString(),
+};
+
+const noReindexFixtures: FixtureType = {
```
fnocetti, line 29: What does `noReindexFixtures` mean?

fnocetti (approved):
Appart from the comment, approved. Looks good and works.

app/api/migrations/migrations/149-remove_inconsistent_relationships_metadata/specs/fixtures.ts
LaszloKecskes: If a migration might involve reindexing, I usually try to test a case where it should not reindex. In this case, if none of the entities needed changing, we should not reindex. These are the part of the fixtures that will remain unchanged. We could remain it to something else to make it more semantic.

fnocetti: That makes sense, thanks!

PR #6239, opened 2023-11-06
Reseting password unlocks user

### LINKED ISSUE #6179: Resetting password does not unlock an account
**Describe the bug**
It appears that resetting the password does not unlock the account being reset.  The password appears to be reset, but the account itself is kept on the locked status until you click on the "unlock account" link sent on the email when this happened.

**To Reproduce**
1. Block your user by logging in multiple times with a wrong password
2. Reset the password (it works)
3. Try to login with the new password
4. See error (account is still locked, you cannot log in)

**Expected behavior**
This is something we should discuss, but if we decide that the only way to re-activate an account is by following the link, we should at least be able to re-send that email somehow?  Maybe resetting the password (which needs to be done with an associated email account) is enough to unblock the user?

cc @juanmnl @txau

fnocetti:
fixes #6179 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


LaszloKecskes (commented):
Looks good, works.

LaszloKecskes (approved):

PR #6269, opened 2023-11-23
Updates relationships with the latest entities when creating new entity

### LINKED ISSUE #6244: List of entities not updated in a relationship property
**Describe the bug**
When create entities, the list of entities to select in relationship properties are not updated in two cases:
- When adding an entity from a submit form and after, create an entity that uses this data in a relationship property.
- When deleting an entity and after, create an entity that uses this data in a relationship property.

**To Reproduce**
Case 1 - Adding data with a submit form:
1. Go to a submit form and add an entity, as usual.
2. Click on "Create entity" and select a template with a relationship propperty related with the templates from the first form.
3. See error: the added entity with the form does not appear in the list of entities.
4. Once you refresh the browser screen, the entity appears correctly.

Case 2 - Deleting an entity:
1. Delete any entity from a template related with a relationship property in another template.
2. Click on "Create entity" and select a template with a relationship propperty related with the deleted entity.
3. See error: the deleted entity still appears in the list of entities.
4. If you select this entity and save, Uwazi gets locked.
5. Once you refresh the browser screen, the deletd entity does not appear.

**Expected behavior**
When creating an entity, if it has relationship properties, the list of entities to be selected must be updated.

**Screenshots**
Case 1:
image

Case 2:
image

 - [X] Desktop
 - [x] Mobile

**Browser**
Tested with Brave and Chrome.

**Additional context**
This bug is not easy to explain, please let me know if it's better to meet and show it with a demo.

grafitto:
fixes #6244

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Metadata/actions/actions.js
```diff
@@ -140,12 +141,13 @@ export function loadInReduxForm(form, _entity, templates) {
...
       const entity = { ...response, attachments: sortedAttachments };
       loadFetchedInReduxForm(form, entity, templates).forEach(action => dispatch(action));
+      dispatch(reloadThesauri());
     });
   };
 }
 
 export function changeTemplate(form, templateId) {
-  return (dispatch, getState) => {
+  return async (dispatch, getState) => {
```
Zasa-san, line 150: Why is this async?

app/react/Metadata/actions/actions.js
```diff
@@ -10,6 +10,7 @@ import searchAPI from 'app/Search/SearchAPI';
 import { actions } from 'app/BasicReducer';
 import { generateID } from 'shared/IDGenerator';
 import emptyTemplate from '../helpers/defaultTemplate';
+import { reloadThesauri } from 'app/Thesauri/actions/thesaurisActions';
```
Zasa-san: Check the load order for this, it should be higher.

app/react/Metadata/actions/actions.js
Zasa-san: Is the `async` here because of the `setTimeout` below? I don't understand this change =/

app/react/Metadata/components/MetadataFormFields.js
```diff
@@ -414,6 +414,8 @@ class MetadataFormFields extends Component {
 }
 
 MetadataFormFields.defaultProps = {
+  saveThesaurus: () => {},
```
Zasa-san, line 417: Both of these props are required, do they need defaults?

cypress/e2e/share-entities.cy.ts
```diff
@@ -103,9 +104,10 @@ describe('Share Entities', () => {
 
   it('should be able to edit and save', () => {
     cy.contains('h2', titleEntity4).click();
-    cy.get('aside.is-active').contains('button', 'Edit').click();
+    // cy.get('aside.is-active').contains('button', 'Edit').click();
```
Zasa-san: Can this be removed?

cypress/e2e/share-entities.cy.ts
```diff
@@ -103,9 +104,10 @@ describe('Share Entities', () => {
...
+    clickOnEditEntity();
     cy.get('aside.is-active textarea').eq(0).clear();
-    cy.get('aside.is-active textarea').eq(0).type('Edited title');
+    cy.get('aside.is-active textarea').eq(0).type('Edited title', { force: true });
```
Zasa-san, line 109: I see a lot of `{force: true}`. Is something blocking the UI now?

cypress/test_files/short-video.webm
Zasa-san, line 1: We already have a short-video.mp4. Can we use that one instead of adding a new test file? Or better yet, remove the .mp4 since .webm is a better format.

Zasa-san (changes requested):
Functionality looks good! Nice work.

It should be pointed out that when deleting an entity that’s related to another one, the sidepanel and cards will not refresh on their own.

You can see it in the screenshot below, "Jack" is deleted but it’s still visible on the card and sidepanel:

image

This is likely not part of the scope for this issue, we should create a separate one to flag this problem.

On public forms new submitted entities will not appear either. It’s quite obvious when the template can relate to itself, the list of entities is not updated with the latest submission. I think this is ok since a public form should be only used by visitors that are not logged in, and therefore would never have access to a recently created entity since it’s created as private. But we should check if this is the intended functionality.

app/react/Metadata/components/MetadataFormFields.js
grafitto: I added this coz the linter was complaining that its undefined

cypress/e2e/share-entities.cy.ts
grafitto: There is this wierd behavior by cypress, sometimes when typing in an input or a textarea, cypress does not type the whole text before continueing with the test. Using `{force: true}` seems to be a quick fix atleast for now, the internet seems to not have a clear way of fixing that problem

grafitto:
> Functionality looks good! Nice work.
> 
> It should be pointed out that when deleting an entity that’s related to another one, the sidepanel and cards will not refresh on their own.
> 
> You can see it in the screenshot below, "Jack" is deleted but it’s still visible on the card and sidepanel:
> 
> image
> 
> This is likely not part of the scope for this issue, we should create a separate one to flag this problem.
> 
> On public forms new submitted entities will not appear either. It’s quite obvious when the template can relate to itself, the list of entities is not updated with the latest submission. I think this is ok since a public form should be only used by visitors that are not logged in, and therefore would never have access to a recently created entity since it’s created as private. But we should check if this is the intended functionality.

I will create another issue for this, just to be able to close this as it has taken too long

PR #6306, opened 2023-12-11
6128 grouped thesauri aggregations improvement

### LINKED ISSUE #6128: Analyse and improve grouped thesauri aggregations
At the time of first implementing support for the grouped thesauri aggregations we were only storing the final/child thesaurus value in the entities. Needing counts for parent values, and not having them indexed resulted in this kind of aggregations:
```
"select_1.value": {
          "filters": {
            "filters": {
              "parent1_id": {                <==== bucket for parent thesaurus value
                "terms": {
                  "select_1.value": [
                    "parent_1_child_1_id",
                    "parent_1_child_2_id",
                  ]
                }
              },
              "parent_1_child_1_id": {         <==== bucket for first thesaurus child
                "terms": {
                  "metadata.localidad_de_nacimiento__ca_.value": [
                    "parent_1_child_1_id"
                  ]
                }
              },
              "parent_1_child_2_id": {.     <===== bucket for second thesaurus child
                "terms": {
                  "select_1.value": [
                    "parent_1_child_2_id"
                  ]
                }
              },
```
In this query we aggregate on the select_1 values, and then manually create one bucket per thesaurus (child or parent), where the parent is defined as the union of the children.
This can quickly grow to a very big query proportional to the size of the thesaurus. It's also likely that it may introduce some performance degradation given the amount of filters used.

That changed and now we have also the parent thesauri for every value denormalized into the entity.
This should allow us to aggregate based on the parent (for parent aggregations) instead of relying in the child/final values, with the usual `terms` construction.

After a bit of research, it was found that while the denormalization for the parent thesaurus value is stored and indexed in elastic with the entity, it's being mapped as `{ value: text, label: text }` instead of the usual `{ value: keyword, label: text }`, which is the usual for this type of data. That needs to change in order to be able to aggregate the terms over `parent.value`.

LaszloKecskes:
fixes #6128

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/migrations/migrations/153-repair_select_parent_denormalization/index.ts
```diff
@@ -0,0 +1,245 @@
...
+const thesauriByProperty: Record<string, string> = {};
+
+const getPropertyThesaurusMap = async (db: Db) => {
+  const templateCollection = db.collection<Template>('templates');
+  const allTemplates = await templateCollection.find({}).toArray();
+
+  const selectProperties = _.flatMap(allTemplates, t => t.properties?.filter(isSelect) || []);
+  selectProperties.forEach(property => {
+    if (property.content) thesauriByProperty[property.name] = property.content;
+    if (property.content && property._id) {
```
fnocetti: Why is this pair of `if`s needed? I see that you have properties without _id in the fixtures, but is that the case for actual databases? Shouldn't properties always have an _id?

app/api/migrations/migrations/153-repair_select_parent_denormalization/index.ts
```diff
@@ -0,0 +1,245 @@
...
+type ParentInfo = Record<string, Record<string, ThesaurusValueBase>>;
+
+let parentInfo: ParentInfo = {};
+
+const getParentInfo = async (db: Db): Promise<boolean> => {
+  const thesaurusCollection = db.collection<Thesaurus>('dictionaries');
+  const thesauri = await thesaurusCollection.find({}).toArray();
+  let parentsExist = false;
+  parentInfo = Object.fromEntries(
+    Object.entries(_.keyBy(thesauri, '_id')).map(([id, thesaurus]) => {
```
fnocetti: Can't the same result be achieved by mapping the `thesauri` array? What's the benefit of first converting it to a hashmap?

app/shared/commonProperties.js
```diff
@@ -7,6 +7,15 @@ function sameProperty(prop1, prop2) {
   );
 }
 
+function isSelect(property) {
```
fnocetti: I understand this is a new function, but the general concept in v1 is that if it inherits a (multi)select, it is a relationship property. I would rename it accordingly (something like `isOrInheritsSelect`) to avoid confusion.

scripts/checkParentDenormalization.mjs
```diff
@@ -0,0 +1,151 @@
+/* eslint-disable max-statements */
+/* eslint-disable no-continue */
+/* eslint-disable no-await-in-loop */
+/*
+Run this script simply with node, from the project root folder: node ./scripts/checkParentDenormalization.mjs
```
fnocetti: I haven't tried yet, but just a comment: If this is useful enough to keep, we should ensure that it works with yarn as a traditional script.

fnocetti (commented):
It looks mostly good. I left a couple comments, not requesting changes, leaving it up to you. I'll proceed to the testing.
EDIT: by "mostly" I mean, apart from the comments :)

app/api/migrations/migrations/153-repair_select_parent_denormalization/index.ts
LaszloKecskes: Theoretically yes. Technically no, the type has _id as optional, and I did not want to have that much difference in the migration types and actual types. I think having a subset is fine, but assuming an optional thing non-optional is not, even if it's most likely just the types.
... but I should add the id's to the missing ones in the fixtures.

LaszloKecskes: Done.

app/api/migrations/migrations/153-repair_select_parent_denormalization/index.ts
LaszloKecskes: Of course, changed. This was probably left here only after I was experimenting with other data formats.

app/shared/commonProperties.js
LaszloKecskes: Done.

scripts/checkParentDenormalization.mjs
LaszloKecskes: I ended up removing it after some deliberation. It does not use uwazi code, so it would only be a situational script we would need to separately maintain so it matches the db structure. It's not worth it.

fnocetti (approved):
The new changes LGTM. Tested the functionality including permissions and it works as expected. Nice work!

PR #6309, opened 2023-12-11
Allow entity metadata to create relationships v2

### LINKED ISSUE #5347: Allow relationship fields v2 to be writable
We have discussed that we are going to allow the user to create relationships v2 by updating the list of selected values in the metadata relationships property v2.
It is only going to be allowed if the property has a query that allows to completely define the underlying to-be-created relationship by providing the entity on the other side.
Queries of that kind have the following shape:
- They are at most 1 level deep.
- They can have many branches as long as the templates on the other end only appear in one branch, and each branch has only one relationship type.

i.e.:
```
() -[relType1] -> (template1, template2)
 \
   -[relType2] -> (template3)
```

Since an empty list of templates comprehensively means "all templates", the only query with that kind of match is:

```
() -[relType1]-> ()
```

fnocetti:
fixes #5347 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/shared/types/api.v2/templates.createTemplateRequest.ts
```diff
@@ -16,6 +16,7 @@ export interface RelationshipPropertyData {
   name: string;
   query: TraverseQuery[];
   denormalizedProperty?: string;
+  targetTemplates?: string[];
```
LaszloKecskes: During functional testing, I've found that you cannot re-save templates with an "ineligible" property, because "targetTemplates must be an array". Probably this line is the reason.

Specifically, to reproduce: with any ineligible-for-editing property, you can initially save it, then after reload, targetTemplates is going to be false, and you cannot save the template again.

app/api/files/specs/v2_support.spec.ts
```diff
@@ -8,22 +8,45 @@ const factory = getFixturesFactory();
...
-          value: 'entity3',
-          label: 'entity3',
-        },
-      ],
-    }),
-    factory.entity('entity2', 'template1', {
-      relProp: [{ value: 'entity3', label: 'entity3' }],
-    }),
-    factory.entity('entity3', 'template1'),
+    factory.entity(
```
LaszloKecskes: Just to check if I read it correctly: here the obsoleteMetadata was added, and a relprop to the third one, the rest is prettier formatting. Since this is a standalone test not connected to the rest, what happened? Did it throw an error after the changes?

app/api/entities/specs/v2_newRelationshipMetadata.spec.ts
```diff
@@ -249,16 +317,38 @@ describe('entities.save()', () => {
...
+            entity: 'entity1',
+          },
+          to: {
+            entity: 'entity5',
+          },
+        }),
+      ])
+    );
+
+    expect(saved.metadata).toEqual({
```
LaszloKecskes: Could we test for the other language here as well, in the metadata?

app/api/entities/v2_support.ts
```diff
@@ -1,67 +1,188 @@
...
+    sharedId: string;
+    language: string;
+  },
+  method: 'Creating' | 'Updating'
+) => {
+  if (await newRelationshipsEnabled()) {
+    const transactionManager = DefaultTransactionManager();
     const denormalizationService = await DenormalizationService(transactionManager);
-    await denormalizationService.denormalizeAfterCreatingEntities([sharedId], language);
+    await denormalizationService[`denormalizeAfter${method}Entities`]([sharedId], language);
```
LaszloKecskes: I know it's a bit of an overkill for two options and a lot more types, but I would not rely on string operations for function call identification, even for just this much. Can we ad something like a mapping object for the actions?

app/api/entities/v2_support.ts
```diff
@@ -1,67 +1,188 @@
...
-    await transactionManager.executeOnCommitHandlers(undefined);
-  }
+const denormalizeAfterEntityUpdate = async (data: { sharedId: string; language: string }) =>
+  denormalizeAfterEntityAction(data, 'Updating');
+
+const calculateDiff = (currentDoc: EntitySchema, toSave: EntitySchema, name: string) => {
+  const newValues = new Set<string>();
+  const deletedValues = new Set<string>();
+
+  toSave.metadata?.[name]?.forEach(item => {
```
LaszloKecskes: I've started adding a couple of useful operations, like the objectIndex and the shallowObjectDiff in shared/data_utils, and this looks like a very useful operation to generalize. I.e. a diff on array, using a mapping for the diff index and result. Can we implement that (next to shallowObjectDiff)?

app/api/entities/v2_support.ts
```diff
@@ -1,67 +1,188 @@
...
+) => {
+  const transactionManager = DefaultTransactionManager();
+  const dataSource = DefaultRelationshipDataSource(transactionManager);
+  const service = await DeleteRelationshipService();
+  const toDelete: string[] = [];
+
+  await relationships.reduce(async (prev, relationship) => {
+    await prev;
+    const rels = await dataSource
+      .getByDefinition(relationship.from, relationship.type, relationship.to)
```
LaszloKecskes: This seems like an n+1 query, can we make getByDefinition into multi-operation?

app/api/templates.v2/database/TemplateMappers.ts
```diff
@@ -1,16 +1,23 @@
+/* eslint-disable no-redeclare */
```
LaszloKecskes, line 1: According to this link (https://typescript-eslint.io/rules/no-redeclare/), we could even turn this off entirely. Should we bring this for the next tech call?

app/api/templates.v2/routes/validators/createNewRelationshipProperty.ts
```diff
@@ -14,6 +14,7 @@ const createNewRelationshipPropertySchema: ValidatorSchema<
   optionalProperties: {
     _id: { type: 'string' },
     denormalizedProperty: { type: 'string' },
+    targetTemplates: { elements: { type: 'string' } },
```
LaszloKecskes: Boolean (false) also missing here.

app/api/templates/specs/v2_newRelationshipProperties.spec.ts
```diff
@@ -362,7 +366,7 @@ describe('template.save()', () => {
         ]);
       });
 
-      it('on query change, uwazi should save the query properly, and mark the metadata obsolete', async () => {
+      it('on query change, uwazi should save the query properly, mark the metadata obsolete and recalculate the target templates', async () => {
```
LaszloKecskes, line 371: Probably we should also test the "false" result (an ineligible one hop, and one multihop, maybe).

app/api/relationships.v2/database/MongoRelationshipsDataSource.ts
```diff
@@ -43,6 +44,17 @@ export class MongoRelationshipsDataSource
     );
   }
 
+  getByDefinition(sourceEntity: string, type: string, targetEntity: string) {
```
LaszloKecskes: Missing this on the contract interface. 

Also, it's most likely my bad, but we should make the return type of the defaults explicitly the interface, instead of letting typescript to infer, so we avoid this in the future. Can you do it for this one?

app/api/relationships.v2/database/RelationshipMappers.ts
```diff
@@ -114,6 +114,16 @@ export const RelationshipMappers = {
     const { traversal, ...entityData } = entityTraversal;
     return EntityMappers.toModel(entityData);
   },
+
+  toRelationshipIds(entityTraversal: EntityTraversal): string[] {
```
LaszloKecskes: I could not find an usage for this by searching for the string toRelationshipIds. Is this used anywhere?

app/api/relationships.v2/database/specs/MongoRelationshipsDataSource.spec.ts
```diff
@@ -259,3 +260,25 @@ describe('getAll()', () => {
...
+    const rels = await ds
+      .getByDefinition('entity1', factory.id('nullType').toString(), 'hub1')
+      .all();
+    expect(rels).toEqual([expect.objectContaining({ _id: factory.id('rel1').toString() })]);
+    rels.forEach(rel => {
+      expect(rel).toBeInstanceOf(Relationship);
+    });
+  });
+
+  it('should only find relationships that are not text references', async () => {
```
LaszloKecskes, line 275: Because we need this only for the 'writable property' feature, for now? So text references will not show up for this by definition? Make sense, since then you could accidentally delete a reference without any indication of doing so. Nice one!

app/api/relationships.v2/services/DenormalizationService.ts
```diff
@@ -46,13 +47,23 @@ export class DenormalizationService {
...
-    language: string
+    language: string,
+    relatedEntities: Entity[] = []
   ) {
     const properties = await this.templatesDS.getAllRelationshipProperties().all();
     const entities: { sharedId: string; property: string }[] = [];
     await Promise.all(
-      properties.map(async property =>
-        Promise.all(
+      properties.map(async property => {
```
LaszloKecskes, line 56: Let's have a call about this one. I think I understand, but I'd like to double-check with you. :)

app/api/relationships.v2/model/MatchQueryNode.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
```
LaszloKecskes, line 1: Eslint.

app/api/relationships.v2/model/MatchQueryNode.ts
```diff
@@ -207,6 +216,76 @@ export class MatchQueryNode extends QueryNode {
     return this.traversals.map((t, index) => t.getTemplatesInLeaves([...path, index])).flat();
   }
 
+  // eslint-disable-next-line max-statements
```
LaszloKecskes: Eslint.

app/api/relationships.v2/model/MatchQueryNode.ts
```diff
@@ -207,6 +216,76 @@ export class MatchQueryNode extends QueryNode {
...
+  determinesRelationships(): false | string[] {
+    const hasDepth2 = this.getDepth() === 2;
+    const hasSigleTypePerBranch = this.traversals.every(
```
LaszloKecskes: Typo. "hasSigle..."

app/api/relationships.v2/model/MatchQueryNode.ts
```diff
@@ -207,6 +216,76 @@ export class MatchQueryNode extends QueryNode {
...
+      record.templates.forEach(template => {
+        if (!templatesOccurences[template]) {
+          templatesOccurences[template] = 0;
+        }
+
+        templatesOccurences[template] += 1;
+      });
+    });
+
+    const hasOneLeave = templatesInLeaves.length === 1;
```
LaszloKecskes: Typo, singular is "leaf".

app/api/entities/v2_support.ts
LaszloKecskes: Also, I'm willing to do it, if you think it makes sense.

app/api/entities/v2_support.ts
```diff
@@ -1,67 +1,188 @@
...
+    relationships.push(query.determineRelationship(targetEntity));
+  });
+  return relationships;
 };
 
 const ignoreNewRelationshipsMetadata = async (
   currentDoc: EntitySchema,
   toSave: EntitySchema,
   template: TemplateSchema
+): Promise<DefinitionsToUpdate> => {
```
LaszloKecskes, line 98: I could not find a place where the backend checks the targetTemplates value of the property. Is there any?
If not, I think it would make sense to do so, because at this point only the empty result of the calculateDiff stops the process for non-relationship-determining properties, but only if the frontend or any request to the API does not change the metadata for those accidentaly. Since this is a template/property thing, which is not in v2 yet, it could as simple as a check here and returning empty arrays for the updates, and one test for it.

LaszloKecskes (changes requested):
Some comments, mostly in the code that acts between v1 and v2. However, the main algorithm for determining relationships is very nicely done and tested thoroughly.

app/shared/types/api.v2/templates.createTemplateRequest.ts
fnocetti: I think I found and fixed this, but likely got lost in a git stash. I will take a look, thanks!

app/api/entities/specs/v2_newRelationshipMetadata.spec.ts
fnocetti: Makes sense!

app/api/entities/v2_support.ts
fnocetti: It looks weird, but template strings are strongly typed when used to access object members.
That said, using a mapping object will provide any stronger type-safety, since you are still relying on object keys which are strings, to decide; but it will likely make the code bulkier. I can give it a try anyway.

app/api/entities/v2_support.ts
fnocetti: I agree. If you **would like to do it yourself**, feel free to do so. If not, I can take care myself.

app/api/entities/v2_support.ts
fnocetti: I don't fully understand the concern. But let me elaborate.
The `targetTemplates` property is only a calculated property, from the query, useful for the clients. The backend doesn't need the value because it doesn't even need to know what templates it would use, because it would just try to determine which relationship it needs to create/delete for the updated value. If the entity provided to the field is not of one of the `targetTemplates` it will throw an exception. In that line, I think what is missing is a validation on the rest api side, that could check that the provided entities are of one of those templates to prevent the exception on execution. Is that what you are refering to?
If that's the case, I agree and I can work on that.

app/api/entities/v2_support.ts
fnocetti: We can definitively make it a multi-operation, nice catch.

app/api/files/specs/v2_support.spec.ts
fnocetti: Yes, I don't fully remember at this point. But I had an error in the tests. I can check and circle back.

app/api/relationships.v2/database/MongoRelationshipsDataSource.ts
fnocetti: I don't think it's missing. This is a conceptual misunderstanding of why we have interfaces in the architecture. Let's talk about it in a meeting.

app/api/relationships.v2/database/RelationshipMappers.ts
fnocetti: Good catch! It is not, it was originally used by the `getByDefinition` method but I ended up not using it. I will anyways double check the flow.

app/api/relationships.v2/database/specs/MongoRelationshipsDataSource.spec.ts
fnocetti: That's exactly why. I anyways think this needs conceptual work in the future, but it's the safest path. If you cannot create text references through this field, you should not be able to delete them. They will anyway show as results of the query and be denormalized.
This could also be better treated in a validation and the UI, but I would wait for that until we start working on the frontend and know what we want to achieve.

app/api/relationships.v2/model/MatchQueryNode.ts
fnocetti: Let's agree on this one over a call. At this point I'm not sure whats our convention on this.

app/api/relationships.v2/model/MatchQueryNode.ts
fnocetti: Will refactor this method.

app/api/relationships.v2/model/MatchQueryNode.ts
fnocetti: Nice catch.

app/api/relationships.v2/model/MatchQueryNode.ts
fnocetti: That's an awful one. I'm tempted to delete the comment :joy: thanks!

app/api/templates.v2/database/TemplateMappers.ts
fnocetti: Let's definitively bring it to the tech call. Also, pretty sure the typescript overrides for this rule must handle this, cause I don't thing every redeclare is good. Only those for method overloads. Good catch, I forgot to dig deeper into this issue.

app/api/templates.v2/routes/validators/createNewRelationshipProperty.ts
fnocetti: I think this is likely the reason for the bug you found. I will take a look at this. For the record, in the ideal world we don't even accept target templates here, because it's a calculated property. But for now we have no option given the v1-v2 integration.

app/api/templates/specs/v2_newRelationshipProperties.spec.ts
fnocetti: We can do so. I didn't because the calculation of the `targetTemplates` value is tested in other places, and was only aiming to check that the calculation is performed.

fnocetti:
> Some comments, mostly in the code that acts between v1 and v2. However, the main algorithm for determining relationships is very nicely done and tested thoroughly.

@LaszloKecskes  Nice review, thanks! I'll address the comments. Also, I'll set up a call to chat on the topics that arised!

app/api/templates/specs/v2_newRelationshipProperties.spec.ts
LaszloKecskes: Good point, let's skip.

app/api/relationships.v2/model/MatchQueryNode.ts
fnocetti: Agreed to remove the disable since it's the current team agreement.

app/api/entities/v2_support.ts
fnocetti: Done.

app/shared/types/api.v2/templates.createTemplateRequest.ts
fnocetti: Fixed.

app/api/entities/v2_support.ts
fnocetti: As discussed, now `targetTemplates` is used by the backend to perform validations on entity update/creation.

app/api/files/specs/v2_support.spec.ts
fnocetti: Glad you noticed this. While this changes are needed, this is not the place and I was testing the wrong thing. Long story short: after deleting the text reference, entity2 and 3 are re-denormalized so they must end up with an empty obsoleteMetadata field and an empty relProp field. Why this is needed in this PR? Really likely because now the ignoreMetadata function in the v1-v2 integration sets the value to [] if it's unset, plus the entity save process is now actually triggering a denormalization check. For some reason at some point I fixed the expect by touching the fixtures instead of the expect 😬 Good catch!

app/api/entities/v2_support.ts
LaszloKecskes: Now that we have the abstracted function, should we also use it here? :)

LaszloKecskes: ...sorry, I got tricked by github, I was looking at the old version. All good. :sweat_smile:

LaszloKecskes (approved):
Everything according to discussions. Functional testing shows that the re-saving bug is fixed.

PR #6314, opened 2023-12-12
Update to ESM test PR

### LINKED ISSUE #6313: Test - Moving from CJS to ESM
This issue is meant to gather information about the topic of moving UWAZI to ESM only.
Related to issues:
https://github.com/huridocs/uwazi/issues/6218
https://github.com/huridocs/uwazi/issues/6186
https://github.com/huridocs/uwazi/issues/5133

Some identified tasks: 
- All of webpack config has to be purged of `requires()` before you can really get into changing the code of the app, to discard certain errors being the result of webpack not working.
- Some imports will need to have the file extension in order to work, and some modules will need to be pointed to the index (for example `import { DB } from './api/odm'` to `import { DB } from './api/odm/index';` in `app/server.js`). This could be because we did not finish configuring webpack for this exercise.
- As you rewrite files to get rid of `require()`, you can find yourself in a situation where that rewritten file imports another file that has a `require()` still. So it's slow work.
- Most dynamic `require()`s will have to be switched to dynamic `import()`s. For example in `server.js`:
```
if (NODE_ENV === 'production') {
  try {
    require('./app/server.js');
  } catch (err) {
    require('./prod/app/server.js');
  }
} else {
  require('@babel/register')({ extensions: ['.js', '.jsx', '.ts', '.tsx'] });
  require('./app/server.js');
}
```
will have to be changed to 
```
(async () => {
 if (NODE_ENV === 'production') {
   try {
     await import('./app/server.js');
   } catch (err) {
     await import('./prod/app/server.js');
   }
 } else {
   const { default: babelRegister } = await import('@babel/register');
   babelRegister({ extensions: ['.js', '.jsx', '.ts', '.tsx'] });
   await import('./app/server.js');
 }
})();
```
- As cjs files are moved to esm, many errors of this style will pop up:
```
SyntaxError: Named export 'loader' not found. The requested module 'mini-css-extract-plugin' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'mini-css-extract-plugin';
const { loader: _loader } = pkg;
```
- In summary, there’s a lot of slow manual labor that needs to be done. Until it is all completed and the app compiles it is not possible to know if we might run into a situation where a 3rd party library will have a conflict, blocking us until we update it or replace it.

Additional context:
- Important documentation about the use of `import()` to replace dynamic `require()` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#module_namespace_object

Zasa-san:
fixes #6313

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


Zasa-san:
Stale. Closing this PR.

PR #6320, opened 2023-12-14
fix bash.js and usage

### LINKED ISSUE #6246: Development scripts not handling command line parameters
Currently our development scripts do not handle command line parameters, when invoked through the corresponding package json command. Example:
 - blank_state.sh us supposed to take a parameter for the collection name: `blank_state.sh collection_name`
 - but, when invoked through `yarn blank-state collection_name`, the parameter is not passed and the script will use the default uwazi_development
 
The reason is that the recently introduced runner script `bash.js` truncates the argument list. The task is:
- fix `bash.js`
- double check that all commands using it are working fully as intended
- double check whether there are other scripts that might have similar problems (`run.js` or others?)

Note for prioritization: I would consider this high priority, since the fix should be relatively simple, and the current setup makes our development scripts unusable in some scenarios (for example, when setting up multiple tenants for sync related issues).

LaszloKecskes:
fixes #6246 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti (approved):
LGTM!

PR #6384, opened 2024-01-25
Hotfix: update settings links

### LINKED ISSUE #6382: Menu configuration is failing validation on previously existing GROUPS and its internal links
**Describe the bug**
Existing Menus (maybe new ones as well?) are failing validation on two fronts:
- It says its expecting Type for the sublinks, which can only be type: "link"
- It's complaining about "extra properties", because for sublinks it's sending the LocalID. Usually that LocalID is purged in client before making the backend call

Maybe the grouped menu was not tested? Maybe the approach changed and we are not using the LocalID, but the existing menus already have that and, therefore, we need to still sanitize the data from the backend (or do a migration, which I would hope is not needed)

LaszloKecskes:
fixes #6382 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/migrations/migrations/153-update_settings_link_structure/index.ts
+++ b/app/api/migrations/migrations/153-update_settings_link_structure/index.ts
@@ -1,0 +1,93 @@
+import { Db } from 'mongodb';
+import {
+  CorrectGroup,
+  CorrectLink,
+  CorrectSimpleLink,
+  FaultyLinkOrGroup,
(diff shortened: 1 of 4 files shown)


daneryl (approved):
LGTM

PR #6386, opened 2024-01-26
5627 - handle public form errors

### LINKED ISSUE #5627: Handle PublicForm errors
Right now when a user misconfigured the template in a public form element 
```
<PublicForm template="wrongTemplateId" file />
```
Uwazi will throw an unhandled error on the page. If this page is set as the landing page, it might block users from using the instance.

Errors on pages should not break the application and should display an error in a way that users can continue working.

**Additional context**
We should check if other components that users can implement in custom pages break the app in the same way.
Technically, pages have an error boundary component, but in this case we are not seeing our custom error component.

https://github.com/huridocs/uwazi/blob/e547cfe4583fdc318ebe3ace3b7ef79a507efd61/app/react/Pages/PageView.js#L55-L67

mfacar:
fixes #5627

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


Zasa-san (approved):
LGTM!

PR #6452, opened 2024-02-08
Make getByQuery transactional.

### LINKED ISSUE #6378: Transitive denormalization not working on relationship v2 creation
**Describe the bug**
When having a 2 hop query in a template, and creating a relationship that matches the second hop, the algorithm won't re-calculate the relationship fields for the entities that have that query.

**To Reproduce**
1. Have templates A, B and C
2. In template A, create a relationships v2 field with the query: A-[relType1]->B-[relType1]->C
3. Create entities A1 and B1 and create a relationship between them as follows: A1-[relType1]->B1
4. Create an entity C1 and relate it to B1 as follows: B1-[relType1]->C1
5. Check the value of the relationship field in A1.

**Expected behavior**
It should include C1 in its value since it matches the query.

**Actual behavior**
C1 is not in the value of the field

fnocetti:
fixes #6378 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/common.v2/database/MongoDataSource.ts
```diff
@@ -16,17 +16,32 @@ export abstract class MongoDataSource<CollectionSchema extends Document = any> {
...
+  }
+
+  protected async dropCollection() {
+    if (await this.collectionExists()) {
+      await this.db.dropCollection(this.collectionName, { session: this.getSession() });
+    }
+  }
+
+  protected async createCollection() {
+    await this.db.createCollection(this.collectionName, { session: this.getSession() });
```
LaszloKecskes, line 40: Asking, since there's a check in dropCollection: similarly, will this one handle a case where the collection does already exists? Ideally, we would like it to do nothing in that case.

LaszloKecskes (approved):
There's one minor question. Please take a look at it, other than that, everything as discussed.

app/api/common.v2/database/MongoDataSource.ts
fnocetti: This piece of code was extracted exactly as it was in it's original place. If I'm correct it will fail when trying to create a an already existing collection. I don't think it's a bad thing if it fails because making it 'non-failing' should be a responsibility of the consuming code. That brings me to another line of thinking which is: we should not add the check here, but actually move the one in the drop back to the consumer.

PR #6475, opened 2024-02-13
New Settings/Thesauri UI

### LINKED ISSUE #6433: New Thesauri UI
Implement the new Thesauri UI as in the design

- Can top-level thesaurus be dragged into a group? What happens to denormalization processes?

grafitto:
fixes #6433 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/V2/Components/Forms/InputField.tsx
```diff
@@ -60,9 +60,11 @@ const InputField = React.forwardRef(
 
     return (
       <div className={className}>
-        <Label htmlFor={id} hideLabel={hideLabel} hasErrors={Boolean(hasErrors || errorMessage)}>
-          {label}
-        </Label>
+        {label && (
+          <Label htmlFor={id} hideLabel={hideLabel} hasErrors={Boolean(hasErrors || errorMessage)}>
```
Zasa-san: Why is there a conditional to show the label if the component already has a prop to hide it?

app/react/V2/Components/UI/Button.tsx
```diff
@@ -2,7 +2,7 @@ import React, { MouseEventHandler } from 'react';
 
 interface ButtonProps {
   children: string | React.ReactNode;
-  styling?: 'solid' | 'outline' | 'light';
+  styling?: 'solid' | 'outline' | 'light' | 'action';
```
Zasa-san, line 5: "action" is very similar to "light", exepct it has no hover. Maybe we don't need a fourth type of button, just modify light to be more like action?

app/react/V2/Components/UI/Table/TableElements.tsx
```diff
@@ -24,6 +24,7 @@ interface TableProps<T> {
   onSelection?: Dispatch<SetStateAction<Row<T>[]>>;
   subRowsKey?: string;
   draggableRows?: boolean;
+  blankLabel?: string | React.ReactNode;
```
Zasa-san: Is this used?

app/react/V2/Routes/Settings/Thesauri/components/TableComponents.tsx
```diff
@@ -0,0 +1,108 @@
...
+    cell: ThesaurusValueLabel,
+    meta: { headerClassName: 'w-11/12' },
+  }) as ColumnDef<TableThesaurusValue, 'label'>,
+  columnHelper.accessor('_id', {
+    header: ActionHeader,
+    cell: EditButton,
+    enableSorting: false,
+    meta: {
+      action: actions.edit,
+      headerClassName: 'text-center w-0',
```
Zasa-san: Here you should add the sr-only classname to hide the 'Action' label for the header of the action column

Zasa-san (changes requested):
Some general comments:
- When adding items or groups the sidepanel should be titled “Add items” and “Add group” respectively, and when editing change to “Edit item” and “Edit group”. Also, when editing the blue message on top doesn't really apply. In the case of adding/editing items the blue message appears only when editing, when in fact it should only appear when adding.

- The action column on the table should have the “sr-only” classname on the heading.

- Sorting does not sort the group internally. Is that intended? In development it does.

- The buttons at the bottom seem to scroll upwards on longer tables. This behavior affects all settings screens and is not present in development. You can check it by creating a longer thesaurus, or in the system translations list.

- Selecting and removing a group, and then selecting and removing an item, or selecting both a group and an item, deletes everything on the list (this is all before hitting save). Also, selecting and deleting a single item removes all the items on the thesauri, except for groups. This is observed on thesauri created in development, not on new ones created in the branch.

app/react/V2/Components/UI/Button.tsx
grafitto: I did not want to modify light because it is being used by other parts of the UI

app/react/V2/Components/Forms/InputField.tsx
grafitto: Fixed

app/react/V2/Components/UI/Table/TableElements.tsx
grafitto: Removed

app/react/V2/Routes/Settings/Thesauri/components/TableComponents.tsx
grafitto: Added

salvalacruz:
NEEDS TO CHECK

When trying to change the order of items or moving an item into or out of a group, the behaviour in unsteady.

Also, moving the items sometimes you get this error:

image

PR #6485, opened 2024-02-15
6445 remove joi

### LINKED ISSUE #6455: Remove Joi from the application
Currently, Joi is in the application to support legacy usage in a few of our older modules, in new developments we are now using Ajv. Since now there would be a need to upgrade Joi through versions with breaking changes, we have opted for removing it altogether.

LaszloKecskes:
fixes #6455  

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/auth2fa/routes.ts
```diff
@@ -1,14 +1,16 @@
...
 import { validation } from 'api/utils';
+import { ObjectIdAsString } from 'api/utils/ajvSchemas';
 
 export default (app: Application) => {
   app.post(
     '/api/auth2fa-secret',
     needsAuthorization(['admin', 'editor', 'collaborator']),
-    validation.validateRequest(Joi.object().keys({})),
+    validation.validateRequest({
+      type: 'object',
```
fnocetti, line 12: Isn't this one missing the nested `body` object? Shouldn't that one be emtpy?

app/api/utils/specs/validateRequest.spec.js
```diff
@@ -16,92 +14,61 @@ describe('validateRequest', () => {
     };
   });
 
-  describe('using JOI schema', () => {
```
fnocetti, line 19: To see if I understood the changes: here you only removed this describe, right?

fnocetti (approved):
In general it looks good! I'm only wondering if we are safe with the route validation testing that we have or we should invest some efforts there before merging this...

app/api/utils/specs/validateRequest.spec.js
LaszloKecskes: Yes, and also moved the contents of the other describe up one level, since now we don't have multiple options for this.

app/api/auth2fa/routes.ts
LaszloKecskes: If it was, the main Joi object would have been marked required. With that said, in other cases, I've appended the required tag, even if it was not present in the original, since if in an endpoint we require something in a body, I think it's safe to say that the body itself is required as well. Here, I don't think requiring an empty object makes any difference, so I've kept to the original.

PR #6522, opened 2024-02-21
Emtpy action header 

mfacar:
Removes Action header for tables in V2 components

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


Zasa-san:
@mfacar @juanmnl 
Now that the action column has no text, the gray backdrop in the headers for the columns does not show for the actions column: (it's hard to see)
imagen
imagen

Can we do a fix for this so that the gray background shows even if the header is empty?

Zasa-san:
@mfacar @juanmnl  Another issue, technically empty headers on tables are an a11y violation:
imagen

What we should do, in think, is add a translatable title to the header for the column, and hide it with tailwind's `sr-ony` class.

PR #6528, opened 2024-02-22
6509 - Custom uploads new UI

### LINKED ISSUE #6509: Custom Uploads new UI
[Figma](https://www.figma.com/file/HZX8qH0Aw8WiSch3Axczrn/new-comps?type=design&node-id=1979-19837&mode=design&t=cTJyPbiGkX2rtXYu-0)

- Do not implement the status column.

Screenshot 2024-02-20 154042
Screenshot 2024-02-20 154140
Screenshot 2024-02-20 154200

Zasa-san:
fixes #6509

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/V2/Components/Forms/FileDropzone.tsx
```diff
@@ -0,0 +1,94 @@
...
+      result += file.size;
+    });
+    setTotalSize(result);
+  }, [files]);
+
+  useEffect(() => {
+    if (onChange) {
+      onChange(files);
+    }
+  }, [files]);
```
konzz, line 30: Could these 2 useEffect be the same?

app/react/V2/Routes/Settings/CustomUploads/CustomUploads.tsx
```diff
@@ -0,0 +1,185 @@
...
+      setConfirmNavigationModal(true);
+    }
+
+    if (blocker.state === 'proceeding') {
+      uploadService.abort();
+    }
+  }, [blocker, setConfirmNavigationModal]);
+
+  const notify = (responses: (FileType | FetchResponseError)[], message: React.ReactNode) => {
+    const hasErrors = responses.find(response => response instanceof FetchResponseError);
```
konzz: I don't think we should delegate the show or not show an error based on a type, that can change in the future.
This needs to be a try-catch or similar. If you go to the route and throw an error on the request this does not show anything.

app/react/V2/Routes/Settings/CustomUploads/components/UploadProgress.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const UploadProgress = ({ queueLength }: UploadProgressProps) => {
+  const { filename, progress } = useRecoilValue(uploadProgressAtom);
+
+  return filename ? (
+    <div className="flex grow">
+      <Translate>Uploading</Translate>...&nbsp;
+      <span className="font-semibold">{filename}</span>&nbsp;
+      {progress}%&nbsp;
+      <span className="flex flex-nowrap before:content-['('] after:content-[')']">
```
konzz, line 18: Why not just add the "(" to the HTML?

app/react/V2/api/files/UploadService.ts
```diff
@@ -0,0 +1,101 @@
...
+    this.aborted = true;
+    this.requests.forEach(request => request.abort());
+  }
+
+  public isUploading() {
+    return Boolean(this.requests.length);
+  }
+}
+
+export { UploadService };
```
konzz, line 101: Very nice service 👍

konzz (commented):
I've noticed there is no popup when hovering over an image like in the designs, did we discard that?

app/react/V2/Routes/Settings/CustomUploads/components/UploadProgress.tsx
Zasa-san: They don’t convey meaning, it’s just for looks. In the past for this type of case we have done this to allow users who customize the UI to decide if they want the parenthesis or not, since they are there for looks only. They would not have to alter the HTML to remove the parentheses, just touch the custom CSS. I don’t have a preference one way or another.

app/react/V2/Routes/Settings/CustomUploads/CustomUploads.tsx
Zasa-san: I’ll look at an alternative, this is done this way in a number of places in the new UI, I can change it here and then let’s review the other cases.

app/react/V2/Components/Forms/FileDropzone.tsx
Zasa-san: Makes sence since they have the same dep array :ok_hand:

app/react/V2/Routes/Settings/CustomUploads/CustomUploads.tsx
Zasa-san: I can check that there's at least a valid file in the response to determine if there's an error. But I believe we should look into how we want to detect errors in responses generally, since other places in the new routes are doing this.

Issue #5068, opened 2022-09-01
Translation modal is covered by other modals


Zasa-san:
**Describe the bug**
While live translating if you have a modal open, the translation modal is covered and cannot be used. This is not for every modal.

**To Reproduce**
1. Activate live translate
2. Go to an entity
3. Click on add supporting file button
4. Try to translate any term in the supporting file modal

**Expected behavior**
Either modals should not allow live translate, or the live translate modal should always sit on top.

**Screenshots**
Deleting users:
![image](https://user-images.githubusercontent.com/71732018/187968685-e09afceb-24ba-42f1-80d5-d1a727e4928c.png)

Add supporting file modal:
![image](https://user-images.githubusercontent.com/71732018/187969146-be89c2de-8972-46f3-89ff-b0845cd27776.png)

**Device (please select all that apply)**
 - Desktop

**Browser**
Firefox

**Additional context**
It's likely this problem is worse in mobile, since actions buttons are very difficult to use while live translate is on, and closing the modal can be tricky.

aphilop:
This seems to be resolved. 
The translation modal comes on top of the opened modal and I can live translate elements of the opened modal

<img width="980" height="1001" alt="Image" src="https://github.com/user-attachments/assets/d4f088be-cf40-42c2-a513-d73cafaee449" />

<img width="1473" height="1040" alt="Image" src="https://github.com/user-attachments/assets/ec76e00c-8d86-451c-99cf-719e0e2f7f3a" />

PR #6569, opened 2024-03-06
V2 translate

### LINKED ISSUE #6691: V2 Translate component
Front end team has been working on a V2 Translate component. The aim is to remove redux dependencies from translations, so that new routes have as few dependencies from V1 as possible.

To cover in the PR we have been working on:
- [x] Migrate the `Translate` component to use atoms and not redux
- [x] Migrate the `t` function in the same way
- [ ] ~~If possible, use V2 translations API data structure~~
- [ ] ~~If possible, stop depending on redux for the initial state of the app~~
- [x] Update translation sockets actions to update atom store and not redux
- [x] Tests (test that atoms update via sockets, test that when you submit translations the socket even is fired?, we are not updating the translationsAtom after submitting, so consistency across the UI depends on sockets)
- [x] Refactor (for example t has circular deps)
- [x] API changes need to be reviewed
- [x] Modal styles
- [x] Cleanup
- [x] After merging, create issue to migrate form so that we can remove the old translations endpoitns
- [x] Modal stacking. Translate modal can be triggered by other modals. It should always sit on top of all other app elements.
- [x] unit test for `app/react/V2/api/translations/index.ts` postV2

Zasa-san:
fixes #6691 #7350


Zasa-san:
We should also check this -> https://github.com/huridocs/uwazi/issues/5068

app/api/i18n/routes.ts
```diff
@@ -158,6 +169,45 @@ export default (app: Application) => {
...
+          },
+        },
+      },
+      required: ['body'],
+    }),
+    async (req, res) => {
+      await upsertTranslationEntries(req.body);
+      req.sockets.emitToCurrentTenant('translationKeysChange', req.body);
+      res.status(200);
+      res.json({ success: true });
```
daneryl: I think this is missing important functionality from the old translations.save(), maybe some of those things are no longer needed but at least the propagateTranslation is, which denormalizes the translations in the entities metadata, this should probably be encapsulated in an action.
Also the new endpoints belong to i18n.v2 module.
This needs more work from the server team.

app/api/i18n/routes.ts
```diff
@@ -158,6 +169,45 @@ export default (app: Application) => {
     }
   );
 
+  app.post(
```
daneryl: Are this changes still needed ? looks like i18n.v2 has the same endpoints defined ?

Zasa-san: :man_facepalming: An oversight. Fixed!

PR #6572, opened 2024-03-07
Multiselect allows option updating, returns values and respects selection order

### LINKED ISSUE #6354: Multiselect V2 should have an standard input API
- The new V2 Multiselect does not accept property value
- It expects the options to have a property "selected" to calculate the value
- onChange does not return the value, but instead, it returns all the options (selected or not)
- The options are set in an internal state preventing updates of the possible options
- The order in which the options are selected does not matter since the selection is kept in the options and those don't change

It SHOULD

- Return an array of selected values (i.e. `['red', 'green', 'blue']`
- Allow for the options to be updated after the first initialization
- Use a property `value` to calculate the initial selected options

konzz:
fixes #6354 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


Zasa-san (approved):
LGTM!

PR #6594, opened 2024-03-15
No more migrations for translations

konzz:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


scripts/updateTranslationsCSV.mjs
```diff
@@ -0,0 +1,262 @@
...
+    .concat(...files)
+    .filter(
+      file =>
+        !file.match('.spec') &&
+        !file.match('.stories') &&
+        !file.match('.cy') &&
+        (file.endsWith('.js') ||
+          file.endsWith('.ts') ||
+          file.endsWith('.tsx') ||
+          file.endsWith('.jsx'))
```
mfacar: This condition could be simplified with something like: ```!file.match(/(\.spec|stories|\.cy)/) && file.match(/(\.js|\.ts|\.tsx|\.jsx)$/)```, what do you think?

scripts/updateTranslationsCSV.mjs
```diff
@@ -0,0 +1,262 @@
...
+    );
+}
+
+const getKeysFromRepository = async (locale = 'en') =>
```
mfacar, line 31: In order to avoid duplication, what about extracting this and getAvailableLanguages into a new file that exports them as `module.exports = { getKeysFromRepository, getAvailableLanguages };`

mfacar (changes requested):
Great job! It's a super improvement. While testing I found a problem with the command to update the database, it is deleting entries for contexts different from 'System', like template properties and menu items.

Also, please don't forget the observation from @Zasa-san, about a check between the differences of the CSV files across languages.

konzz:
Closing PR to create a new one pointing to production

PR #6595, opened 2024-03-15
Scroll export

### LINKED ISSUE #2924: Check and improve CSV Export after Library Pagination

fnocetti:
fixes #2924 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/csv/csvExporter.ts
+++ b/app/api/csv/csvExporter.ts
@@ -1,3 +1,4 @@
+/* eslint-disable no-await-in-loop */
 import { Writable } from 'stream';
 import { EventEmitter } from 'events';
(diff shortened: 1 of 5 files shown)


fnocetti:
This PR includes a proof of concept of using the Scroll API from elastic to export more than 10k results.
It implements a scroll method in the search module that returns an iterator. That iterator is then used in the csv exporter module.
To continue this work, testing needs to be taken care of.
If there was a need to consume all of the results at once, an `all` method could be included into the iterator that leverages the `next` and `hasNext` methods of itself to list every result.

PR #6664, opened 2024-04-12
fixes flickering caused by date range picker

mfacar:
fixes flickering at the moment of focus into the date range picker


mfacar:
Closing in favor of #6667

PR #6683, opened 2024-04-18
Empty metadata as empty array

### LINKED ISSUE #6568: Empty metadata values should always be set to empty arrays
We have played with this idea for some time and the team wants to start considering it.
We want to implement that in an entity, every empty value is set to an empty array. Up to now, the "empty value" concept can be represented in one of 3 ways: unset (`undefined`), `null`, or empty array (`[]`).

LaszloKecskes:
fixes #6568

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/migrations/migrations/165-default_empty_metadata/specs/fixtures.ts
```diff
@@ -0,0 +1,343 @@
...
+  },
+];
+
+const correctFixtures: Fixture = {
+  settings,
+  templates,
+  entities: correctEntities,
+};
+
+const fixtures: Fixture = {
```
fnocetti: Can we add a case in which the entity has some metadata properties with correct values and some others to be fixed?

app/api/suggestions/specs/suggestions.spec.ts
```diff
@@ -828,14 +828,14 @@ describe('suggestions', () => {
           .find({ sharedId: 'shared1' })
           .toArray();
         const ages1 = entities1?.map(entity => entity.metadata.age[0].value);
-        expect(ages1).toEqual(['17', '17']);
+        expect(ages1).toEqual([17, 17]);
```
fnocetti: Why was this needed?

app/api/entities/entities.js
```diff
@@ -534,8 +535,17 @@ export default {
...
+    const indexedTemplates = objectIndex(
+      await templates.get({ _id: { $in: templateIds } }),
+      t => t._id.toString(),
+      t => t
+    );
+
+    await docs.reduce(async (prev, _doc) => {
       await prev;
+      const template = indexedTemplates[_doc.template];
+      const doc = this.sanitize(_doc, template);
```
fnocetti, line 548: Just to understand: this was the only change needed because `sanitize` already set's empty as arrays, but we were missing it in `saveMultiple` and to fix the db?

fnocetti (commented):
The code looks good. I left a couple questions.
I will proceed to test it.

app/api/migrations/migrations/165-default_empty_metadata/specs/fixtures.ts
LaszloKecskes: Sure, will do so.

app/api/suggestions/specs/suggestions.spec.ts
LaszloKecskes: The suggestions were using saveMultiple, which did not have sanitation. The sanitation also turns string values numeric metadata into proper numbers. So technically, this was an undiscovered IX bug, the previous test was testing the wrong thing.

It's not relevant to this PR, but In the current relationship IX support PR we are going one step further and switching out the saveMultiple to the proper entity save.

app/api/entities/entities.js
LaszloKecskes: Yes. I made sure to go through every place that directly uses the entity model, or even just the mongodb `collection('entities)`, but all of those are either in tests, or updates to existing metadata in a way that does not mess up a correct database state (for example, the efficient re-denormalizations). This is the reason the PR took longer, but has relatively few changes.

app/api/suggestions/specs/suggestions.spec.ts
fnocetti: Got it!

app/api/entities/entities.js
fnocetti: Nice work!

fnocetti (approved):
Looks good and works!

Issue #6676, opened 2024-04-15
Thesauri settings: order items


salvalacruz:
When adding a new item into a thesaurus, it goes to the last position.
If you need to set this new item more to the top of the list, it is not posible to drag and drop it easily.

RafaPolit:
It appears that this is not exclusive to the last element, actually ANY attempt to sort a children sends the item to the bottom of the list.

konzz:
The issue with dragging children is reported already in a different issue. #6681

mfacar:
@salvalacruz, Are the difficulties while re-ordering with drag and drop perceived when there is a vertical scroll the the list? If that is the case, there is a [known bug](https://github.com/react-dnd/react-dnd/issues/3482) in React-DnD, it is missing to fire the auto-scrolling while dragging elements outside the visible section of the screen. There are a couple of alternatives proposed in the same link that we can try to improve the experience. In the meantime it is needed to scroll to the very top/bottom of the page to fire the scrolling, keeping the row dragged until the target space is visible, or moving by parts to temporal positions, not ideal but can help until a solution is implemented.

salvalacruz:
Yes @mfacar it's about moving an intem into a thesaurus to change its order in the list.
This temporary solution could work internally for us, but any user could be confused when trying to do that.
@konzz please, can you consider assing again Priority-High? In some contexts (longs lists of terms when order is not alphabetical but it's sensitive), setting up thesauri becames really hard.
Thank you all

PR #6700, opened 2024-04-24
6681 - DnD table children rows sorting issues

### LINKED ISSUE #6681: Issues with dnd tables
Tables with drag and drop functionality present two problems.

- On tables that scroll vertically, due to the number of entries they have, dragging and item makes the vertical scroll jump around. This could be because when you stop dragging an item the UI tries to show where the item would drop, moves the adjacent rows, and messes with the vertical scroll. You can see this happen in the "Roles" thesauri in the Demo instance.

- Nested items cannot be reordered by dragging, or the dragging behaves inconsistently. This is noticeable with the last item of a group, that cannot be moved up by dragging it.

(Observed in Firefox)

mfacar:
fixes #6681

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


Zasa-san (commented):
- When adding items to an existing group, you get the error 
Uncaught TypeError: row is undefined
    TableRow TableRow.tsx:28
    React 16
    withLazy componentWrappers.tsx:13
    promise callback*withLazy/</< componentWrappers.tsx:11
    React 15
    next index.esm.mjs:2531
    next index.esm.mjs:1294
    handleSubmit index.esm.mjs:2219
    React 23
    <anonymous> entry-client.tsx:36
    tsx main.js:7504
    __webpack_require__ main.js:8441
    __webpack_exports__ main.js:10263
    O main.js:8562
    webpackJsonpCallback main.js:10249
    <anonymous> vendor.js:9
2 TableRow.tsx:28:21

Since a similar error was fixed in this PR, maybe this could be fixed too? Otherwise a new issue could be created.

- Resorting items still causes the viewport scroll to reset, annoying but not
breaking usability. It can make it hard to work when dragging something from a long list to the middle of it, since you loose view of the change when the scroll resets. Is this reported in another issue?

Zasa-san (approved):

mfacar:
@Zasa-san thanks for your review, I added a control to fix the first part of your comment. For the second one, the problem is already reported in #6676, although it is reported for thesauri, it applies to all the catalogs with dnd tables. I'm closing this PR for now and leaving the other issues to be solved in other pull requests.

PR #6727, opened 2024-05-03
ix multiselect support

### LINKED ISSUE #6633: IX Frontend Select && Multiselect support
Add support for 2 new properties, select and multiselect

- Backend is already working in the support, coordinate with them
- Accepting a suggestions in the case of a select should work as previous properties ( replacing one value for another)
- Accepting a suggestion for a multiselect can be:
  - a. Adding a missing option that was not previously in the value and is coming in the suggestion
  - b. Removing an option that is present in the value but not in the suggestion
- Bulk actions should do every possible action for the document(s)
- There is no "Select text & fill step", and we should not try to render the rectangles when opening the PDF.
- There will be a context property coming with the sugestion.

Screenshot 2024-03-26 123941
Screenshot 2024-03-26 123951

konzz:
fixes #6633

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/services/informationextraction/InformationExtraction.ts
```diff
@@ -365,10 +368,30 @@ class InformationExtraction {
...
+    };
+
+    if (property.type === 'select' || property.type === 'multiselect') {
+      const [thesauri] = await dictionatiesModel.get({ _id: property.content });
+
+      params.options =
+        thesauri?.values?.map(value => ({ label: value.label, id: value.id as string })) || [];
+    }
+
+    console.log(params);
```
LaszloKecskes: extra console log.

app/api/services/informationextraction/InformationExtraction.ts
```diff
@@ -365,10 +368,30 @@ class InformationExtraction {
       return { status: 'error', message: 'No labeled data' };
     }
 
+    const [template] = await templatesModel.get({ _id: extractor.templates[0] });
```
LaszloKecskes: Use templates.getById. It's the same, but let's keep to the layering and not use the model directly, if we can avoid.

app/api/services/informationextraction/InformationExtraction.ts
```diff
@@ -365,10 +368,30 @@ class InformationExtraction {
...
+      return { status: 'error', message: 'Property not found' };
+    }
+
+    const params: TaskParameters = {
+      id: extractorId.toString(),
+      multi_value: property.type === 'multiselect',
+    };
+
+    if (property.type === 'select' || property.type === 'multiselect') {
+      const [thesauri] = await dictionatiesModel.get({ _id: property.content });
```
LaszloKecskes: Same here, thesauri.getById.

PR #6734, opened 2024-05-06
6715 - thesauri screen crash on dnd

### LINKED ISSUE #6715: Thesauri screen crash on dnd
**Describe the bug**
When dragging and dropping items in the thesauri screen you can cause the app to crash. This seems to involve dragging and dropping root items inside a group.

**To Reproduce**
1. Go to a thesaurus, for example "Region" thesaurus in demo
2. Create a group with at least one item (don't save yet)
3. Grab root items and start trying to drop them inside the group
4. See error eventually

**Expected behavior**
You should be able to drag items from root to group, group to group, and group to root.

**Screenshots**
image

Desktop

**Browser**
Firefox

mfacar:
fixes #6715

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/V2/Components/UI/Table/TableBody.tsx
+++ b/app/react/V2/Components/UI/Table/TableBody.tsx
@@ -50,7 +50,6 @@ const setRowId: <T>(
 const TableBodyComponent = <T,>({
   draggableRows,
   allowEditGroupsWithDnD,
-  // eslint-disable-next-line react/jsx-no-useless-fragment
   DndProvider,
   HTML5Backend,
   items,
--- a/app/react/V2/Components/UI/Table/TableBody.tsx
+++ b/app/react/V2/Components/UI/Table/TableBody.tsx
@@ -8,6 +8,7 @@ import { useDnDContext } from '../../Layouts/DragAndDrop';
 
 interface TableBodyProps extends PropsWithChildren {
   draggableRows: boolean;
+  editableGroups?: boolean;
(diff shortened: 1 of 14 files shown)


app/react/V2/Components/UI/Table/TableBody.tsx
```diff
@@ -48,6 +49,7 @@ const setRowId: <T>(
 // eslint-disable-next-line comma-spacing
 const TableBodyComponent = <T,>({
   draggableRows,
+  allowEditGroupsWithDnD,
   // eslint-disable-next-line react/jsx-no-useless-fragment
```
Zasa-san: I think this can be removed

Zasa-san (changes requested):
LGTM! This is a very good solution.

app/react/V2/Components/UI/Table/TableBody.tsx
mfacar: I checked it and the property is needed to initialize the DnDOperations settings that are needed [later](https://github.com/huridocs/uwazi/blob/6715-thesauri-screen-crash-on-dnd-rows/app/react/V2/Components/Layouts/DragAndDrop/useDnDContext.tsx#L27) in the draggable items. So I'm keeping it.

PR #6735, opened 2024-05-07
6713 = Activity log filter not working

### LINKED ISSUE #6713: Activity log: Filter not working
It seems the filter feature in the /settings/activitylog screen does not work:

- If you filter by user, it does work properly.
- If you filter by ids, methods, keywords, etc., it does NOT work, the result is always empty.
- Also, filtering by date (From..To) does NOT work too.

mfacar:
fixes #6713

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/V2/Components/Forms/DatePicker/DatePickerComponent.tsx
```diff
@@ -90,6 +91,7 @@ const DatePickerComponent = React.forwardRef(
       dateFormat = 'yyyy-mm-dd',
       hideLabel = true,
       className = '',
+      mainClassName = '',
```
Zasa-san: Here we could change className to inputClassName and mainClassName to either containerClassName or just className, I think.

app/react/V2/Components/Forms/DatePicker/DatePickerComponent.tsx
```diff
@@ -132,9 +134,16 @@ const DatePickerComponent = React.forwardRef(
       return () => (instance?.current?.hide instanceof Function ? instance?.current?.hide() : {});
     }, [id, locale, labelToday, labelClear, datePickerFormat, clearFieldAction]);
 
+    useEffect(() => {
+      if (instance.current !== null && ref.current !== null) {
+        //@ts-expect-error
```
Zasa-san: I think it's best to check if value exists, and then do a toString() if it's a number.

app/react/V2/Components/Forms/DatePicker/DateRangePickerComponent.tsx
```diff
@@ -80,9 +86,22 @@ const DateRangePickerComponent = React.forwardRef(
       return () => (instance?.current?.hide instanceof Function ? instance?.current?.hide() : {});
     }, [locale, labelToday, labelClear, dateFormat]);
 
+    useEffect(() => {
+      if (instance.current !== null && fromRef.current !== null && toRef.current !== null) {
+        //@ts-expect-error
```
Zasa-san: Same as the comment in the other component.

app/react/V2/Components/Forms/DatePicker/DateRangePickerComponent.tsx
```diff
@@ -111,12 +130,12 @@ const DateRangePickerComponent = React.forwardRef(
                 </svg>
               </div>
               <InputField
+                id="from"
                 // @ts-ignore
```
Zasa-san, line 132: I've seen this ts-ignore in our InputField component and the input field tag in `V2/Components/Forms/DatePicker/DatePickerComponent.tsx`. 
Would this not by the attribute type="date" ? Or is this something specific to our implementation.

app/react/V2/Components/Forms/MultiSelect.tsx
```diff
@@ -59,6 +61,12 @@ const MultiSelect = ({
     onChange(newValue);
   };
 
+  useEffect(() => {
+    if (!isFirstRender) {
+      setCurrentValue(value);
```
Zasa-san: I'm not convinced that we should be manually calculating if it's a first render. Why do we want to set the value only if it's not the first render? The value we are setting comes from the props, and it's the value assigned to currentValue when the component mounts. So when it mounts we give it that value, and if it re renders we give it that same value again? Can a memoized value not be a better choice?

app/react/V2/Routes/Settings/ActivityLog/components/FiltersSidePanel.tsx
```diff
@@ -0,0 +1,174 @@
...
+  const [currentFilters, setCurrentFilters] = useState(appliedFilters);
+
+  const debouncedChangeHandler = useMemo(
+    () => (handler: (_args?: any) => void) => debounce(handler, 500),
+    []
+  );
+
+  useEffect(() => {
+    if (!isFirstRender) {
+      setCurrentFilters(appliedFilters);
```
Zasa-san: Similar to my other comment.

contents/ui-translations/ar.csv
```diff
@@ -211,6 +211,7 @@ Data imported,البيانات المستوردة
 Date added,تم إضافة التاريخ
 Date format,صيغة التاريخ
 Date modified,تاريخ التعديل
+Date Range,Date Range
```
Zasa-san: We do have key named `property daterange`. That key is used to translate the template property of date range type. Do you think that it would be worthwhile to reuse it?

Zasa-san (changes requested):
Partially done. Still need to review some other parts of the code and functionality. So far I see these problems issues in functionality:

Filter by name seems to not be working consistently
- Filter by a random string of characters (keyboard mash)
- Naturally get 0 results
- Filter by 'admin', a known username that has entries in the activity log
- Get 0 results
- Filter again by 'admin'
- Now I get results

Url params
- Filter by 'admin'
- The url stays the same
- Filter by admin again
- Now the url changes

Request filters:
- I see that there is a 'RAW' type that cannot be selected for filtering (this appeared when I published and entity)

app/react/V2/Components/Forms/DatePicker/DatePickerComponent.tsx
mfacar: Those names are much better, thanks!

app/react/V2/Components/Forms/DatePicker/DatePickerComponent.tsx
mfacar: Those names are much better, thanks!

app/react/V2/Components/Forms/DatePicker/DateRangePickerComponent.tsx
mfacar: That's right, this is because in our implementation we are using Flowbite-Datepicker, and the attributes in the input fields make the component behave like a calendar. Unfortunately, the clause datepicker is not recognized as a valid param of an HtmlInputElement

app/react/V2/Components/Forms/MultiSelect.tsx
mfacar: For the record, we've seen that useEffect is required to update the state when the props have changed. The first render condition effectively is not needed.

contents/ui-translations/ar.csv
mfacar: Using `property daterange` was my first approach but I thought that could be confusing, given your suggestion, I think it's fine to use the existing key and take advantage of already having some of the translations.

mfacar:
@Zasa-san, Thanks for your helpful review! Regarding the functionality, I could reproduce the absence of the param when the key enter is pressed on the filter field, it was because I implemented a debounce that was preventing the update of the form state before submitting the form, after reducing the debounce time it is working well.

The RAW action is not a specific HTTP method, it is shown as default when there isn't any `beautifier` parser that provides semantic information about the entry. We can extend the filtering logic in the backend to discriminate this case, but the current implementation is already quite complex. In my opinion, we should add a parser to handle the endpoint to publish entities.

app/api/activitylog/activitylogParser.ts
```diff
@@ -245,4 +245,4 @@ const getSemanticData = async (data: any) => {
   return { ...activityEntry };
 };
 
-export { getSemanticData };
+export { getSemanticData, entryValues as ParsedActions };
```
LaszloKecskes: At this point, doesn't it make sense just to remain the variable?

app/api/activitylog/activityLogFilter.ts
```diff
@@ -0,0 +1,252 @@
...
+  searchQuery: {}[] = [];
+
+  query: ActivityLogQuery;
+
+  constructor(requestQuery: ActivityLogQuery) {
+    const methodFilter = ((requestQuery || {}).method || []).map(method => method.toUpperCase());
+    this.query = { ...requestQuery, method: methodFilter };
+  }
+
+  timeQuery() {
```
LaszloKecskes, line 93: Can we try and type these properly? I think the first place that "looses" type is the reduce-sanitize where we try to remove null values. Then, the prepareToFromRanges input is already 'any'.

With that said, in general, a big props for finally moving this thing from js to ts!

app/api/activitylog/activityLogFilter.ts
```diff
@@ -0,0 +1,252 @@
...
+
+type ActivityLogQuery = ActivityLogGetRequest['query'];
+
+const prepareToFromRanges = (sanitizedTime: any) => {
+  const fromDate = sanitizedTime.from && moment.unix(parseInt(sanitizedTime.from, 10));
+  const toDate = sanitizedTime.to && moment.unix(parseInt(sanitizedTime.to, 10)).add(1, 'days');
+  return {
+    ...(fromDate && { $gte: fromDate }),
+    ...(toDate && fromDate !== toDate && { $lte: toDate }),
+    ...(toDate && fromDate === toDate && { $lt: toDate }),
```
LaszloKecskes: Isn't this last branch going to give you an empty set? Given a constant C, it's looking for things that are:
x >= C
x < C
Isn't leaving this $lte enough? Then you would just get x = C.

If we want to check something like this, maybe we could check that from<=to, and if not, throw an error. But I don't think that's necessarry.

app/api/activitylog/activityLogFilter.ts
```diff
@@ -0,0 +1,252 @@
...
+import { ParsedActions } from './activitylogParser';
+import { EntryValue } from './activityLogBuilder';
+
+type ActivityLogQuery = ActivityLogGetRequest['query'];
+
+const prepareToFromRanges = (sanitizedTime: any) => {
+  const fromDate = sanitizedTime.from && moment.unix(parseInt(sanitizedTime.from, 10));
+  const toDate = sanitizedTime.to && moment.unix(parseInt(sanitizedTime.to, 10)).add(1, 'days');
+  return {
+    ...(fromDate && { $gte: fromDate }),
```
LaszloKecskes: Double check: the mirror of this (i.e. only toDate) is covered by the next line, and is just going to give you an $lte: toDate.

app/api/activitylog/activityLogFilter.ts
```diff
@@ -0,0 +1,252 @@
...
+    ...(toDate && fromDate === toDate && { $lt: toDate }),
+  };
+};
+
+const parsedActionsEntries = Object.entries(ParsedActions);
+
+const queryURL = (matchedEntries: [string, EntryValue][]) =>
+  matchedEntries.map(([key]) => {
+    const entries = key.split(/\/(.*)/s);
+    const condition: { $and: {}[] } = {
```
LaszloKecskes: I don't think that the {}[] type is correct. Since this is something that goes to a mongoquery, what I would do for the sake of simplicity is:
- let TS infer the type
- make sure that it gets all the way to the mongo call through the functions that consume this result (i.e. we're not switching to any or unknown at some pont)
- let TS figure out if it matches what mongo needs or not

This works most of the time.

app/api/activitylog/activityLogFilter.ts
```diff
@@ -0,0 +1,252 @@
...
+      $and: [
+        {
+          url: {
+            $regex: `^\\/${entries[1].replace(/[.*\/+?^${}()|[\]\\]/g, '\\$&')}$`,
```
LaszloKecskes: There are multiple instances of this same regex. Are these for escaping special characters? Are they the same in every line that uses it? If yes, at least refactor it out into a separate function. 

Also, let's ask on the standup if we have some standardized way of doing this already.

app/api/activitylog/activityLogFilter.ts
```diff
@@ -0,0 +1,252 @@
...
+  queryMethods: string[],
+  value: EntryValue,
+  methods: string[]
+) =>
+  key.toUpperCase().match(`(${queryMethods.join('|')}).*`) &&
+  ((value.method || '').toUpperCase().match(`(${methods.join('|')}).*`) ||
+    value.desc.toUpperCase().match(`(${methods.join('|')}).*`));
+
+class ActivityLogFilter {
+  andQuery: {}[] = [];
```
LaszloKecskes: Let's try to type this here. My guess would be the `FilterQuery<T>[]` from mongoose, where T should be our log type.

app/react/V2/Components/Forms/MultiSelect.tsx
```diff
@@ -59,6 +59,10 @@ const MultiSelect = ({
     onChange(newValue);
   };
 
+  useEffect(() => {
```
Zasa-san, line 62: Nice work refactoring this! Is there a test case that would fail in the component test if we remove it? Do you think it's needed or will that be covered in a e2e?

app/react/V2/Routes/Settings/ActivityLog/components/FiltersSidePanel.tsx
```diff
@@ -0,0 +1,170 @@
...
+              <InputField
+                id="search"
+                label={<Translate>Search</Translate>}
+                className="my-4"
+                placeholder={t('System', 'by ids, methods, keywords, etc.', null, false)}
+                {...register('search')}
+                clearFieldAction={() => {
+                  setValue('search', '');
+                }}
+                onChange={debouncedChangeHandler(handleInputSubmit('search'))}
```
Zasa-san: Just to understand. What is the goal of debouncing here?

Zasa-san (changes requested):
I've noticed an error when paginating (I've forced the route to only load 5 items per page to get some pagination). When I reach the last page this odd thing happens with pagination component.

image

I've checked with, for example, the pagination in an IX table and this did not occur.

app/api/activitylog/activityLogFilter.ts
mfacar: It works perfectly, thanks!

app/api/activitylog/activityLogFilter.ts
mfacar: I changed the logic and added some tests to clarify the algorithm

app/react/V2/Components/Forms/MultiSelect.tsx
mfacar: Totally, I added a case in MultiSelect component tests.

app/react/V2/Routes/Settings/ActivityLog/components/FiltersSidePanel.tsx
mfacar: Really any goal, It was carried from the inline search, I just kept it to reduce the calls to setValue but it doesn't imply any significant computation so I removed it👍

mfacar:
> I've noticed an error when paginating (I've forced the route to only load 5 items per page to get some pagination). When I reach the last page this odd thing happens with pagination component.

Great catch! thanks!

Zasa-san (approved):
LGTM!

mfacar:
Closing in favor of #6866

PR #6787, opened 2024-05-21
6603 - [HOTFIX] - API error handling

### LINKED ISSUE #6263: SSRed pages not properly handling API errors
This is a follow-up of #6163 

Accessing non existent entities or documents can fall into one of this cases:
1. If it is an in-app navigation, it will show an infinitely loading screen
2. If it is a direct url navigation, it will show a JSON message stating that the server got a 404. The request itself will return a code 500.

This is related to missing implementation in the RouteHandler components. When they get a 404 from the server, the process will just fail. That means that:
1. If it was an in-app navigation: there's no handling of the 404 received, so the UI won't show accordingly,
2. If it was a direct navigation: The SSR utility will fail with an unhandled exception, and therefore return a 500.

### LINKED ISSUE #6603: Define Uwazi Custom Errors for responses from fetch at JSONRequest
Currently, we are creating a FetchResponseError for all the errors with response codes above 399. This makes it difficult to track errors in tools like Sentry where it's not possible to distinguish the nature or get statistics for a specific error.

Related to: #5494, #5540

mfacar:
fixes #6603 
fixes #6263

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/App/Root.js
```diff
@@ -129,7 +132,11 @@ Root.propTypes = {
   content: PropTypes.string,
   language: PropTypes.string,
   assets: PropTypes.object,
+<<<<<<< HEAD
```
Zasa-san: :face_with_open_eyes_and_hand_over_mouth:

app/react/App.tsx
```diff
@@ -45,7 +46,9 @@ const App = () => (
   <ReduxProvider store={store as any}>
     <CustomProvider>
       <Provider store={atomStore}>
-        <RouterProvider router={router} fallbackElement={null} />
+        <ErrorBoundary error={undefined}>
```
Zasa-san: I'm guessing the error prop can be just removed?

app/react/Pages/components/PageViewer.js
```diff
@@ -7,19 +7,12 @@ import React, { Component, Suspense } from 'react';
...
-import { ErrorBoundary } from 'app/App/ErrorHandling/ErrorBoundary';
 import { Icon } from 'UI';
 import { Translate } from 'app/I18N';
-import { ErrorFallback } from 'app/App/ErrorHandling/ErrorFallback';
-import { parseRenderingError } from 'app/App/ErrorHandling/ErrorUtils';
 import { NeedAuthorization } from 'app/Auth';
+import { ErrorBoundary, ErrorFallback } from 'app/V2/Components/ErrorHandling';
 import Script from './Script';
 
-const parseSSRError = error => {
```
Zasa-san, line 18: We are not doing this anymore because of the new `catch` in `app/react/App/RouteHandler.js`?

app/react/V2/Components/ErrorHandling/ErrorUtils.ts
```diff
@@ -0,0 +1,30 @@
...
+    status: 404,
+  },
+  500: {
+    name: 'Unexpected error',
+    message: 'Something went wrong',
+    status: 500,
+  },
+};
+
+export interface RequestError extends Error {
```
Zasa-san: I would move this type into `app/react/V2/shared` if you think we might use it at some other point.

app/react/V2/Components/ErrorHandling/ErrorUtils.ts
```diff
@@ -0,0 +1,30 @@
...
+    status: 400,
+  },
+  404: {
+    name: 'Not Found',
+    message: "We can't find the page you're looking for.",
+    status: 404,
+  },
+  500: {
+    name: 'Unexpected error',
+    message: 'Something went wrong',
```
Zasa-san: I think the old string was 'Unexpected error' and has now changed. Just a reminder to check the translations.

app/react/entry-server.tsx
```diff
@@ -299,11 +302,14 @@ const EntryServer = async (req: ExpressRequest, res: Response) => {
       user={req.user}
       reduxData={initialState}
       assets={assets}
+      loadingError={loadingError}
```
Zasa-san, line 309: On my IDE, loadingError is not defined in the context of EntryServer

app/react/entry-client.tsx
```diff
@@ -33,6 +40,6 @@ if (window.SENTRY_APP_DSN) {
 }
 
 const container = document.getElementById('root');
-const root = hydrateRoot(container!, <App />);
+const root = window.__loadingError__ === undefined ? hydrateRoot(container!, <App />) : container;
```
Zasa-san, line 43: If we have a `loadingError` we would just show Root.js? Is that correct.

Zasa-san (commented):
There a few code comments, generally it looks great! I still need to test functionality.

app/react/Pages/components/PageViewer.js
mfacar: That's right, now the ErrorFallback component is handling the error as it is.

Zasa-san:
LGTM

PR #6825, opened 2024-05-27
6824 - Silence warnings for React 19 deprecated APIs

### LINKED ISSUE #6824: Hundreds of warnings for deprecated features of the React next version
After #6752 and as it's described in the [upgrade guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide), there are hundreds of warnings in the browser console for deprecated APIs. It is ideal to silence these warnings while being aware of the problematic areas by decorating the window.console.error function.

mfacar:
fixes #6824

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/entry-client.tsx
+++ b/app/react/entry-client.tsx
@@ -65,5 +65,23 @@ const App = () => (
 
 const container = document.getElementById('root');
 const root = window.__loadingError__ === undefined ? hydrateRoot(container!, <App />) : container;
+const silentWarnings = [
+  'Warning: %s uses the legacy childContextTypes API which is no longer supported and will be removed in the next major release.',
+  'Warning: %s: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.%s',
+  'Warning: %s uses the legacy contextTypes API which is no longer supported and will be removed in the next major release.',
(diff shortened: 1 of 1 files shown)


Zasa-san (approved):
LGTM! You can still see the errors on the terminal when loading for the first time. I think that's a good thing so that we don't forget about them.
 
That is my only concern, that we have no way of remembering these silenced errors exist, except the case I mentioned above. Is there be a way that we could transform them into warning, instead of errors, on the console?

mfacar:
@Zasa-san, I think it is possible to call a `console.warn` but it would imply having the same hundreds of messages in the browser console, maybe is better to keep the warnings in the server terminal while we start planning the upgrade to React 19, which was already released as RC a month ago.

Zasa-san:
> @Zasa-san, I think it is possible to call a `console.warn` but it would imply having the same hundreds of messages in the browser console, maybe is better to keep the warnings in the server terminal while we start planning the upgrade to React 19, which was already released as RC a month ago.

Makes sense! 👍

mfacar:
Information extraction cypress tests consistently fail across all the PRs, so ignoring that suite

PR #6830, opened 2024-05-28
HOTFIX Abstract file not found error in storage

fnocetti:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/files/storage.ts
+++ b/app/api/files/storage.ts
@@ -18,6 +18,7 @@ import {
   uploadsPath,
 } from './filesystem';
 import { S3Storage } from './S3Storage';
+import { FileNotFound } from './FileNotFound';
 
 type FileTypes = NonNullable<FileType['type']> | 'activitylog' | 'segmentation';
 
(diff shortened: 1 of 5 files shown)


app/api/files/storage.ts
```diff
@@ -59,15 +60,32 @@ const readFromS3 = async (filename: string, type: FileTypes): Promise<Readable>
...
-    return createReadStream(paths[type](filename));
+    return catchFileNotFound(async () => {
+      if (tenants.current().featureFlags?.s3Storage) {
+        return readFromS3(filename, type);
+      }
+      return createReadStream(paths[type](filename));
+    }, filename);
   },
   async fileContents(filename: string, type: FileTypes) {
-    return streamToBuffer(await this.readableFile(filename, type));
```
LaszloKecskes, line 70: Isn't this function the one above? Do we need to wrap again?

LaszloKecskes (changes requested):
Looks good, one question.

app/api/files/storage.ts
fnocetti: The writable stream returned by `readableFile` doesn't work with promises (i.e. it doesn't throw) but works with event listeners. The job of steamToBuffer is to promisify it, so we can only "catch" the error at this level.
I happened to design the wrapper to work with s3 or local file not found errors.  So I'm using it in both places where it can "throw". We can definitively split it into 2 different wrappers if you think it adds value.

LaszloKecskes: Makes sense based on the discussion, thanks!

LaszloKecskes (approved):

fnocetti:
Converting this into a stabilization fix to be merged in #6832

PR #6850, opened 2024-05-31
6843 UI does not respect the default language

### LINKED ISSUE #6843: UI does not respect the default language
In Uwazi instances with more than one language installed, it seems that the UI always appears in English at login.

To reproduce the problem:

- Log in to an instance and install two or more languages from ‘/settings/languages’.
- Select a default language other than English.
image

- Log out of the system; log in to Uwazi again.
- Problem: the instance is back to the English UI.
- If you select another language, the UI changes accordingly, but when logging out and logging in again, it insists on English.

I believe that this behaviour was different some weeks ago.

mfacar:
fixes #6843

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


Zasa-san (approved):
LGTM

PR #6854, opened 2024-06-03
Account changes fix

Zasa-san:
PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/V2/Components/UI/ConfirmationModal.tsx
```diff
@@ -71,6 +74,25 @@ const ConfirmationModal = ({
...
         )}
+
+        {usePassword && (
+          <div className="py-4">
+            <span className="block mb-2 font-medium text-gray-900 text-md dark:text-white">
+              <label htmlFor="confirm-password">
+                <Translate>Enter your current password to confirm</Translate>&nbsp;
+              </label>
+            </span>
+            <input
```
mfacar, line 85: Is there a way to disable autocomplete for the input? When the password is stored in the browser the confirmation input is filled and all the actions are consequently confirmed.

app/react/V2/Components/UI/ConfirmationModal.tsx
```diff
@@ -8,12 +8,13 @@ type confirmationModalType = {
   size?: modalSizeType;
   header?: string | React.ReactNode;
   body?: string | React.ReactNode;
-  onAcceptClick?: () => void;
+  onAcceptClick?: (value: string) => void;
```
mfacar, line 11: Given the parameter passed to this function is  `passwordValue` is it possible to keep that name here?

mfacar (approved):
LGTM!

Maybe it is intended but I noticed the following regarding the functionality.
On Users: 
   - When the password is not valid, the side panel is closed, and it is necessary to open it again to continue the action, in the case of creating a user the form remains dirty, the edition however is reseting the values. It would be great, if it is possible to keep the side panel opened after the error, with the dirty fields for both cases.
   - Reset Password doesn't ask for the confirmation password, as I think it was agreed by the team, however, there is a pre existent validation error while applying the action on many users.
   
On Account:
   - The email field is also reset after a password error. 
    
typos on submmitedPassword and worngPass

app/react/V2/Components/UI/ConfirmationModal.tsx
Zasa-san: My thinking here is that later on we might want to return other things than a password when the user confirms. I'm going to change the internal state name so that it's `inputValue` so that it's more consistent.

app/react/V2/Components/UI/ConfirmationModal.tsx
Zasa-san: I can disable the autocomplete for sure. `all the actions are consequently confirmed.` Do you mean that the modal auto accepts, without having to click on the accept button?

mfacar: No, the actions are not auto accepted, I'm sorry I expressed myself wrong. Given the browser fills the confirmation modal with the stored valid password, any confirmation modal with this control will just require clicking on accept to get approved.

Zasa-san:
>  When the password is not valid, the side panel is closed, and it is necessary to open it again to continue the action, in the case of creating a user the form remains dirty, the edition however is reseting the values. It would be great, if it is possible to keep the side panel opened after the error, with the dirty fields for both cases.

That's a good point. I will leave the sidepanel open when the password is incorrect. That way the user will not loose changes.

> Reset Password doesn't ask for the confirmation password, as I think it was agreed by the team, however, there is a pre existent validation error while applying the action on many users.

If you are testing this with our e2e fixtures that happens because one of the users (`editor`) doesn't have an email. That's not possible to archive without manually manipulating the DB and removing the user email, since you can't create a user without an email.

> The email field is also reset after a password error.

Will fix!

> typos on submmitedPassword and worngPass

Thanks for the catch :eye: 

@mfacar

app/react/V2/Components/UI/ConfirmationModal.tsx
Zasa-san: Let me know if the latest change fixes this in your environment. :pray:

app/react/V2/Routes/Settings/Account/Account.tsx
```diff
@@ -1,58 +1,73 @@
...
     userData.password = userData.password ? userData.password : userAccount.password;
-    await UsersAPI.save(new RequestParams(userData));
-    await revalidator.revalidate();
-    setNotifications({
-      type: 'success',
-      text: <Translate>Account updated</Translate>,
-    });
-  };
 
-  useEffect(() => {
```
Zasa-san, line 53: @konzz  I'm not sure why this effect is here, since I can't find a problem with removing it. (Asking you since I think you added this initially)

app/react/V2/Components/UI/ConfirmationModal.tsx
mfacar: It seems like the required option for autocomplete is "new-password". However some browsers list the stored password and for sure that's out of the control of app, so please don't care about it.

mfacar (approved):

fnocetti (approved):

PR #6861, opened 2024-06-05
HOTFIX Adjust password recovery ttl. Validate empty login credentials.

### LINKED ISSUE #6834: The "create password" process for new user fails
**Describe the bug**
When a new user tries to create a password, the process fails. As a consequence, the user cannot login to their account.

**To Reproduce**
1. Create a new user
2. Open the link to create a password in the welcome email
3. Use the form to provide a password and click save

**Expected behavior**
The user must be able to create a password that can use to log in their account.

**Screenshots**

Error message

image

 - [x] Desktop
 - [ ] Mobile

**Browser**
Safari 17.5 on MacOS 12.7.5

### LINKED ISSUE #6860: Login + blank password + hitting enter bricks the instance for that particular user
Attempting to login into Uwazi with a blank password AND hitting the enter key (not hitting the login button with the mouse) will brick the instance for that particular user due to something related with the cookie. It will display this error:

image

And attempting to reload the instance will show how it is bricked.

image

Clearing the cookies for this instance will resolve the problem.

fnocetti:
fixes #6860 
fixes #6834 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/auth/routes.js
+++ b/app/api/auth/routes.js
@@ -46,8 +46,8 @@ export default app => {
         body: {
           type: 'object',
           properties: {
-            username: { type: 'string' },
-            password: { type: 'string' },
+            username: { type: 'string', minLength: 1 },
+            password: { type: 'string', minLength: 1 },
             token: { type: 'string' },
           },
           required: ['username', 'password'],
--- a/app/api/auth/specs/__snapshots__/routes.spec.js.snap
+++ b/app/api/auth/specs/__snapshots__/routes.spec.js.snap
@@ -6,12 +6,14 @@ Object {
     "body": Object {
       "properties": Object {
         "password": Object {
+          "minLength": 1,
           "type": "string",
         },
         "token": Object {
           "type": "string",
         },
         "username": Object {
+          "minLength": 1,
           "type": "string",
         },
       },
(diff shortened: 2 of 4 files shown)


app/api/auth/routes.js
```diff
@@ -46,8 +46,8 @@ export default app => {
         body: {
           type: 'object',
           properties: {
-            username: { type: 'string' },
-            password: { type: 'string' },
+            username: { type: 'string', minLength: 1 },
+            password: { type: 'string', minLength: 1 },
```
aphilop, line 50: If I understand correctly, this would allow a password of one character, right?
Shouldn't a password be at least 8 characters in length (the current best practice)?

konzz: Yes, but we don't have that policy right now, so if we put that validation in the login we will block many users from being able to log in, probably we will need to change the creation/edition of passwords, and then after some time maybe change it here.

fnocetti: Yeah, I think we are mixing stuff :) We might definitively want to revisit our password policy. That's on user creation and password updates, where we decide what password a user can have.
This is the login. Technically, a user can try to login with any password, as long as they provide some.

aphilop: Ok, thanks for clarifying.

LaszloKecskes (approved):
Functionally tested both issues, works as intended.

Approved with the understanding that issues will be written for the outstanding tech debt (testing, other non-critical fixes).

fnocetti:
Thanks @LaszloKecskes. For the record, this is the tech debt issue #6862

PR #6886, opened 2024-06-11
6873 - Moved login inside changes array

### LINKED ISSUE #6873: Move Sync login logic only if LastChanges has items
Currently, Sync process logs in once per cycle, independent of finding changes or not.

Ideally, this should be more evolved and actually only log-in if the changes affect the child instance (and could even keep the cookie to avoid new login), but, for the time being, we can move this logic further down the stream to avoid logging in more than once per second.

RafaPolit:
fixes #6873 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


LaszloKecskes (approved):
Looks good, works. Functionally tested through a simple local sync setup and logging logins.

Optionally, we have a `syncedPromiseLoop` function in the shared folder, if you'd like to simplify the reduce-await syntax.

PR #6897, opened 2024-06-13
6894 - Create ValidatedElement component and sanitized Markdown.

### LINKED ISSUE #6894: Using malformed HTML can lead to a server error
Using invalid HTML in rich texts and pages can trigger an error.

Here is a simplified text that breaks:

```
<b>Little red ridding hood</b>

<p>I don't know <..long pause..> a minute later  <..pause..> a grandma story.
When I heard it</p>
```

RafaPolit:
fixes #6894

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/Markdown/specs/__snapshots__/MarkdownViewer.spec.js.snap
```diff
@@ -4,24 +4,18 @@ exports[`MarkdownViewer render should be able to render properly custom componen
 <div
   className="markdown-viewer"
 >
-  <div
-    key="0"
```
Zasa-san, line 8: I am concerned the 'key' element gets removed everywhere in the snapshot.

app/react/Markdown/utils.js
```diff
@@ -13,3 +13,121 @@ export const logError = (err, propValueOf, propLabelOf) => {
   console.error(err);
   /* eslint-enable no-console */
 };
+
+export const validHtmlTags = new Set([
```
Zasa-san: As you mentioned, iframes will not work since they are not considered valid. I don't have an opinion on whether we should include them or not. I do think a user might want to embed a video from another platform in a custom page, and would not be able to without iframes.

Same with SVGs. Someone using custom icons on their page would get an invalid tag error.

Zasa-san (changes requested):
I think the complexity of the recursive function is ok. It's readable.

app/react/Markdown/utils.js
Zasa-san: We have decided to allow svg and iframe.

PR #6927, opened 2024-06-21
Clean insights

johnhess:
No associated issue; pre-arranged with CTO.

Scope:

- Stub out a get_analytics_opt_in() function.  It will always return true along with logging a message to that effect.
  - Seeing as this may require another roundtrip to the server, we'll make sure the page load doesn't block waiting on this.
  -Based on our conversation, I'm assuming you all would be implementing the server-side endpoints to surface this data.  If you have documentation on the endpoints that report whether the instance is self hosted and, if not, whether it's opted in, we're happy to provide a real implementation of get_analytics_opt_in()
  -Note: I've not included any server-side engineering here.  If you would like that built out as well, let me know and I can look into what that would take.

- Capture DAU, WAU, MAUs using the Clean Insights JS SDK
  - This will capture logged-in and anonymous users.
  - This will aggregate on-device in 1, 7 and 30 day periods.  The user's browser will report their membership in the set of active users after the period of interest (e.g. after the month ends).
    - Note: during implementation, I selected 28 day intervals for MAUs so that all periods contain the same # of weekdays/weekends). This is configurable, but should be set before releasing.
  - Data will be routed to a single Matomo instance/siteID.  During development, we'll use Clean Insights' community server, but you can switch over to your own server by changing config values.
  - We'll embed this in a default template so it loads everywhere users are opted in.
  - At your option, we can capture additional data, such as the domain or instance name (assuming this is available client side).

- No consent UI is required for this.  You'll manage consent in your relationships with those you self host and will populate your tenants database accordingly.
- We'll develop a basic dashboard in Matomo and share screenshots and instructions on how to duplicate and interpret it along with our Pull Request.
- The final deliverable will be a Pull Request to github.com/huridocs/uwazi

PR checklist:
- [N/A] Update READ.me ?
- [N/A] Update API documentation ?

QA checklist:
- [ ] Smoke test the functionality described in the issue (Checked that data is recorded & sent to server as described)
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


johnhess:
> We'll develop a basic dashboard in Matomo and share screenshots and instructions on how to duplicate and interpret it along with our Pull Request.

The Matomo instance is set up, and has a very minimal amount of data in it at present (from testing).  I'm happy to provide accounts on that server to the appropriate folks on your end, or you can use your existing Matomo installation along with the [Clean Insights Matomo Proxy](https://docs.cleaninsights.org/docs/cimp/).

johnhess:
CC @Anderson526

konzz:
Hi @johnhess we use yarn to handle dependencies and one of the CI steps is failing because the `yarn.lock` was not updated after adding the Clean Insights library, could you please commit it? Thanks!

johnhess:
> yarn.lock was not updated after adding the Clean Insights library, could you please commit it?

Committed!

PR #6933, opened 2024-06-24
Files healthcheck

fnocetti:
fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/files/specs/storage.spec.ts
```diff
@@ -317,6 +318,126 @@ describe('storage', () => {
     });
   });
 
+  describe('listFiles', () => {
+    describe('when s3 flag is active', () => {
+      it('should list all files on disk', async () => {
```
LaszloKecskes: "...all files on **disk**" should be on s3.

fnocetti: Nice catch. Thanks

fnocetti:
I see the unit tests are failing. It's kind of awkward because the files list looks differently depending on your local setup. Will check.

app/api/files/specs/storage.spec.ts
```diff
@@ -317,6 +318,126 @@ describe('storage', () => {
...
+        testingTenants.changeCurrentTenant({
+          name: 'tenant1',
+          dbName: 'uwazi_development',
+          indexName: 'index',
+          featureFlags: {
+            s3Storage: false,
+          },
+        });
+        await storage.storeFile(
+          'file_created.txt',
```
LaszloKecskes: Missing cleanup for this file, it remains on disk after the test.  Do we need to create a file here? This is only testing the listing functionality, having it to test only the known existing files could be good enough.

Also, it took me some work to get this green locally, since I had a lot of extra files in my folders, probably due to early stopped tests. When this goes live, let's drop a few lines on the tech channel to warn everybody that they might need to cleanup their local upload folders in the dev environment.

app/api/files/specs/storage.spec.ts
```diff
@@ -317,6 +318,126 @@ describe('storage', () => {
...
+      it('should list all files on disk', async () => {
+        testingTenants.changeCurrentTenant({
+          name: 'tenant1',
+          dbName: 'uwazi_development',
+          indexName: 'index',
+          featureFlags: {
+            s3Storage: true,
+          },
+        });
+        await storage.storeFile(
```
LaszloKecskes, line 334: Also missing cleanup here. It's less problematic in S3, since the big afterAll dismantles the test buckets, but technically, it still could interfere with other tests in this file, in one particular run. Let's not leave around resources after testing even if we know that at the moment they do not cause problems, just as we would do with restoring mocks.

app/api/files/storage.ts
```diff
@@ -55,6 +55,21 @@ const s3KeyWithPath = (filename: string, type: FileTypes) => {
   );
 };
 
+const fromS3Key = (key: string) => {
+  const [, type, nestedTypeOrFilename, ...rest] = key.split('/');
+  if (!['segmentation', 'activitylog'].includes(nestedTypeOrFilename)) {
```
LaszloKecskes: Is this because these are in subfolders, and S3 handles those as parts of the key?

fnocetti: Correct. There's already a toS3Key function. This is just the inverse.

app/api/files/storage.ts
```diff
@@ -121,6 +136,37 @@ export const storage = {
...
+          if (file.isFile()) {
+            files.push({
+              type: filetype,
+              filename: file.name,
+              path: join(paths[filetype](''), file.name),
+            });
+          }
+        });
+      } catch (err) {
+        if (err?.code === 'ENOENT') {
```
LaszloKecskes, line 141: What can cause an ENOENT that we ignore it here? The files are already given by a readdir, so those should be there. A missing folder from the ones configured in `path`?
If it's a missing folder, is that ok, since we're auto creating it when they are needed and it's fine that they are not present for a given tenant?

fnocetti: > What can cause an ENOENT that we ignore it here? The files are already given by a readdir, so those should be there. A missing folder from the ones configured in `path`?

Correct.

> If it's a missing folder, is that ok, since we're auto creating it when they are needed and it's fine that they are not present for a given tenant?

There will be other places in the app that will handle such scenarios. For this particular API, the semantic is "list every file there". I think that, when there's no directory for one type of file, we should assume "no files there" rather than erroring, but it's discussable.

app/api/files/specs/storage.spec.ts
fnocetti: I added new files to test the different file types.
The problem of the extra files is a tricky one. It's right now failing in the pipeline runs likely because of something around that. I will try to find a strategy to work that around.
Regarding clean up, I didn't do it because other test cases neither do it. But may be the solution is to clean up all of them.

app/api/files/specs/storage.spec.ts
fnocetti: Same as above.

app/api/files/storage.ts
LaszloKecskes: This was more of an "double-check" type question. If we're handling these in other places well, here assuming no files as you say is good.

app/api/files/specs/storage.spec.ts
LaszloKecskes: Yeah, I've checked and it looks tough. It looks like there are other tests in Uwazi that do not clean up files properly. And cleaning it all up is probably not good, since there are files that are needed by default (as you've marked in the comment), so we should leave those intact. 
Maybe we could setup different folders for this one test, then cleanup those? But that's just an idea, I'll let you decide, whatever you think it's best in the context.

LaszloKecskes (changes requested):
It looks good and works, aside from the problems mentioned in the test.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "uwazi",
-  "version": "1.172.4",
+  "version": "1.172.5",
```
LaszloKecskes: Just a reminder, because this is something I would potentially miss: we've had a deploy since, current version will be 1.173.x

fnocetti: Appreciate you noting it! That's usually (and the case for this one) signaled as a git conflict :)

LaszloKecskes (approved):
The changes look good, works.

PR #6940, opened 2024-06-27
6686 remove entity save multiple

### LINKED ISSUE #6686: [IX] Review the usages of entities.saveMultiple
We have an old entities.saveMultiple function, which is an old function that does not contain all hooks that the singular save function does (in order to be more efficient). Most notably, the denormalization functionality is not triggered by it. 

On a cursory glance, it is currently used in the places: in setting permissions, and accepting suggestions in IX. Not having these extra actions might cause problems in the future, for example:
- accepting a suggestion for a property that is inherited through a relationship will not update the denormalized value

The task is:
- review the saveMultiple functionality to see what is missing
- review it's usages to see whether the missing functionality might cause an error
- consider a solution

Since these usages are most likely getting input through user interaction, and thus will have short inputs, a simple solution would be just remove this saveMultiple function and save those entities one by one, as we usually do in other places in Uwazi.
```[tasklist]
### Tasks
```

LaszloKecskes:
fixes #6686 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti (commented):
The code looks good. I'll proceed to the manual testing.
One question is: the ix side will be handled in other PR?

LaszloKecskes:
> The code looks good. I'll proceed to the manual testing. One question is: the ix side will be handled in other PR?

That's already done by the IX relationships PR. There we needed the full entity save for all the denormalizations.

fnocetti (approved):
I tested the permissions functionality:
- Setting permissions
- Publishing entities
And checked that:
- The metadata is preserved and correct.
- The permissions are as set.
- That the search results are correct after changing permissions.

LGTM

PR #6943, opened 2024-06-27
Thesaurus label matching hotfix

### LINKED ISSUE #6942: Thesaurus label case error on csv import
The csv import throws an error on importing new elements for existing groups with an upper case letter in their names. The error is caused by the thesauri.appendValues function using the the original label instead of the normalized value in one comparison.

To reproduce:
1, Create a thesaurus that has a group name with capital letter(s).
Group: A
2, Create a template that uses this for a select property.
3, Try to import an entity through csv, that has a new value in this thesaurus, in this group:
title;select
imported1;Group::B

LaszloKecskes:
fixes #6942

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


fnocetti (approved):
The code looks good and it works.
What I'm wondering is: creating thesaurus values through the thesaurus ui allows for labels that only differ in casing. e.g. "a" and "A" can co-exists.
Is this "case insensitivity" feature desired in the csv import?

LaszloKecskes:
> The code looks good and it works. What I'm wondering is: creating thesaurus values through the thesaurus ui allows for labels that only differ in casing. e.g. "a" and "A" can co-exists. Is this "case insensitivity" feature desired in the csv import?

As far as I can remember the discussion, yes. We are explicitly testing for it, we've just missed this case on the csv side when added the option to import grouped values. So I've added some more thorough testing on the thesaurus side.

fnocetti:
> > The code looks good and it works. What I'm wondering is: creating thesaurus values through the thesaurus ui allows for labels that only differ in casing. e.g. "a" and "A" can co-exists. Is this "case insensitivity" feature desired in the csv import?
> 
> As far as I can remember the discussion, yes. We are explicitly testing for it, we've just missed this case on the csv side when added the option to import grouped values. So I've added some more thorough testing on the thesaurus side.

Makes sense!

PR #6990, opened 2024-07-09
[Stabilization-fix] - Media file creation fix

### LINKED ISSUE #6969: Intermittent e2e: cypress/e2e/pages/public-form.cy.ts
Apparently, the test intermittency resides upon trying to check the results:  the click on "Restricted" files is not accomplished and, therefore, the check fails as unsuccessful.

### LINKED ISSUE #6986: Public form with media fields not doing cleanup
Public forms are not removing locally stored media files after successfully finishing a submit. We are using object urls https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static to show new media files in the form fields (like we do for the metadata form). After a submission is done we should remove these objects from memory to avoid problems with the browsers' storage. 

The public form is calling the cleanup method with the incorrect parameter (it should be the object url itself) at the incorrect time (it's calling it before submitting, so if the submission fails due to an incorrect captcha it would remove the object url causing problems for the next submit attempt).

https://github.com/huridocs/uwazi/blob/b014e7ae462a9abd204030cd83d582d5c684809a/app/react/Metadata/helpers/wrapper.js#L39

Currently this is not breaking the public forms since we are not really deleting the object URL, so even if the user enters the wrong captcha, the object will remain in memory for the next submission attempt. But to fix the cleanup process we also need to delete these objects at the correct time (likely once the submission has successfully completed)

mfacar:
fixes [155](https://github.com/huridocs/Internal-Issues/issues/155)
fixes #6986 
fixes #6969 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/Markdown/components/PublicForm.js
+++ b/app/react/Markdown/components/PublicForm.js
@@ -54,17 +54,8 @@ class PublicFormComponent extends Component {
 
   constructor(props) {
     super(props);
-    this.removeAttachment = this.removeAttachment.bind(this);
-    this.fileDropped = this.fileDropped.bind(this);
-    this.handleSubmit = this.handleSubmit.bind(this);
-    this.validators = {
-      captcha: { required: val => val && val.text.length },
-      ...validator.generate(props.template.toJS()),
-    };
-    const generatedIdTitle = props.template
-      .get('commonProperties')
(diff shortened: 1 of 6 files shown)


konzz:
LGTM

PR #7028, opened 2024-07-23
Native fetch on reindex_elastic script

daneryl:
- this is a test with native fetch to avoid econreset error during ci pipelines


database/reindex_elastic.js
```diff
@@ -83,18 +91,21 @@ const prepareIndex = async () => {
...
-    setTimeout(async () => {
-      await request.put(getIndexUrl(), elasticMapping);
-    }, 1000);
-  }
+  await fetch(getIndexUrl(), {
+    headers,
+    method: 'PUT',
+    body: JSON.stringify(elasticMapping),
+  });
+
```
RafaPolit, line 99: I was wondering about this timeout and agree we probably don't need it and it was already a smell of something wrong with the process flow and library resets.  Nice!

database/reindex_elastic.js
```diff
@@ -17,13 +16,22 @@ const getIndexUrl = () => {
   return `${elasticUrl}/${config.defaultTenant.indexName}`;
 };
 
+const headers = {
+  Accept: 'application/json',
+  'Content-Type': 'application/json',
+};
+
```
RafaPolit, line 23: As we move forward with replacing libraries with the native Fetch, I would probably put this on an abstracted request object, but much simpler (and less responsibilities) than the one we have now.

RafaPolit (commented):
LGTM.

PR #7084, opened 2024-08-01
Improve references

Zasa-san:
PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


konzz:
My only comment would be that to refactor `activateReference` to instead of receiving 5 arguments, expect some of the arguments as in an options object, like the `delayActivation` or `showContextMenu` because I see that in different scenarios you had to pass "undefined" arguments to pass the last argument. But also I can see that the function is widely used so this change could be a bit tricky

Apart from that LGTM

Issue #6431, opened 2024-02-05
Difficult to determine editing row on larger screens


RafaPolit:
On large screens, our design makes it really hard to know what row your cursor is targeting (in this case, for example) for editing:

<img width="1906" alt="image" src="https://github.com/huridocs/uwazi/assets/4309019/3d7dac29-5a80-4ab9-a255-a1581a384796">

As you can see, it's hard to determine which row you are actually interacting with.  Could we introduce a hover highlight to help visually determine which row you are at? Or maybe a striped table could be enough?

cc @juanmnl @konzz

konzz:
Yes, it makes sense, Juanma is this in any design and we missed it, what about adding the same gray background the header has on hover?

Issue #6837, opened 2024-05-30
Map settings - Can not type Latitude and Longitude


salvalacruz:
The Map settting in the '/settings/collection' screen does not allow to enter geographical coordinates.

User is allowed to select a point in the map whit no problems, but typing exact Latitude+Longitude is not posible.

![image](https://github.com/huridocs/uwazi/assets/73410334/32e868f8-27e5-481b-8570-8c8eaaf03807)

RafaPolit:
The Medium priority reflects the fact that clicking works, so you can bypass the problem fairly easily.  But these type of errors should not escape our testing processes.

Issue #6757, opened 2024-05-09
Popup for contextual help is not displayed properly


aphilop:
**Describe the bug**
When a popup is displayed on the left of the help (?) icon, then the popup is "cut" on the left side of the screen. As a result, the full content of the popup is not visible. This does not seem to happen when the popup is displayed on the right of the icon.

**To Reproduce**
1. Go to **Settings > Collection**
2. Scroll down to the **Forms and email configuration** section
3. Hover over the ? icon of the **Contact form email** field
4. Popup is not displayed properly

**Expected behavior**
The popup should display the whole textual content so that if fulfills its purpose (provide contextual help).

**Screenshots**
<img width="1434" alt="image" src="https://github.com/huridocs/uwazi/assets/102693686/dbb82803-eca1-4b03-86e5-49e61ab480f6">

**Device (please select all that apply)**
 - [x] Desktop
 - [x] Mobile

**Browser**
Safari, Firefox on MacOS (12.7.4)
Safari on iOS (17.4.1)

chinmaym07:
Hey @RafaPolit  can I fix it ?

RafaPolit:
Hello @chinmaym07 .  Yeah, sure.  Give it a go! Thanks.

Maybe it would be worth it to discuss the approach you intend to take to fix this?  Also, be sure to run the tests to ensure they are passing... but yeah, external contribution is more than welcome!

Issue #6435, opened 2024-02-05
Improve settings layout


Zasa-san:
**(This issue is to document a spike to refactor the settings main layout)**

Currently, in the new settings UI, most of the pages have a footer with actions buttons.

This footer needs to remain always visible while the contents scroll if necessary. To achieve this we made the footer fixed to the bottom, and added a padding to the content to avoid it being covered by the footer.

This is the wrong approach but the only one we could achieve since we need to insert this UI in the context of the V1 settings route. It also causes responsible issues on mobile for the V2 settings screens.

A better approach would be to make the whole settings layout like this:

![imagen](https://github.com/huridocs/uwazi/assets/71732018/4f14db07-c9f3-4450-a020-fb404e85426c)

By nesting two flex or grid containers like this we can get the desired behavior of having independent scrolling in the content (with footer and header always visible), and scrolling in the side menu if necessary. This is also completely responsive.

The current hurdle to do this is that we need to refactor the parent element of all these routes, `app/react/Settings/Settings.js`. This breaks V1 setting pages, and the sidebar menu. The amount of work necessary to make V1 setting pages in this new layout is almost as much as rewriting them, since we are almost halfway there.

cxx5208:
Could you please assign this to me. I would like to work and improve the styling

Zasa-san:
> Could you please assign this to me. I would like to work and improve the styling

Hey @cxx5208 ! Thanks for your interest in our project!

This issue is not ready to be developed. You can certainly create a branch and develop a solution for practice, but It wouldn't be merged.

You can find other issues with the Frontend tag if you are interested in working with styles.

PR #7131, opened 2024-08-19
Tables migration

### LINKED ISSUE #7052: Migrate to new table
After #6826 is merged we need to use the new table for all the v2 UIs.

We should migrate all uses of `V2/Components/UI/Table` and replace it with `V2/Components/UI/TableV2`, and remove the corresponding story and test. Lastly, rename `TableV2` module, story, and test to `Table`.

This is important because `TableV2` handles groups differently and has a corrected dnd functionality. 

Keep in mind that the new table requires that all its rows and sub rows have a unique `rowId` to properly handle dnd, selection and grouping. 

Doing this migration will fix: #6431& #6432

Sections under Settings where migration to new table is needed:
- [ ] User & Groups
- [ ] Menu
- [ ] Pages
- [ ] Languages
- [ ] Translations
- [ ] Filters
- [ ] Bug in Filters: Moved here from issue #6820 
       **Describe the bug**
When a user extracts all the items from the group in the filter settings, the group becomes a filter. It seems to be a bug because we use entity types (templates) for the creation of primary filters; labels that we use for grouping should not become filters. As a result, when a former group name is displayed on the filters panel, it always shows 0 entities, because there is no template associated with it.
**To Reproduce**
     1. Go to Settings and navigate to Filters from the left menu
     2. Add group and add entity types to the group
     3. Save the change
     4. Drag all the entity types belonging to the group outside of the group
     5. See that the group is displayed as a filter item as a result
- [ ] Relationship Types
- [ ] Activity Log
- [ ] Uploads
- [ ] Migrate FileDropzone to use dnd-kit (from #7065)
       We should migrate FileDropzone to use dnd-kit, instead of react-dnd so that v2 uses one library for drag and drop functionality. There’s also a DragAndDrop.stories.tsx for an implementation of dnd that’s not used in v2 and should also be removed. If https://github.com/huridocs/uwazi/issues/7052 is done, react-dnd can be removed as a dependency.

Zasa-san:
fixes #7052 #6431

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/V2/Routes/Settings/Filters/components/helpers.ts
```diff
@@ -60,65 +64,69 @@ const updateFilters = (
...
+  filters?.forEach(filter => {
+    const { rowId, subRows, ...sanitizedFilter } = { ...filter };
+
+    if (subRows && subRows.length === 0) {
+      return;
+    }
+
+    if (subRows) {
+      sanitizedFilter.items = subRows.map(item => {
+        const { rowId: itemRowId, _id, ...sanitizedItem } = { ...item };
```
mfacar: The destructuration of { ...item } seems not to be needed, actually { rowId: itemRowId, _id, ...sanitizedItem } could be directly in the params of the map like subRows.map(({rowId: itemRowId, _id, ...sanitizedItem }) => ...)

app/react/V2/Routes/Settings/Filters/FiltersTable.tsx
```diff
@@ -70,43 +71,57 @@ const FiltersTable = () => {
...
         type: 'error',
         text: <Translate>An error occurred</Translate>,
         ...(response.message && { details: response.message }),
       });
     }
-
     setSettings(response);
     setDisabled(false);
     setHasChanges(false);
+    const formattedFilters = formatFilters(savedFilters || []);
```
mfacar: Could it be that the reported inconsistencies are related with keeping the local filters instead of the data in the response?

app/react/V2/Routes/Settings/Filters/components/FiltersSidepanel.tsx
```diff
@@ -60,19 +60,21 @@ const FiltersSidepanel = ({
...
     });
 
-  const handleSave = (values: Omit<ClientSettingsFilterSchema, 'items'> & { items: string[] }) => {
-    const result = { ...values, items: formatSelected(values.items) };
+  const handleSave = (values: Omit<Filter, 'subRows'> & { subRows: string[] }) => {
+    const result = { ...values, subRows: formatSelected(values.subRows) };
 
     if (!filter?._id) delete result._id;
     if (!filter?.id) result.id = uniqueID();
+    if (result.rowId === defaultValues.rowId) result.rowId = uniqueID();
```
mfacar, line 77: To avoid this comparison, uniqueID() could be assigned directly instead of the value 'NEW'

mfacar (changes requested):
Great job! I did some tests having the following comments:
* The Filters page has some unexpected behaviors after doing some editions:
      - Editing an instance without filters, adding several filters, saving, deleting 1 filter, the table is emptied, the data is correct after reloading.
      - After some editions, DnD starts working wrongly, some elements can't be moved out or into a group, or moving an item into a group ends with moving other items into it or changing the order. Two times when this  happened, select all starts a loop rendering having the error "Maximum update depth exceeded"
* Some actions column headers don't have the bottom border. 
Screenshot 2024-08-28 191331
* Filters and Menu items are not translatable.
* It isn't clear to me what is the solution for #6431

There are also a couple of things from previous developments that maybe can be easily fixed in this PR.
- Filters action buttons have a different order than other pages
Screenshot 2024-08-28 192426
- Relationship Types show the translated value followed by the original name wrapped by parenthesis, this is different in other pages like templates 
Screenshot 2024-08-28 192927

Zasa-san:
> Great job! I did some tests having the following comments:
> 
>     * The Filters page has some unexpected behaviors after doing some editions:
>       - Editing an instance without filters, adding several filters, saving, deleting 1 filter, the table is emptied, the data is correct after reloading.
>       - After some editions, DnD starts working wrongly, some elements can't be moved out or into a group, or moving an item into a group ends with moving other items into it or changing the order. Two times when this  happened, select all starts a loop rendering having the error "Maximum update depth exceeded"

This was a mistake in rowIds, fixed!

> 
>     * Some actions column headers don't have the bottom border.
This is a problem on all these new tables, if it's simple I'll fix it in this PR, otherwise it's going to be fixed separately.
> 
>     * Filters and Menu items are not translatable.

Menu items are translatable after saving. Can you confirm that? Translate them before saving would not be possible.
 
Filters it's a whole other problem. Groups are translatable after saving, they are part of the `Filters` context, but single filters are not. They are the actual template names and do not exist on the `Filters` context. I could make them translatable but the user would have no way of knowing that what they are translating is the NAME of the template, not the name of the filter. I will remove translate from the filter UI for items, since I don't think it's a good idea that users translate tamplate names by mistake.

>     * It isn't clear to me what is the solution for [Difficult to determine editing row on larger screens #6431](https://github.com/huridocs/uwazi/issues/6431)

Rows highlighted as you hover over them ;)
> 
> 
> There are also a couple of things from previous developments that maybe can be easily fixed in this PR.
> 
>     * Filters action buttons have a different order than other pages
Fixed!

>     * Relationship Types show the translated value followed by the original name wrapped by parenthesis, this is different in other pages like templates

That's right, removed the translate there.

app/react/V2/Routes/Settings/Filters/FiltersTable.tsx
Zasa-san: Kinda yes. It was not a good idea to avoid using the revalidator.

app/react/V2/Routes/Settings/Filters/components/FiltersSidepanel.tsx
Zasa-san: Calling the function for uniqueID on the instantiation of the form was triggering a rerender loop. I had a hard time figuring out was was triggering the endless rerenders on the component and removing the uniqueID from the default values fixed it.

mfacar: I got it, it doesn't matter 👍

mfacar (approved):
LGTM! Really important improvements! 💯

aphilop:
Some questions/comments to check if they should be part of this PR:

- I am wondering why in the Filters table we don't show "Selected x of y" next to the Delete button when item(s) are selected?

image

- Same as above in the Uploads table. Also in this table the action buttons at the bottom are not at the left as in all other tables.

image

Zasa-san:
> Some questions/comments to check if they should be part of this PR:
> 
> * I am wondering why in the Filters table we don't show "Selected x of y" next to the Delete button when item(s) are selected?
> 
> 
> 
> * Same as above in the Uploads table. Also in this table the action buttons at the bottom are not at the left as in all other tables.
> 

@aphilop We have a few UI inconsistencies in the settings screen that are the result of iterating over the designs as we develop each screen.

I would suggest having an issue to list these inconsistencies, and to work on it after we finish all the settings screens (I think only templates is missing now). 

We have this issue here to get started: https://github.com/huridocs/uwazi/issues/6424

And then some minor issues that could be solved on a final pass on the settings UI:

https://github.com/huridocs/uwazi/issues/2032
https://github.com/huridocs/uwazi/issues/6837
https://github.com/huridocs/uwazi/issues/6757
https://github.com/huridocs/uwazi/issues/6425
https://github.com/huridocs/uwazi/issues/6435

Usually we don't find time to address these minor issues, so it could be a good thing to compile them into a big one and solve them after we have migrated all the settings screens.

Issue #6906, opened 2024-06-17
Thesauri - It is not possible to change the group of an item


salvalacruz:
Currently, it is not possible to change the group of an item in a thesaurus.

If an item is in a group, when you edit it from settings, the UI does not allow you to change the group it is in.

This functionality existed previously, and its loss is making troubles in the workflow of some instances.

Zasa-san:
This is something that we have explicitly blocked for thesauri. Is not a bug, but something we are preventing on purpose. We should review if this is correct.

salvalacruz:
Thank you @Zasa-san , I think it is not a good policy to remove a feature without warning users in advance.
There are use cases where thesauri have a certain dynamism, and an item can change its group depending on the context. 
The option of deleting the item and create it again in another group is not feasible or fluent, if this item is already linked to property values in the entities.

aphilop:
We include this as one of the issues relevant to V2 Thesauri (#6699) and we could revisit this when we would be able to tackle again the topic. At the moment, the priority is on IX and Automatic translations.

mfacar:
Although it was supported in the past (#2240 contains a reference for this functionality), approximately since #3726, without the restriction in the front end, the endpoint doesn't update the denormalized labels in the entities. We will keep restricting the change of groupings until a validation of the groups and a property update of the normalized data are implemented.

aphilop:
We should validate if work that has been done to move Thesauri to data layer V2 could potentially help fixing this and re-assess priority.

cc @RafaPolit @huridocs/backend

Issue #4053, opened 2021-11-25
Thesauri (dictionaries) is not generating _id for values


RafaPolit:
Thesauri are not generating `_id` for values, and relying on `id`, which should be deprecated and is only used in client to sort originally created values.

This needs to be reworked from the ground up.

LaszloKecskes:
Clarification: the problem is that the unique id is generated by the front-end for a different purpose, then the back-end accepts it and uses it for identification.

Joao-vi:
I believe now we do not use `_id` from Thesaurus **values** and always use the manual generated `id` 

```
    {
      _id: 'mongo generated',
      name: 'Thesaurus',
      values: [
        {
          _id: 'mongo generated', // Not used anymore.
          id: 'manual generated',
          label: 'Label',
        },
      ],
    }
```

`_id` is going to be removed on this PR https://github.com/huridocs/uwazi/issues/8679

@RafaPolit Can you confirm ?

aphilop:
This is already fixed

Issue #905, opened 2017-04-07
Models using _id and id should be using only _id


RafaPolit:
Previously couch subdocuments had no _id, so there is an id-generation process that stores an ID for each subdocument.

With Mongo, that is done automatically, so models and processes depending upon old ID should be migrated to just use _id.

txau:
@RafaPolit is this still a thing?

RafaPolit:
I believe it is, at least for template properties or thesauri items.  We need to check.

txau:
@RafaPolit you are right, I just checked and botth template properties and thesauri items still use id.

RafaPolit:
Thesauri, relationshiptypes (including route validation and model) still use id.  We should create issues to remove this.

fnocetti:
At this point we think the only place is the thesauri values.

PR #7132, opened 2024-08-20
6699 problems with v2 thesauri

### LINKED ISSUE #6676: Thesauri settings: order items

### LINKED ISSUE #6696: Empty groups on thesaurus
**Describe the bug**
When creating a new thesaurus, you are allowed to create an empty group. 

**To Reproduce**
1. Go to thesauri settings screen
2. Add thesaurus
3. Add group
4. Simply put a name to the group and hit Add
5. If there is no other items, it seems you can save the group, but then you cant edit it to add items. If the thesaurus has items, then it throws an error when trying to save.

**Expected behavior**
Validate that groups cannot be created empty, and when deleting items in a group remove the whole group if it's left with no items.

### LINKED ISSUE #6699: Problems with V2 thesauri
After #6698 we detect the following critical problems:

Thesauri needs better testing in general.

```[tasklist]
### UI issues
- [x] When creating a group the sidepanel save button should say 'Add group'
- [x] When editing a group the sidepanel save button should say 'Edit group'
- [x] Same for item create and edit ('Add' / 'Edit')
- [x] On the settings -> thesauri and on each individual thesaurus, the EDIT button has a disabled look (the border is not "deep blue") and hovering over them has no action.
- [ ] https://github.com/huridocs/uwazi/issues/7043
- [ ] https://github.com/huridocs/uwazi/issues/6696
- [ ] https://github.com/huridocs/uwazi/issues/6959
- [ ] https://github.com/huridocs/uwazi/issues/6676
- [ ] https://github.com/huridocs/uwazi/issues/6906
- [x] It is not possible to translate thersaurus names or values
```

```[tasklist]
### Validation
- [x] Please see #6696
- [x] You can delete groups and items and there is no warning as to the effect that this will have on the collection. We could add a warning before saving letting users now what items where removed and that that will have an effect on entities using those items.
```

```[tasklist]
### Testing
- [x] Unit tests for helpers
- [x] Redo E2E, it must check effects of changing thesauri (or do some of this in a unit test)
- [x] Adding thesauri
- [x] Adding groups
- [x] Editing thesauri
- [x] Editing group names and group item names
- [x] Deleting items
- [x] All of the above checked with library
```

```[tasklist]
### Code quality
- [x] There are `ts-ignore`s. Those ts warning are right, and instead of ignoring we should fix them
- [x] eslint problems. Be careful about having an environment that displays eslint errors for import sorting, arrow function styles, self closing tags, etc.
```

### LINKED ISSUE #6959: Problems when importing a thesaurus from CSV file
**Describe the bug**
The system throws an error if you create a thesaurus and immediately try to import a CSV file.

**To Reproduce**
1. Go to '/settings/thesauri'.
2. Click on 'Add thesurus'
3. Type a thesaurus name
4. Click on 'Import', select a CSV file properly done. 
[Thesaurus Test.csv](https://github.com/user-attachments/files/16054872/Thesaurus.Test.csv)
5. Then you get the list of items:
image
6. Click on 'Save'.
7. you get an error:
image
8. However, if you come back to '/settings/thesauri' and edit the just created thesaurus again, the imported list is available.

**Expected behavior**
Errors should not appear, if the CSV file is properly imported.

 - [X] Desktop
 - [ ] Mobile

**Browser**
Brave

### LINKED ISSUE #7043: Thesauri / All items are deleted when trying to remove a duplicated value
**Describe the bug**
When adding an item in a thesaurus, if this value already exists and then you must delete it, the whole list of values disappear.

**To Reproduce**
1. Go to 'Settings - Thesauri' and select a thesaurus.
image

2. Add an item with a duplicate value on purpose.
image

3. When trying to save it, the system send a warning.
image

4. Select the duplicated item and click on Remove.
image

5. See error: every item has de dissappeared.
image

**Expected behavior**
Only the duplicated and selected item should be removed.

mfacar:
fixes #6699
fixes #6696
fixes #7043
fixes #6959
fixes #6676

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


Zasa-san:
First round of functional review. This is focused solely on the thesauri screen:

- Long labels look odd with the parenthesis on the 'translated'. Super minor thing, I think we can just leave it like that until we have a global solution for multiline values on tables. But maybe it's a quick fix.

image

- When you are adding a group you cannot create it empty. Is this intended? Considering that now empty groups have the placeholder row to add items to it.

- You can drag all the items out of a group and still save the thesauri. This makes the group into a root item. Is this intended? Should the validation fail or the group get deleted? (If you leave the `subRows` property as an empty array it should still register it as a group on the table).

- Editing a group with values is odd, because it lists all the thesauri inside it in the sidepanel. Maybe when you edit a group just show the group label? We are not offering ways of deleting the items inside, or sorting them, so makes no sense to have them there.

- Edit a group -> Change the name of one of the items inside it, or even the label of the group itself -> Save and close the sidepanel -> Open the sidepanel again. It’s empty.

- Expan a group -> edit a child -> save and close the sidepanel -> the change is applied to the group label, not the child label.

- Generally check the intended behavior of dnd. In prod you currently cannot drag items in or out of a group.

Zasa-san:
Fiddling around with thessauri I see some inconsistencies in the `id` property of the database. Some of them have a long format `id`, other have what appears to be one of the client's ID generated codes. 

*Thesauri created/edited in this branch*
image

*Thesauri created/eited in development*
This is a thesauri created in development. The `id` property is consistent:

image

app/react/V2/Routes/Settings/Thesauri/ThesaurusForm.tsx
```diff
@@ -1,212 +1,70 @@
...
+        <Translate>Thesauri added.</Translate>
+      ),
     });
-    setThesaurusValues(thesaurusValuesCopy);
-  };
-
-  const addGroupSubmit = (group: FormThesauriValue) => {
-    const thesaurusValuesCopy = thesaurusValues.filter(item => item._id !== group._id);
-    setThesaurusValues([...thesaurusValuesCopy, group]);
+    navigate(`../edit/${savedThesaurus._id}`);
```
Zasa-san: Is this to reload the page? We have something similar here 
https://github.com/huridocs/uwazi/blob/cde642be4baffd54d9876c182d229ff08d58a900/app/react/V2/Routes/Settings/Pages/PageEditor.tsx#L86

The example also prevents a problems where if the user navigates back they would end up in the new thesaurus screen in some cases.

app/react/V2/Routes/Settings/Thesauri/components/TableComponents.tsx
```diff
@@ -1,46 +1,39 @@
...
-        color="indigo"
-        className="bg-indigo-200 border-none rounded-md drop-shadow-none"
-      >
-        <Translate>Group</Translate>
-      </EmbededButton>
-    )}
-  </div>
-);
+const ThesaurusValueLabel = ({ getValue, cell }: CellContext<ThesaurusValueSchema, string>) => {
+  const { thesaurus } = cell.getContext().column.columnDef.meta!.data;
```
Zasa-san, line 39: Is this info nowhere on the `row`?

app/react/V2/Routes/Settings/Thesauri/components/TableComponents.tsx
```diff
@@ -75,23 +68,30 @@ interface TableThesaurusValue extends ClientThesaurusValue, Omit<ClientThesaurus
...
-const columns = (actions: { edit: Function }) => [
+const columns = (actions: { edit?: Function }, thesaurus: ClientThesaurus) => [
   columnHelper.accessor('label', {
     id: 'label',
     header: LabelHeader,
     cell: ThesaurusValueLabel,
-    meta: { headerClassName: 'w-11/12' },
-  }) as ColumnDef<TableThesaurusValue, 'label'>,
+    meta: { headerClassName: 'w-11/12', data: { thesaurus } },
+  }) as ColumnDef<ThesaurusRow, 'label'>,
```
Zasa-san: The type should be inferred automatically. No need to cast it maybe?

app/react/V2/Routes/Settings/Thesauri/components/ThesauriTable.tsx
```diff
@@ -0,0 +1,77 @@
...
+  currentThesauri: ThesauriRow[];
+  setSelectedThesauri: React.Dispatch<React.SetStateAction<ThesauriRow[]>>;
+}
+
+const ThesauriTable = ({ currentThesauri, setSelectedThesauri }: ThesauriTableProps) => {
+  const navigate = useNavigate();
+  const navigateToEditThesaurus = (thesaurus: Row<ThesauriRow>) => {
+    navigate(`./edit/${thesaurus.original._id}`);
+  };
+  const columnHelper = createColumnHelper<any>();
```
Zasa-san: You can use the type here, and avoid casting in every column definition below

app/react/V2/Routes/Settings/Thesauri/specs/Thesauri.spec.tsx
```diff
@@ -0,0 +1,469 @@
...
+          loader: editThesaurusLoader({}),
+        },
+      ],
+      {
+        initialEntries: ['/'],
+      }
+    );
+
+    const renderComponent = () => {
+      const reduxStore = createMockStore([thunk])(() => ({
```
Zasa-san: If we incorporate this method of testing routes, the mock store default basic values should probably be extracted into the shared folder for V2 ``V2/shared`

app/react/V2/Routes/Settings/Thesauri/specs/Thesauri.spec.tsx
```diff
@@ -0,0 +1,469 @@
...
+                      label: 'Child 1',
+                    },
+                    {
+                      label: 'new child 2',
+                    },
+                  ],
+                },
+              ],
+            },
+            undefined
```
Zasa-san: Is this undefined in the expect correct?

app/react/V2/Routes/Settings/Thesauri/specs/Thesauri.spec.tsx
```diff
@@ -0,0 +1,469 @@
...
+          );
+        });
+      });
+    });
+
+    describe('Edit Thesaurus', () => {
+      it('should edit a root item', async () => {
```
Zasa-san, line 346: There are some tests that don't have any `expect`s. Is that correct? How do you check if the test fails?

app/react/V2/Routes/Settings/Thesauri/specs/fixtures.ts
```diff
@@ -0,0 +1,46 @@
+const thesauri = [
+  {
+    _id: 'thesaurus1',
+    name: 'Colors',
+    values: [{ _id: 't1_black', label: 'black', id: 'id1_black' }],
```
Zasa-san, line 5: I don't think thesaurus values have `_id`. Just mentioning in case some test relies on it being there.

Zasa-san (changes requested):
Lengthy review process, but it's looking great! 
I have some doubt's about test using `_id` in the fixtures for a thesaurus values. If there are cases where the client is inserting those `_id` in the values we should check to avoid putting them in rather than sanitizing them. (Unless I'm mistaken thesaurus values would not haven an `_id` property ever).

app/react/V2/Routes/Settings/Thesauri/components/TableComponents.tsx
mfacar: Totally!

app/react/V2/Routes/Settings/Thesauri/specs/Thesauri.spec.tsx
mfacar: It was expected because the headers are not passed to the calls of the API from the component, however, the methods are not called anywhere else so I removed the param in the API client.

app/react/V2/Routes/Settings/Thesauri/specs/Thesauri.spec.tsx
mfacar: It's a good one, I'm using screen.get... functions which fail in case the looken element doesn't exists in the document (a couple use screen.query... which doesn't fail if the element doesn't exist), also @testing-library/react/pure to keep the state between the tests, but adding expects definitively helps a lot with the readability of the tests, I added a function to check the right call of the save function where it applies.

app/react/V2/Routes/Settings/Thesauri/specs/fixtures.ts
mfacar: I'm adding the _id to ensure the values are kept through the sanitization functions, this because #905 & #4053  suggest there could be values with _id, the component is only handling id given that that currently that is the property generated by the endpoint.

mfacar:
Thank you @Zasa-san! 
> * Long labels look odd with the parenthesis on the 'translated'. Super minor thing, I think we can just leave it like that until we have a global solution for multiline values on tables. But maybe it's a quick fix.
- In the meantime, I added a border instead of the parenthesis.
> * When you are adding a group you cannot create it empty. Is this intended? Considering that now empty groups have the placeholder row to add items to it.
- In fact, an empty group could not be used as a valid option of a template property, a validation in the endpoint is also required, I created #7144 regarding to that.
> * You can drag all the items out of a group and still save the thesauri. This makes the group into a root item. Is this intended? Should the validation fail or the group get deleted? (If you leave the `subRows` property as an empty array it should still register it as a group on the table).
- I disabled the group edition due to #6906 and #7144.  By the moment if all the items of a group are deleted by selection, the group is also removed.
> * Editing a group with values is odd, because it lists all the thesauri inside it in the sidepanel. Maybe when you edit a group just show the group label? We are not offering ways of deleting the items inside, or sorting them, so makes no sense to have them there.
- I'm not sure about this, although it is different from other places where items can be deleted, you can also make a massive update of all the members without clicking each row. @juanmnl any thoughts? 
> * Edit a group -> Change the name of one of the items inside it, or even the label of the group itself -> Save and close the sidepanel -> Open the sidepanel again. It’s empty.
- Fixed
> * Expan a group -> edit a child -> save and close the sidepanel -> the change is applied to the group label, not the child label.
- OMG!, Thanks for this caught, it's fixed.

mfacar:
> Fiddling around with thessauri I see some inconsistencies in the `id` property of the database. Some of them have a long format `id`, other have what appears to be one of the client's ID-generated codes.

For the record, @Zasa-san found the different formats for id are caused by adding values into a thesaurus from the metadata editor.   There is some legacy code in app/react/Thesauri/actions/thesauriActions.js that should be removed when we rewrite the related components.

Update: It was possible to have a long identification format by delegating the generation of IDs to the endpoint.

app/react/V2/Routes/Settings/Thesauri/specs/Thesauri.spec.tsx
Zasa-san: Ok, if we are sure those headers are never passed, we can remove them.

app/react/V2/Routes/Settings/Thesauri/components/ThesauriTable.tsx
```diff
@@ -0,0 +1,77 @@
...
+  currentThesauri: ThesauriRow[];
+  setSelectedThesauri: React.Dispatch<React.SetStateAction<ThesauriRow[]>>;
+}
+
+const ThesauriTable = ({ currentThesauri, setSelectedThesauri }: ThesauriTableProps) => {
+  const navigate = useNavigate();
+  const navigateToEditThesaurus = (thesaurus: Row<ThesauriRow>) => {
+    navigate(`./edit/${thesaurus.original._id}`);
+  };
+  const columnHelper = createColumnHelper<ThesauriRow>();
```
Zasa-san: I believe that the columnHelper and the columns themselves can be moved outside the component, they don't use anything from the component (only `edit` and that's passed when the columns are built in :59). I'm not sure how much it impacts component render time to rebuild all of this on every rerender, but it's safe to declare it outside. 

You could even momize the call to the column builder function and have it depend on the edit parameter. But maybe that's no so necessary.

app/react/V2/Routes/Settings/Thesauri/specs/helpers.spec.ts
```diff
@@ -0,0 +1,464 @@
...
+  addSelection,
+  compareThesaurus,
+  removeItem,
+  sanitizeThesauri,
+  sanitizeThesaurusValues,
+  sortValues,
+  thesaurusAsRow,
+} from '../helpers';
+
+const thesaurusValues: {}[] = [
```
Zasa-san: Just an idea, feel free to ignore it: use the correct type here and if you have type issues ts-ignore them with a message that explains why those incorrect values are there. Like _id is there to check correct sanitization, or why is the empty value array. Could be helpful for the next dev that has to come here.

Also useful for sanitizedThesaurusValues so that we are sure it's the correct expected values.

app/react/V2/Routes/Settings/Thesauri/EditThesaurus.tsx
```diff
@@ -0,0 +1,220 @@
...
+          {selectedThesaurusValue.length === 0 && (
+            <div className="flex justify-between w-full">
+              <div className="flex gap-2">
+                <Button onClick={addItem}>
+                  <Translate>Add item</Translate>
+                </Button>
+                <Button styling="light" onClick={addGroup}>
+                  <Translate>Add group</Translate>
+                </Button>
+                <Button styling="light" onClick={sortValues(thesaurusValues, setThesaurusValues)}>
```
Zasa-san, line 162: The table has it's own `sortingFn` prop that can be used to pass a custom sorting function. Could this not be used instead of a custom sorting function? Just rely on the table to run the sorting by title when this button is clicked? `app/react/V2/Routes/Settings/IX/IXSuggestions.tsx:252` does something like it.

cypress/e2e/settings/thesauri.cy.ts
```diff
@@ -20,104 +25,212 @@ describe('Thesauri configuration', () => {
...
+    cy.contains('.item-document', 'País select').click();
+    cy.contains('.metadata-name-select', 'Colores Importados: Azul Importado');
+  });
+
+  it('should update the values in the entities', () => {
+    changeLanguage('English');
+    cy.get('.only-desktop a[aria-label="Settings"]').click();
     cy.contains('span', 'Thesauri').click();
-    cy.get('tbody tr:nth-of-type(1) input').should('be.disabled');
+    cy.contains('New Thesaurus')
```
Zasa-san: I think you can do a `cy.contains('tr', 'New Thesaurus').contains('button', 'Edit').click();` here

cypress/e2e/settings/thesauri.cy.ts
```diff
@@ -1,11 +1,16 @@
 /* eslint-disable max-statements */
```
Zasa-san, line 1: The e2e snapshots fail locally for me. If they are green on the CI, I'm ok. It's a bit difficult for me to check that the e2e covers all the expected flows correctly, although.

Zasa-san (approved):
Amazing!

The remaining comments are suggestions, but generally I'm ok with merging as is.

cypress/e2e/settings/thesauri.cy.ts
mfacar: Certainly, much simpler!

cypress/e2e/settings/thesauri.cy.ts
mfacar: I have the same problem locally, the problem is reported in Cypress and happens when using Chrome as the browser, at the moment running with Electron or the console is the alternative to check the tests.

app/react/V2/Routes/Settings/Thesauri/EditThesaurus.tsx
mfacar: @Zasa-san I agree with you, but my attempt to use the sortingFn to update the sortingState from the table client turned into a more complex option, for sure I'm missing some part of the configuration, so for now I'll keeping the external sort function

txau:
The problem with importing the CSV still persists. 

image

txau (approved):
LGTM

Issue #6763, opened 2024-05-10
Entities only index one document per language


konzz:
When indexing the full text we currently use the Mongo ID of the entity to index it, this means we can only index one document per language and we can not get full-text search results from any other document.

This issue might be side effect fixed by the implementation required to fix this other issue: #6749

PR #7394, opened 2024-10-21
6763/index-all-documents

Joao-vi:
This PR was created based on this issue https://github.com/huridocs/uwazi/issues/6763

## Current progress

There're two tests that I couldn't understand why it broke, so I don't have quite sure if I should update the test to cover the new behavior or it's the new implementation that really broke the code. I really need some help about this

Test failing
1. search › relationship aggregations › should return aggregations based on title of related entity
2. search › should perform a fullTextSearch on fullText and title

We are now indexing all documents for each entity, but the problem is that for every entity language we are replicating the file on elasticsearch, this might be a serious **memory** issue, for example:

If a tenant have 5 languages, one entity with 2 primary files, in practice this will generate 10 primary files. 
On mongodb, this same scenario will have only 2 primary files (files are not replicated, but reused, using relationship file<->entity, by each entity in different language)

### DIFF
--- a/app/api/search.v2/searchResponse.ts
+++ b/app/api/search.v2/searchResponse.ts
@@ -11,24 +11,55 @@ function getSnippetsForNonFullText(hit: ElasticHit<EntitySchema>) {
     : [];
 }
 
+type ElasticFullTextFile = {
+  filename: string;
+  language: string;
+  originalname: string;
+  text: string;
+};
+
+type Props = ElasticFullTextFile;
+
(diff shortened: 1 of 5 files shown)


Joao-vi:
After discussing possible implementations to tackle this duplication of data on elastic search, I'm closing this PR because the best approach to solve this problem is a revamp of the current architecture of entities and its languages, which will be treated as totally separated issue in the near future. 

@txau @daneryl @RafaPolit

PR #7451, opened 2024-11-18
v2 NeedAuthorization

Zasa-san:
fixes issue when trying to migrate and test v2 translation component on the client, since it uses NeedAuthorization which is connected to the store. This new component uses atoms.

Also, included in this PR, there is a new testing strategy for unit testing components with atoms.

v1 NeedAuthorization is left unchanged to avoid messy breaking changes.

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


mfacar (approved):
LGTM! Good catch 👏 

Just a minor comment regarding the location of app/react/V2/atoms/TestAtomStoreProvider.tsx that I'm leaving to your criteria; I think it could be better to have this file as part of a test folder and not between the production code.

Zasa-san:
> LGTM! Good catch 👏
> 
> Just a minor comment regarding the location of app/react/V2/atoms/TestAtomStoreProvider.tsx that I'm leaving to your criteria; I think it could be better to have this file as part of a test folder and not between the production code.

Good idea! I have moved all v2 testing utils into a folder, and updated some older testing methods we used with atoms

mfacar (approved):
LGTM! 👏

Issue #7439, opened 2024-11-12
Files can't be set to Ukrainian


gabriel-piles:
For the OCR feature to function correctly, files must be assigned a specific language. While the OCR service supports Ukrainian, Uwazi currently does not allow files to be properly set to the Ukrainian language.

When a file is set to Ukrainian, the database correctly stores it as "ukr," but the frontend incorrectly displays "other." This likely causes the OCR service to use the "other" label for processing decisions.

PR #7465, opened 2024-11-21
Add language support to the OCR manager

Joao-vi:
fixes https://github.com/huridocs/uwazi/issues/7439

It's not too clear in code, but the Uwazi context and OCR context are using  different language code:
Uwazi = ISO639_3
OCR = ISO639_1

We were using getLanguage that I believe it was used to convert Uwazi language codes to Elastic search languages codes, as Ukranian is not supported in Elastic search, it was retuning `null` and not executing the OCR service.

I've created LanguageCodeMapper that can convert between language code.

A better solution, it would be to make OCR models, properties more explicit  in code so we can make a well defined boundary that can tell what belongs to Uwazi and what belongs to OCR

With that well defined Models we can create a Anti-Corruption Layer and whenever we need to use some feature from OCR we use this Anti-Corruption Layer to get the models from OCR.


daneryl (approved):

daneryl (approved):
I think we are good to go with this one :+1:
the whole languages list and how we use them on every component is still very confusing, lets merge this one and try to fix that part on next issue ?

Joao-vi:
> I think we are good to go with this one 👍 the whole languages list and how we use them on every component is still very confusing, lets merge this one and try to fix that part on next issue ?

Yes, let me know if you want me to create a new issue addressing that. 

Update: new issue addressing that https://github.com/huridocs/uwazi/issues/7514

Issue #7514, opened 2024-12-02
Tech debt | Available languages on the system.


Joao-vi:
### Problem
Uwazi system needs to work with different languages code and is to clear on code that "language" method, exported by languagesList.ts file, is used to map one language code to another one.

This is how which language code is used: 
1. Files => ISO639_3
2. Entity => ISO639_1
3. OCR => ISO639_1

### Todo

- [x] Create a mapper class to be responsible for translate one language to another.
- [x] Remove 'franc' (franc is ISO639_3 code and also, "franc"  is not a language code).
- [x] Try to remove ElasticLegacyObject.
- [ ] Remove key from available languages (as key stands for ISO639_1)

Joao-vi:
I believe replacing 'key' for 'ISO639_1' is not going be doable because this will require to change in a lot of files, check image bellow. 

<img width="340" alt="image" src="https://github.com/user-attachments/assets/430520a9-9da8-472b-a351-c4194717e9e7">

PR #7516, opened 2024-12-02
refactor language list

Joao-vi:
Please, go to issue bellow to gather more information about this pull request.

[Issue](https://github.com/huridocs/uwazi/issues/7514)


app/api/migrations/migrations/32-add-file-field-to-connections/index.js
```diff
@@ -1,12 +1,12 @@
 /* eslint-disable no-await-in-loop */
-import languages from 'shared/languages';
+import { elasticLanguages } from 'shared/language';
```
RafaPolit: This kind of breaks the rule that migrations should not use internal code, as changing the codebase will change the behavior of migrations.  I understand that this error is not introduced by this PR, it was already there... I'm wondering if this refactor should correctly implement this migration?

app/api/migrations/migrations/32-add-file-field-to-connections/specs/32-add-file-field-to-connections.spec.js
```diff
@@ -36,12 +36,9 @@ describe('migration add-file-field-to-connections', () => {
   });
 
   it('should work with unsuported languages by elastic without errors', async () => {
-    fixtures.settings[0].languages = [
-      {
-        key: 'ja',
-        default: true,
-      },
-    ];
+    const fullyUnsupportedElasticLanguage = { key: 'kn', default: true };
```
RafaPolit: Could you please elaborate on why was the 'ja' language of the original test changed to 'kn' here?  Thanks.

app/react/App/Root.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import serialize from 'serialize-javascript';
 
-import { availableLanguages as languagesList } from 'shared/languagesList';
+import { availableLanguages as languagesList } from 'shared/language';
```
RafaPolit: Since we are already doing a refactor, maybe let's remove the `as` and simply use the `availableLanguages` variable?

app/react/Attachments/components/AttachmentForm.js
```diff
@@ -4,17 +4,17 @@ import { connect } from 'react-redux';
...
 import ShowIf from 'app/App/ShowIf';
 
 export class AttachmentForm extends Component {
   render() {
     const { model } = this.props;
     const validators = { originalname: { required: val => !!val && val.trim() !== '' } };
-    const languageOptions = Object.keys(elasticLanguages).map(key => ({
-      value: elasticLanguages[key].franc,
-      label: elasticLanguages[key].elastic,
+    const languageOptions = elasticLanguages.map(item => ({
```
RafaPolit: Our code base usually either uses a descriptive name of what the iterable `item` is, or, in case of very short and "obvious" maps, we use a shorthand letter, as every iterable thing is an "item".  So, either:
`elasticLanguages.map(language => ({`
or something similar... or maybe directly destructure in the function call?
`elasticLanguages.map(({ISO639_3, elastic}) => ({value: ISO639_3, label: elastic}))`

Something more semantic than `item`?

app/react/UI/Icon/Icon.js
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { availableLanguages as languagesList } from 'shared/languagesList';
+import { availableLanguages as languagesList } from 'shared/language';
```
RafaPolit: Again, maybe just remove the `as`?

app/shared/language/available-languages.ts
```diff
@@ -3,23 +3,12 @@ import { LanguageSchema } from 'shared/types/commonTypes';
...
-  { franc: string; elastic: string; ISO639_1: string | null }
->;
-
-type Language = {
-  label: string;
-  value: string;
-};
-
-const undefinedLanguage: LanguageSchema = {
+const OTHER_LANGUAGE_SCHEMA: LanguageSchema = {
```
RafaPolit: Why the UPPER_SNAKE_CASE naming here? Maybe just adhere to the usual schema naming convention `otherLanguageSchema`?

app/shared/language/available-languages.ts
RafaPolit, line 1: Please, use our file naming convention `availableLanguages.ts`, not hyphened.

app/shared/language/language-mapper.ts
RafaPolit, line 1: Please, use our file naming convention `languageMapper`, not hyphened.

app/shared/language/language-utils.ts
RafaPolit, line 1: Please, use our file naming convention `languageUtils`, not hyphened.

app/shared/language/language-utils.ts
```diff
@@ -0,0 +1,22 @@
+import { LanguageSchema } from 'shared/types/commonTypes';
+import { LanguageCode, availableLanguages, OTHER_LANGUAGE_SCHEMA } from './available-languages';
+
+class LanguageUtils {
+  static getLanguagesByCode(code: LanguageCode = 'elastic'): LanguageSchema[] {
+    return availableLanguages.filter(item => Boolean(item[code]));
+  }
+
+  private static uniqueValues(array: string[]) {
```
RafaPolit: Maybe just for readability move the private method to the top of the class definition?

app/shared/language/language-utils.ts
```diff
@@ -0,0 +1,22 @@
+import { LanguageSchema } from 'shared/types/commonTypes';
+import { LanguageCode, availableLanguages, OTHER_LANGUAGE_SCHEMA } from './available-languages';
+
+class LanguageUtils {
+  static getLanguagesByCode(code: LanguageCode = 'elastic'): LanguageSchema[] {
```
RafaPolit: If these are already LanguageUtils, I'm wondering if the methods should include the term `Language` or `Languages` still?  For example `LanugageUtils.getByCode` may be all that is needed.  Same as `LanguageUtils.getCodes` or `LanguageUtils.getSchema`.  Not sure about this though.  Mabye it's OK as is.

app/shared/language/language-mapper.ts
```diff
@@ -0,0 +1,14 @@
+import { availableLanguages, LanguageCode, OTHER_LANGUAGE_SCHEMA } from './available-languages';
+
+export class LanguageMapper {
+  private static getLanguageByCode(code: string, languageCode: LanguageCode) {
+    return availableLanguages.find(item => code === item[languageCode]);
+  }
+
+  static fromTo(code: string, from: LanguageCode, to: LanguageCode) {
```
RafaPolit: I'm not really sold on this mapper.  `LanguageMapper.fromTo('EN', 'ISO639_1', 'ISO639_3')` reads weird.

What if this is just a "getter" and you use whichever key you want after the get?  For example: 
`LanguageUtils.get('EN', 'ISO639_1').ISO639_3`

What is the advantage of the mapper? simply the default value?  Maybe that can be included into the getter as well?

Let's discuss this on the call tomorrow?

RafaPolit (changes requested):
This looks mostly OK.  I have some form and naming observations.  Mainly it's the mapper that I am not convinced about.  Let's review this further?

app/api/migrations/migrations/32-add-file-field-to-connections/index.js
Joao-vi: Agreed.

This migration needs to map language code used on settings collections to the language code used on files collections.

I wonder if I copy elasticLanguages and paste it inside this folder would solve this dependency to internal code

app/api/migrations/migrations/32-add-file-field-to-connections/specs/32-add-file-field-to-connections.spec.js
Joao-vi: Elastic search have full support to `cjk` language code and it seems this language code stands for Chinese, Japanese and Korean languages [reference](https://en.wikipedia.org/wiki/CJK_characters).

On available languages array, we need to include `ISO639_3`, `elastic`, `ISO639_1`, that's why `ja` "became" a fully supported language.

And also all the variants from Chinese and Korean.

```
 {
    label: 'Japanese',
    key: 'ja',
    ISO639_3: 'jpn',
    elastic: 'cjk',
    ISO639_1: 'ja',
    localized_label: '日本語',
    translationAvailable: false,
  }
```

app/react/App/Root.js
Joao-vi: Got it!

app/react/Attachments/components/AttachmentForm.js
Joao-vi: Got it!

app/react/UI/Icon/Icon.js
Joao-vi: Got it!

app/shared/language/available-languages.ts
Joao-vi: Got it!

app/shared/language/available-languages.ts
Joao-vi: Got it!

app/shared/language/language-mapper.ts
Joao-vi: Got it!

app/shared/language/language-utils.ts
Joao-vi: Got it!

app/shared/language/language-utils.ts
Joao-vi: Good point, since this is under the language context, clients consuming this methods will have enough idea just by seeing `LanguageUtils.getCodes`

Going to refactor it.

app/shared/language/language-mapper.ts
Joao-vi: Yes, we can I discuss this, but the main downside of this is if we change this we're going to need to change all clients consuming this new style on the codebase

PR #7525, opened 2024-12-04
Fix/160 invalid date format on import csv

Joao-vi:
This pull request was created based on this [issue] (https://github.com/huridocs/Internal-Issues/issues/160)

### Problem
We were not respecting the date format, defined at settings -> collections, to import Dates in the translated versions of the Entities.


app/api/csv/importEntity.ts
```diff
@@ -1,8 +1,10 @@
+/* eslint-disable no-await-in-loop */
+/* eslint-disable no-unreachable-loop */
+/* eslint-disable no-restricted-syntax */
```
RafaPolit: We really cannot put exceptions to es-lints like this.  The fact that this practice has happened before should not be followed these days.  So, please, remove this exceptions and lets fix the code to not have these eslint warnings:
To solve this, please create a reduce with an async callback.

Also, the no-unreachable-loop is not really solving any warning, so this was probably a previous version of the code which is no longer present, all the more reason not to include this exceptions that can later be left in the header of the file and we end up with code that is against our own guidelines.

Issue #7389, opened 2024-10-21
permissionsContext.setCommandContext(); dangerous in some contexts


daneryl:
`permissionsContext.setCommandContext()` in some scenarios can result in the creation of an entity with the id of the fake user `permissionsContext.setCommandContext();` sets saved into the permissions, not being a valid ObjectId, this will throw errors on future permissions operations.

A good first fix can be to properly validate at the DB level that we are saving an ObjectId, this will be a good indirect way to know that we are saving the wrong user, and most importantly we will have a quicker feedback, with the current implementation   the problem is not obvious until any permission operations is performed, like trying to public an entity, which can happen days after or weeks.

Malove98:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

PR #7605, opened 2025-01-21
Add user ID validation and custom error handling in ModelWithPermissions

Joao-vi:
fixes #[7389](https://github.com/huridocs/uwazi/issues/7389)


app/api/odm/ModelWithPermissions.ts
```diff
@@ -13,6 +13,7 @@ import {
   UwaziQueryOptions,
   EnforcedWithId,
 } from './model';
+import { ObjectId } from 'mongodb';
```
RafaPolit: Let's try to adhere to the hierarchy of imports:
- external modules
- absolute paths
- relative paths

app/api/odm/ModelWithPermissions.ts
```diff
@@ -127,20 +128,44 @@ const controlPermissionsData = <T>(
   return { ...data, permissions: undefined };
 };
 
+export class InvalidUserIdError extends Error {
```
RafaPolit, line 133: I'm unsure about the Error practices we defined.  Are we extending native JS Error or are we using our Uwazi Error?  I thought we agreed the latter, but maybe I misunderstood something?

app/api/odm/ModelWithPermissions.ts
```diff
@@ -127,20 +128,44 @@ const controlPermissionsData = <T>(
...
+    super('You are trying to persist a userId that is invalid');
+    this.name = 'InvalidUserIdError';
+  }
+}
+
 export class ModelWithPermissions<T> extends OdmModel<WithPermissions<T>> {
+  private static validateUser(user: DataType<UserSchema> | undefined) {
+    try {
+      if (typeof user === 'undefined') return;
+      ObjectId.createFromHexString(user?._id?.toString()!);
```
RafaPolit: Is it your IDE Typescript demanding all these type conditionals?  If you are already returning for user === undefined, what is the possibility of here `user` needing the `?` ? And toString() with a `!`.  If you have _id and it is of the right class, how can toString() still require the `!`?

app/api/entities/specs/entitySavingManager.spec.ts
```diff
@@ -80,11 +83,12 @@ describe('entitySavingManager', () => {
 
     describe('new entity', () => {
       it('should create an entity without attachments', async () => {
+        const mockedUser = userInContextMock.mockEditorUser();
         const entity = { title: 'newEntity', template: template1Id };
         const { entity: savedEntity } = await saveEntity(entity, { ...reqData });
 
         expect(savedEntity.permissions).toEqual([
-          { level: 'write', refId: 'userId', type: 'user' },
+          { level: 'write', refId: mockedUser._id?.toString(), type: 'user' },
```
RafaPolit: Under what condition can the _id of mockedUser._id can be undefined?  I'm concerned about the IDE configuration and typing throughout these commit.

app/api/odm/specs/ModelWithPermissions.spec.ts
```diff
@@ -18,6 +18,7 @@ describe('ModelWithPermissions', () => {
     permissions: { type: mongoose.Schema.Types.Mixed, select: false },
     fixed: Boolean,
   });
+  const userId = testingDB.id();
```
RafaPolit: This file used `user1` and `user2` (in line 67 for example).

Probably these are not being used in the test, but maybe it's worth it to create a:
```
const user1Id = testingDB.id();
const user2Id = testingDB.id();
```

and replicate the original structure but removing the deprecated practice of assigning strings to the user?

RafaPolit (changes requested):
Added some minor comments.  There are some e2e failing, please rerun the failed tests until they are green.  Thanks.

app/api/odm/ModelWithPermissions.ts
Joao-vi: I see, is there any code example of this so I can replicate same behavior here ?

PR #7645, opened 2025-02-09
refactor(library): simplify document handling in entity save process

daneryl:
Prior implementation was converting blobs unnecessarily when originalFile was already available.


mfacar (approved):
LGTM! Nice finding :clap: 

As a comment apart, there are other places where this simplification is feasible, I'll take a look.

PR #7770, opened 2025-03-12
Remove OneUp Review and Suggested Metadata

RafaPolit:
fixes https://github.com/huridocs/Internal-Issues/issues/256

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


RafaPolit:
I have removed the creation of __[property name] from the query builder and all implementations that used the `includeReviewAggregations` flag.

There is however still a `.suggestedMetadata` property in the entities that, as far as I can tell, is bound to all this functionality as well and has no other impact, but that is rooted into several areas of the code.  I'd like to talk to the backend team about this to see if we remove this as part of this issue or create a new one.

Issue #7159, opened 2024-08-28
[IX] Adjust the review panel to display the relevant text field when the selected property is select, multi-select or relationship


aphilop:
**Is your feature request related to a problem? Please describe.**
The information extraction feature currently works with PDF documents as source.
We want to expand the sources also to text fields within a template so that information can also be extracted from these text fields.

**Describe the solution you'd like**
When the source of an extractor is defined as text field, then the action name for opening the review panel in the table should be adjusted accordingly. Then the review panel should display on the top part the relevant text field while the lower part remains the same as we have it now.

@juanma will add the discussed designs and description of the flow for this enhancement.

juanmnl:
The selected source of the entity is added as a block in the drawer (similar as we have the PDF now), using the multi-select and select pattern in the hub.

<img width="700" alt="hub - multi select - edit" src="https://github.com/user-attachments/assets/0fb76371-3f9c-423b-9a8c-6a0daa447064">

**[Design](https://www.figma.com/design/Ypgf1qGtcxgMJEgWchoqTx/WIP?node-id=6896-5403&t=XgM4MPyaNOKIDXY0-11)**

Issue #7158, opened 2024-08-28
[IX] Adjust the review panel to display the relevant text field when the selected property is text, numeric or date


aphilop:
**Is your feature request related to a problem? Please describe.**
The information extraction feature currently works with PDF documents as source.
We want to expand the sources also to text fields within a template so that information can also be extracted from these text fields.

**Describe the solution you'd like**
When the source of an extractor is defined as text field, then the action name for opening the review panel in the table should be adjusted accordingly. Then the review panel should display on the top part the relevant text field while the lower part remains the same as we have it now.

@juanma will add the discussed designs and description of the flow for this enhancement.

juanmnl:
The selected source of the entity is added as a block in the drawer (similar as we have the PDF now), retaining the click to fill pattern in the hub.

<img width="700" alt="hub -click to fill - edit" src="https://github.com/user-attachments/assets/5757b413-97a6-4fa3-878b-5bd24436a58d">

**[Design](https://www.figma.com/design/Ypgf1qGtcxgMJEgWchoqTx/WIP?node-id=6851-6223&t=XgM4MPyaNOKIDXY0-11)**

RafaPolit:
The click-to-fill in text has multiple nuances and it can be affected by how the service segments the text, so this may not be possible to fully implement.  Still, the functionality of assigning a value by selecting is still useful, so there are three possible stages to achieve this:
- Do not implement click-to-fill at all, just allow input of data
- Implement click-to-fill as a "copy paste" functionality where you select text, select click-to-fill and this puts the value in the input.  This would be consistent with other processes, even if under the hood it has no effect larger than just typing in the value
- Create the full text-to-fill process where we store context and location (will not work with absolute positioning as it is done in PDF) and more info.

juanmnl:
We are also "standarizing" the row action that triggers the drawer from "Open PDF" to just "Open", everywhere.

<img width="543" alt="Screenshot 2024-09-11 at 8 47 03 AM" src="https://github.com/user-attachments/assets/16e37d0e-ed3a-4f5a-a14f-e0ca37ebdaaf">

aphilop:
As discussed in the product call (2025-03-17):
- Implement click-to-fill as a "copy paste" functionality where you select text, select click-to-fill and this puts the value in the input. This would be consistent with other processes, even if under the hood it has no effect larger than just typing in the value

PR #7788, opened 2025-03-17
7157 - Support text fields as sources

### LINKED ISSUE #7157: [IX] Support text fields as source when creating an extractor
The information extraction feature currently works with PDF documents as source.
We want to expand the sources also to text fields within a template so that information can also be extracted from these text fields.

The dialogue of creating a new extractor should be expanded to display available text fields as sources for the information extraction. An extractor can have as source either PDF documents or text fields.

@juanmnl will add the discussed designs and description of the flow for this enhancement.

Zasa-san:
fixes #7157 #7159 #7158 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/services/informationextraction/ixextractors.ts
```diff
@@ -106,32 +109,36 @@ export const Extractors = {
     await model.delete({ _id: { $in: ids } });
     await Suggestions.delete({ extractorId: { $in: ids } });
   },
-  create: async (name: string, property: string, templateIds: string[]) => {
+  create: async (extractor: Omit<IXExtractorType, '_id'>) => {
```
RafaPolit: There something inconsistent regarding IDs as string arrays or ObjectId arrays. Here, the create is being passed an "IXExtractorType" and you assume templates is an array of object ids, but in the test you are passing an array of strings. Since the request is coming from the route directly, the test is correct in passing an array of strings, but the implementation here of the type is not. And this requires you to call toString further down the path several times because this is wrongly typed here.

app/api/services/informationextraction/ixextractors.ts
```diff
@@ -106,32 +109,36 @@ export const Extractors = {
...
       property,
       templates: templateIds,
     });
     await createBlankSuggestionsForExtractor(saved);
     return saved;
   },
-  update: async (id: string, name: string, property: string, templateIds: string[]) => {
-    const [extractor] = await model.get({ _id: new ObjectId(id) });
-    if (!extractor) throw Error('Missing extractor.');
+  update: async (extractor: IXExtractorType) => {
```
RafaPolit: Same here with the types.

app/api/services/informationextraction/specs/ixextractors.spec.ts
```diff
@@ -179,7 +185,12 @@ describe('ixextractors', () => {
 
   describe('create()', () => {
     it('should create a new ixextractor', async () => {
-      await Extractors.create('age_test', 'age', [fixtureFactory.id('personTemplate').toString()]);
+      await Extractors.create({
+        name: 'age_test',
+        property: 'age',
+        source: { pdf: true },
+        templates: [fixtureFactory.id('personTemplate').toString()],
```
RafaPolit, line 192: Here the test is correctly calling the create method with strings of ids.

app/api/suggestions/blankSuggestions.ts
```diff
@@ -75,25 +120,58 @@ export const createBlankSuggestionsForPartialExtractor = async (
...
+                file,
+                extractorId: extractor._id,
+                template,
+                propertyName: extractor.property,
+                propertyType: exampleProperty.type,
+                defaultLanguage,
+              })
+            )
+        );
+      } else if (tenants.current().featureFlags?.ixExtraSources) {
```
RafaPolit: Not sure I understand this conditional.  As I see it, if the source is PDF it should get a blanksSuggestionsForPdf, if the source is a property, it should do the blank suggestion for property.  The fact that the feature flag is activated could be part of the conditional, but it could not be the only conditional, right? There should be an additional trigger to this code than only the feature flag enabled? Or are you assuming that if PDF is false, then you will just execute this else and, as soon as we integrate the feature, this will be a simple else?

app/api/suggestions/blankSuggestions.ts
```diff
@@ -75,25 +120,58 @@ export const createBlankSuggestionsForPartialExtractor = async (
...
+        const entityData = await fetchEntitiesData(template, undefined, batchSize);
+
+        suggestionsToSave.push(
+          ...entityData.map(entity =>
+            getBlankSuggestionForProperty({
+              entityId: entity.sharedId,
+              extractorId: extractor._id,
+              template,
+              propertyName: extractor.property,
+              propertyType: exampleProperty.type,
```
RafaPolit: This is not from you, but the naming "exampleProperty" is not really ideal, right?  Should it be `sampleProperty`?

RafaPolit (changes requested):

app/api/services/informationextraction/ixextractors.ts
Zasa-san: I'll create a type that extends IXExtractorType so that we can cast templates as a string[].

app/api/suggestions/blankSuggestions.ts
Zasa-san: the new property `source` is not optional, and if you are not defining it as `pdf: true` then the other option is that it's a property. But for safety I'll add a check that the `source` is defined with a `property`.

Zasa-san:
Closed in favor of https://github.com/huridocs/uwazi/pull/7815#pullrequestreview-2717915759

PR #7815, opened 2025-03-26
7157 - IX text field support

### LINKED ISSUE #7157: [IX] Support text fields as source when creating an extractor

Zasa-san:
fixes #7157 #7158 #7159 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/services/informationextraction/ixextractors.ts
```diff
@@ -19,6 +19,13 @@ type AllowedPropertyTypes =
     >
   | 'title';
 
+interface ExtractorType extends IXExtractorType {
+  templates: string[];
+}
+interface NewExtractorType extends Omit<IXExtractorType, '_id'> {
+  templates: string[];
+}
```
RafaPolit, line 27: I think this can just be `type NewExtractorType = Omit<ExtractorType, '_id'>;` so you don't repeat the templates definition.

app/api/suggestions/blankSuggestions.ts
```diff
@@ -2,52 +2,97 @@ import entitiesModel from 'api/entities/entitiesModel';
...
+  entityId: string;
+  extractorId: ObjectIdSchema;
+  propertyName: string;
+  propertyType: string;
+  template: ObjectIdSchema;
+  language: string;
+}) => ({
+  language,
+  entityId,
+  entityTemplate: typeof template === 'string' ? template : template.toString(),
```
RafaPolit: Again I'm confused by this conditional.  There is a typing on the arguments for template.  If `template` is of type `ObjectIdSchema`, this conditional should not exist.  Is this because the test is passing the wrong thing or can this function really receive both types?  If the former scenario, we should fix the test, if the later, then we either dual-type the property (not ideal) or we ensure that we are passing the right type from all the calls to the function.

I see this is also the case in line 91 and 101 which is unfortunate. :(

app/api/services/informationextraction/ixextractors.ts
Zasa-san: That is because we want to avoid inheriting templates as  ObjectIdSchema[], and then having to do .toString() on every access to it.

app/api/suggestions/blankSuggestions.ts
Zasa-san: This was originally as such, but now that we are casting templates as string[], and knowing that those come from the routes (and are therefore strings) we can remove that.

PR #7912, opened 2025-04-16
Support for elasticsearch auth via api key

daneryl:
Also now scripts like reindex now use elastic client instead of fetch

### DIFF
--- a/app/api/search/elastic.ts
+++ b/app/api/search/elastic.ts
@@ -1,13 +1,14 @@
 import elasticSearch, { RequestParams } from '@elastic/elasticsearch';
+import { IndicesPutSettings } from '@elastic/elasticsearch/api/requestParams';
 import {
-  TransportRequestOptions,
   RequestBody,
   RequestNDBody,
+  TransportRequestOptions,
 } from '@elastic/elasticsearch/lib/Transport';
-import { tenants } from 'api/tenants';
 import { config } from 'api/config';
+import { tenants } from 'api/tenants';
 import { EntitySchema } from 'shared/types/entityType';
-import { SearchResponse, IndicesPutMapping, IndicesDelete, IndicesCreate } from './elasticTypes';
+import { IndicesCreate, IndicesDelete, IndicesPutMapping, SearchResponse } from './elasticTypes';
 
 const elasticClient = new elasticSearch.Client({
   nodes: config.elasticsearch.nodes,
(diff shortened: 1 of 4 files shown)


app/api/search/elastic.ts
```diff
@@ -38,6 +39,16 @@ const elastic = {
   },
 
   indices: {
+    async putSettings(params: IndicesPutMapping, options?: TransportRequestOptions) {
```
RafaPolit: This is great! But IndicesPutMapping is probably the wrong type here.  The `requestParams.d.ts` has separated definitions:
```
export interface IndicesPutMapping<T = RequestBody> extends Generic {
  index?: string | string[];
  type?: string;
  include_type_name?: boolean;
  timeout?: string;
  master_timeout?: string;
  ignore_unavailable?: boolean;
  allow_no_indices?: boolean;
  expand_wildcards?: 'open' | 'closed' | 'hidden' | 'none' | 'all';
  write_index_only?: boolean;
  body: T;
}

export interface IndicesPutSettings<T = RequestBody> extends Generic {
  index?: string | string[];
  master_timeout?: string;
  timeout?: string;
  preserve_existing?: boolean;
  ignore_unavailable?: boolean;
  allow_no_indices?: boolean;
  expand_wildcards?: 'open' | 'closed' | 'hidden' | 'none' | 'all';
  flat_settings?: boolean;
  body: T;
}
```

Maybe it's worth it to type properly? Since the files that use this are .js this is going unnoticed, but I think it's worth exporting in our `elasticTypes` the right type?

database/reindex_elastic.js
```diff
@@ -91,12 +70,10 @@ const indexEntities = async () => {
...
   process.stdout.write(`Deleting index ${tenants.current().indexName}...`);
   try {
-    await fetch(getIndexUrl(), { method: 'delete' });
+    await elastic.indices.delete({});
   } catch (err) {
-    // Should not stop on index_not_found_exception
-    if (err.json.error.type === 'index_not_found_exception') {
-      process.stdout.write('\r\nThe index was not found:\r\n');
-      process.stdout.write(`${JSON.stringify(err, null, ' ')}\r\nMoving on.\r\n`);
+    if (err?.meta?.body?.error?.type === 'index_not_found_exception') {
```
RafaPolit, line 75: How do you get to use TS '?' operators in a JS file?

RafaPolit (changes requested):

database/reindex_elastic.js
daneryl: Optional chaining operator is a JS operator introduced in ecma2020 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

app/api/search/elastic.ts
daneryl: fixed :+1:

PR #7928, opened 2025-04-23
Migration, added indices to mongo px collections

daneryl:


RafaPolit:
I think this PR is perfect as is.  I have a concern that has nothing to do with this PR: mongoose actually checked for the existence of these indexes and created them if they didn't exist.  This process seems less 'versatile' and the indexes could be manually deleted and there will be no info regarding this loss in performance.  But maybe it's a non-issue and we can simply assume that we are not deleting those indexes.

daneryl:
> I think this PR is perfect as is. I have a concern that has nothing to do with this PR: mongoose actually checked for the existence of these indexes and created them if they didn't exist. This process seems less 'versatile' and the indexes could be manually deleted and there will be no info regarding this loss in performance. But maybe it's a non-issue and we can simply assume that we are not deleting those indexes.

The issue with this check as i see it is the performance hit, take into account that this is checked on every request for each db collection, i think migrations is a perfect way of doing this, it gives us more control, our main issue is that migrations right now are expensive to deploy, we need to solve that, or either design a system only for index creation ? nice topic to discuss on the backend team :+1:

PR #7986, opened 2025-05-12
Fix/7981-create-migration-script-to-detect-mimetype-on-files-collection

### LINKED ISSUE #7981: Create migration script to detect and assign mimetype for files without mimetype
Create a script to get all files that does not have a Mime-type  defined.

Try to detect MIME-type from the original name.

Fallback to `application/octet-stream` if detection  is not possible.

This current Issue should solve the following Issues: 
#7529 
#4942

Joao-vi:
fixes #7981


app/api/migrations/migrations/172-files_detect_and_assign_mimetype/arrayUtils.ts
```diff
@@ -0,0 +1,15 @@
...
+    await array.reduce(async (promise, item, index) => {
+      await promise;
+
+      return callback(item, index, array);
+    }, Promise.resolve());
+  }
+
+  static async parallelFor<Item>(array: Item[], callback: Callback<Item>): Promise<void> {
+    await Promise.all(array.map(async (item, index) => callback(item, index, array)));
+  }
```
RafaPolit: This method is not used by any part of the migration, right?  If so, maybe remove it?

Joao-vi: Yes! Good point, going to remove it

PR #8108, opened 2025-06-12
8051 - error handling in IX process results

### LINKED ISSUE #8051: [IX] Back-end is not handling failing tasks
Back-end is not handling in case when a Task returns with status `success` false.

You can go to `InformationExtraction.ts` file, check the method `processResults` and see that there's no statements to handle `success` false. 

When something wrong happens when the Task is processed by the external service and the Task is given to us, since we don't handle it, we are leaving `status` or `findingSuggestions`  on process for ever. 

When we start to train the model we set the flag `mode.findingSuggestions` to true.

When we start the process to generate suggestions we also have a flag, we set `suggestion.status` to `processing`.

So we need to handle the status in case the Task comes back with an error in both cases.

mfacar:
fixes #8051

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


mfacar:
Closing in favor of #8115

PR #8194, opened 2025-07-02
8150 Im backend now

### LINKED ISSUE #8150: [IX] Total of number suggestions on the table is incorrect
When filtering suggestions we are displaying the full total of suggestions as if no filter were applied. The 'total' should be of suggestion in the table, not the full total.

Image

konzz:
fixes #8150

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


RafaPolit (approved):
Great! So we were already calculating it, only not returning it? Nice.

PR #8207, opened 2025-07-08
updated settings layout

### LINKED ISSUE #8134: Migrate Settings layout to V2
After the final section of settings (Templates) is merged we can update the [settings layout](https://github.com/huridocs/uwazi/blob/development/app/react/Settings/Settings.js) to V2.

- Remove unnecesary use of redux in this area.
- Clean dependencies,
- Put all the section under the tailwind css scope.

konzz:
fixes #8134

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/react/I18N/components/I18NLink.tsx
```diff
@@ -12,6 +12,7 @@ type I18NLinkProps = {
   mainContext: { confirm: Function };
```
Zasa-san, line 12: FYI, we have a v2 Link component here: `app/react/I18N/I18NLinkV2.tsx`. 
Maybe it can be easily used in settings to avoid v1 deps?

app/api/config.ts
```diff
@@ -1,7 +1,10 @@
 import uniqueID from 'shared/uniqueID';
+import dotenv from 'dotenv';
 import { Tenant } from './tenants/tenantContext';
 import { version } from '../../package.json';
 
+dotenv.config();
```
Zasa-san, line 6: what are we doing with this?

app/react/Routes.tsx
```diff
@@ -7,7 +7,10 @@ import LibraryRoot from 'app/Library/Library';
 import { LibraryMap } from 'app/Library/LibraryMap';
 import { LibraryCards } from 'app/Library/LibraryCards';
 import { LibraryTable } from 'app/Library/LibraryTable';
-import { PreserveSettings, Settings } from 'app/Settings';
+//import { PreserveSettings } from 'app/Settings';
```
Zasa-san: Remove?

app/react/V2/Routes/Settings/Preserve/Preserve.tsx
```diff
@@ -0,0 +1,90 @@
+import React, { FormEvent, useEffect, useState } from 'react';
```
Zasa-san, line 1: We should have a unit test for this, or at least create a task issue to have it.

app/react/V2/Routes/Settings/Settings.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useMemo } from 'react';
```
Zasa-san, line 1: I think this route too warrants a test. Again, maybe we can have a task issue for it.

app/react/V2/api/preserve/index.ts
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
Zasa-san: Cleanup?

Zasa-san (commented):
Functionality looks good! Nice!

app/api/config.ts
konzz: importing the .env file if present

app/react/I18N/components/I18NLink.tsx
konzz: I've replaced every V2 link with that one

salvalacruz:
Hi, just two small suggestions about the new layout to add properties, to make it more friendly:

- In the selection list to choose the property type, It would be nice to recover the icons of each type.
- Once you add a new property, the new layout sends this new item to the bottom of the template, and then you must move it to the desired order. Previously users may drag and drop the new property straight to the right place, that made work faster.

Thank you

PR #8225, opened 2025-07-14
Process Selected

### LINKED ISSUE #8129: [IX] Process selected [Backend]
At the moment we are not supporting just running the extraction process on selected entities. That is, when a new entity is added, the only means for processing the entity is by running the extraction on all the entities.

----

### Problem statement
Currently, the extraction pipeline only allows processing the entire dataset of entities at once. This means:
- When a new entity is added, users must re-run extraction on ALL entities
- There is no way to selectively process only specific entities of interest
- Resources are wasted processing already-processed entities
- Turnaround time for processing new additions is unnecessarily long
- Users lack the ability to prioritize certain entities for processing

### Proposed solution
Implement a feature that allows users to select specific entities for extraction processing, enabling:
1. Targeted processing of any subset of entities the user chooses
2. More efficient use of computational resources

### Acceptance criteria

- [x] When entities are selected and "Find suggestions" is clicked, the system should:
   - [x] Use the existing trained model (no retraining)
   - [x] Process only the selected entities
   - [x] Update the results for just the selected entities
   - [ ] Maintain all previous results for unprocessed entities
- [x] Provide meaningful feedback during processing (e.g. with count/percentage of selected entities completed)
- [ ] Upon completion, display results to the user
- [ ] Both "Find suggestions" and "Accept" buttons indicate with a number on how many of the selected entities would be applied (eg. the "Accept" button has no effect on entities with already accepted suggestions or no suggestions at all). 
- [ ] Add filter option for entries "Without suggestions"
- [ ] Add "New" pill next to buttons when there are new unprocessed entities
- [ ] Add a "Suggest" button to the each new unprocessed entity so the user can get suggestions for each entity individually

### General considerations
- This feature complements the "Test Run" feature but serves a different purpose
- The extraction pipeline needs to be modified to:
  - Accept a list of specific entities to process
  - Skip the training phase when using an existing model
  - Merge new results with existing results

RafaPolit:
fixes #8129

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/services/informationextraction/InformationExtraction.ts
+++ b/app/api/services/informationextraction/InformationExtraction.ts
@@ -47,7 +47,7 @@ import { DefaultDispatcher } from 'api/queue.v2/configuration/factories';
 import { retryWithBackoff, descriptiveError } from 'api/utils/retryWithBackoff';
 import ixmodels from './ixmodels';
 import { IXModelsModel } from './IXModelsModel';
-import { Extractors } from './ixextractors';
+import { Extractors, ModelNotReadyError } from './ixextractors';
 import {
   CommonSuggestion,
   RawSuggestion,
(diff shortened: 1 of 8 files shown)


RafaPolit:
Did a v2 from scratch.  Closing this one.

PR #8235, opened 2025-07-18
The branch name says it all

konzz:
- Improved CI e2e runs for about  ~3min
- Failed suits reruns are also faster
- Added cache strategy for libraries like mongosh, mongorestore and poppler
- Migration specs run separatedly from api_test to avoid runing them in reruns


.github/workflows/ci_build.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: install dependencies
+        run: yarn install
+      - name: build production
+        run: yarn production-build
+      - name: Upload production build artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: prod
+          path: |
+            ./prod/dist
```
Zasa-san, line 51: You should be able to replace this with `./prod`. I think we always include everything in that folder.

.github/workflows/ci_unit_tests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           - 9200/tcp
         options: -e="discovery.type=single-node" --health-cmd="curl http://localhost:9200/_cluster/health" --health-interval=10s --health-timeout=5s --health-retries=10
       minio:
-        # fixme: let's not depend on external unofficial image
+        # fixme: let's not depend on external unofficial image. #updated coment: ok!
```
Zasa-san, line 23: I think we still are?

scripts/updateTranslationsCSV - Copy.mjs:Zone.Identifier
Zasa-san, line 1: let's just delete this file, no?

Zasa-san (approved):
This is amazing! 

We should probably create an issue to re-instate the run condition only for PRs that are marked as ready. We could also include having a single place to mark the dependency versions (for example the elasticsearch docker image) to avoid having to replace it in multiple files.

scripts/updateTranslationsCSV - Copy.mjs:Zone.Identifier
konzz: That's what is happening here, I removed it

.github/workflows/ci_unit_tests.yml
konzz: Yes we still are, the "ok!" is just my agreement to the previous comment

.github/workflows/ci_build.yml
konzz: I did that intentionally, because it included the node_modules folder, making it an extra 750 MB to upload and then download. However, the node_modules are being cached, which is way faster than the artefact system.

PR #8309, opened 2025-08-04
8278 ix duplicated requests

### LINKED ISSUE #8278: [IX] Duplicated requests on Suggestions table
**Describe the bug**
This bug happens on Suggestions table, so you need to create an Extractor first and then click the `review` button to access the table of Suggestions. 

The following user interactions `change page`, `sort by table column`, `filter by status` are triggering duplicated requests and only refreshing the client-side state when all duplicated request finishes bringing a slow user experience. 

I've attached a screenshot of the duplicated request down bellow. 

Also, when paginating I believe we are also making unnecessary requests to unrelated resources, for instance, when I change the current page, the following resource are called `/api/ixextractors`, `/api/suggestions/aggregation`, `/api/suggestions/status` and for last `/api/templates`, which I believe the only one that should be called is `/api/suggestions?filter` if I'm not wrong.

I wonder if is it really required to also call these endpoint on current page change. Same behavior is seen for sorting by table column and also filtering by status

**To Reproduce**
1. Create an Extractor (it does not matter the configuration)
2. Click to `review` Extractor
3. Open devtools (f12) go to network tab.
4. Change page, see duplicated requests on network tab.

**Expected behavior**
Prevent from client-side make duplicated requests when changing the current page, sorting and filtering. 
Assess if is it really necessary to make API calls to unrelated resource, if not prevent the client-side from query unnecessary resources.

**Screenshots**
Image

konzz:
fixes #8278 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


Zasa-san:
covered by https://github.com/huridocs/uwazi/pull/8327

PR #8352, opened 2025-08-20
8245 fix aggregations add accuracy

### LINKED ISSUE #7414: [IX] Accuracy
Accuracy should be defined as:
Match / Mismatch, excluding those mismatches where the entity data is blank or empty string (not intentionally empty)

### LINKED ISSUE #8245: [IX] - Bring some order to filters
We could help the users understand how to better interpret the filters panel.

We could swap the position of the box containing "Obsolete" and "Error", with the box "Match" and "Mismatch". Also add some extra titles. So instead of like this:

Image

It reads like this:

ALL DATA
Labeled ...................... X
Non-labeled .............. X

Non processed............X
Obsolete .................... X
Error ........................... X

PROCESSED
Match ......................... X
Mismatch ................... X
No context ................. X

No context filter is similar to mismatch, but it also filters by those that have no context. This is important since a very crucial step of labeling the right data is locating those that are returning as "no context" and labeling them.

RafaPolit:
fixes #8245
fixes #7414

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/suggestions/specs/getSuggestionsForTableQuery.spec.ts
+++ b/app/api/suggestions/specs/getSuggestionsForTableQuery.spec.ts
@@ -464,6 +464,71 @@ const fixtures: DBFixture = {
 
       entityTitle: 'extractor_source_text_target_text_entity_5',
     }),
+
+    // Extra nonProcessed suggestions for multiselect extractor (date: null)
+    factory.ixSuggestion({
+      extractorId: factory.id('extractor_source_pdf_target_multiselect'),
+      entityId: 'extractor_source_pdf_target_multiselect_entity_1',
+      entityTemplate: 'extractor_source_pdf_target_multiselect_template',
+      propertyName: 'target_multiselect',
+      language: 'en',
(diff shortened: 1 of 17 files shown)


app/api/suggestions/specs/getSuggestionsForTableQuery.spec.ts
```diff
@@ -907,8 +972,8 @@ describe('getSuggestionsForTableQuery', () => {
     expect(nonProcessedResults).toHaveProperty('suggestions');
 
     // Test that both filters return the expected counts based on test data
-    expect(allResults.total).toBe(10);
-    expect(nonProcessedResults.total).toBe(10);
```
RafaPolit, line 911: This test is really not proving a subset, and no fixture has date: null, so this is passing due to the AND characteristic, which was wrong in the first place.

PR #8353, opened 2025-08-21
7414 - IX better stats display

### LINKED ISSUE #7414: [IX] Accuracy

Zasa-san:
fixes #7414

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/V2/Routes/Settings/IX/components/FiltersSidepanel.tsx
+++ b/app/react/V2/Routes/Settings/IX/components/FiltersSidepanel.tsx
@@ -2,19 +2,27 @@
 /* eslint-disable max-statements */
 /* eslint-disable react/jsx-props-no-spreading */
 import React from 'react';
-import { Translate } from 'app/I18N';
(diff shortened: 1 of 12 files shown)


RafaPolit:
Yeah, I think this works nicely.  On the actual tooltip, are we now doing `<Translate>` components with such a long "key", or do we still use for this type of long text a shorthand key like `ix_accuracy_tooltip` and then in the actual value of the translation we add the long text?

Zasa-san:
> Yeah, I think this works nicely. On the actual tooltip, are we now doing `<Translate>` components with such a long "key", or do we still use for this type of long text a shorthand key like `ix_accuracy_tooltip` and then in the actual value of the translation we add the long text?

Will do the key:value strategy if it's too long, for sure.

PR #8387, opened 2025-09-01
Create funding-manifest-urls

aphilop:
Required for using a funding.json as described in https://fundingjson.org/


aphilop:
FYI @vleiloerie

PR #8403, opened 2025-09-04
New batch flows for IX

### LINKED ISSUE #8277: [IX] -  Bulk get and accept suggestions
```
MODAL: Run extractor

[x] Find [  1000] suggestions for:
  [x] Non processed
  [ ] Obsolete
  [ ] Error

[x] auto-accept suggestions
  (o) from previous step
  ( ) from all suggestions
  --
  (o) For entities with blank values
  ( ) For all entities (overwrite existing values)

[RUN]
```
notes. 
 - Input box defaults to 1000. 
 - Radio buttons are mutually exclusive for UX readability. 
 - Radios default to the first option (which is the safe one, less resource intensive one).
 - Checks default to the probably most used option (find-suggestions, non processed, auto-accept)
 - The check boxes for Find and auto-accept, grey out the whole section if unchecked. So un-checking the "Find ..." will disable and grey out non processed, obsolete and error and so on so forth.

Example usages:

1. User wants to get suggestions and auto-accept newly added documents. 
    They don't need to tweak any params except the number of entities to process if it is more than the default and they want to process all of them at once.
2. User wants to finish running a extractor that is already trained but has some pending suggestions as obsolete or non-processed. The confidence on the extractor is high and they want to accept the resulting suggestions.
    They'll need to check the Obsolete, so they are included. They can leave the radio "from previous step" on and the suggestions processed will be the desired ones, if they switch the radio to "from all suggestions", they will also process other suggestions meeting the criteria, which is ok.
3. User wants to get suggestions without accepting any values. 
     They un-check the "auto-accept", all the "auto-accept" sections gets grayed out. Suggestions will come and be available for revision/manual acceptance.
4. User wants to accept suggestions that have already been inferred.
     They un-check the "Find" and the section is disabled/grayed out. The radio for "previous step/all suggestions" is forced into "all suggestions" and the user cannot change it.

```
MODAL: Train model

Training machine learning models may take from minutes up to a 
couple of hours depending on the amount of labeled data and the 
difficulty of the task. 

[x] Find [  1000] suggestions after training

[TRAIN]
```

notes.
- This modal acts as a confirmation so users don't trigger trains accidentally. 
- On the top of it, it can serve to choose the amount of suggestions to get after training for model evaluation.
- Users can un-check the "find", so no suggestions are inferred after training.
- By default "find" is checked and the amount is 1000.
- In the future we will add options related to the manual selection of the train set.

RafaPolit:
fixes #8277

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/api/services/informationextraction/IXServices.ts
+++ b/app/api/services/informationextraction/IXServices.ts
@@ -9,7 +9,7 @@ import { propertyIsMultiValued } from 'shared/getIXSuggestionState';
 import { IXSuggestionsModel } from 'api/suggestions/IXSuggestionsModel';
 import { IXModelType } from 'shared/types/IXModelType';
 import { TemplateSchema } from 'shared/types/templateType';
-import ixmodels, { TEST_RUN_SUGGESTIONS_SIZE } from './ixmodels';
+import ixmodels, { DEFAULT_MAX_SUGGESTIONS_SIZE } from './ixmodels';
 
 type GetTargetPropertyInput = {
   extractor: IXExtractorType;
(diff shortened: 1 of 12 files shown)


RafaPolit:
Closing in favor of #8407

PR #8427, opened 2025-09-16
249 - media fields saving serialized file

mfacar:
fixes [#249](https://github.com/huridocs/Internal-Issues/issues/249)

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


Zasa-san (changes requested):
I'm noticing an issue with media fields in public forms:

- Create a template with media and image fields.
- Make it a public template.
- Use it in a page like `<PublicForm template="_id" attachments />`
- Submit filling the media or image field with a file uploaded from the computer

The entity is created with the supporting file, but it is not assigned to the field.

(Firefox browser)

(I still have the code review pending, sorry!)

mfacar:
@Zasa-san thank you for testing the changes. I wasn't able to replicate the error in the public forms. Please let me know if you have a chance to help debug it.

Zasa-san:
> @Zasa-san thank you for testing the changes. I wasn't able to replicate the error in the public forms. Please let me know if you have a chance to help debug it.

Yes, I had an issue with my local environment that was causing the problem. Never mind the report.

app/api/entities/entitySavingManager.ts
```diff
@@ -3,6 +3,34 @@ import entities from 'api/entities/entities';
 import { EntityWithFilesSchema } from 'shared/types/entityType';
 import { UserSchema } from 'shared/types/userType';
 import { handleAttachmentInMetadataProperties, processFiles, saveFiles } from './managerFunctions';
+import { isValidUrl, sanitizeUrl } from 'shared/urlValidationUtils';
+
+const validateAndSanitizeUrls = (entity: EntityWithFilesSchema): EntityWithFilesSchema => {
+  if (!entity.metadata) return entity;
+
+  const sanitizedEntity = { ...entity };
+  if (!sanitizedEntity.metadata) return sanitizedEntity;
```
Zasa-san, line 12: Is this check not covered in line 9?

app/api/entities/entitySavingManager.ts
```diff
@@ -3,6 +3,34 @@ import entities from 'api/entities/entities';
...
+
+  Object.entries(sanitizedEntity.metadata).forEach(([_key, values]) => {
+    if (Array.isArray(values)) {
+      values.forEach(value => {
+        if (value && typeof value.value === 'string' && value.value.startsWith('http')) {
```
Zasa-san, line 17: This means a regular text field o a rich text field starting with that pattern will also get sanitized. If that's intended then it's ok.

app/react/Metadata/helpers/wrapper.js
```diff
@@ -10,31 +37,120 @@ const prepareFiles = async (mediaProperties, values) => {
...
           return Promise.resolve();
         }
-        const { data, originalFile } = values.metadata[p.name];
+
+        const metadataValue = values.metadata[p.name];
+        if (shouldSkipValue(metadataValue)) {
+          return Promise.resolve();
+        }
+
+        if (typeof metadataValue === 'string' && /^https?:\/\//.test(metadataValue)) {
```
Zasa-san: I think some of these checks are covered in shouldSkipValue

Zasa-san (approved):
LGTM!

app/api/entities/entitySavingManager.ts
mfacar: @Zasa-san, I'm leaving the sanitization for all the properties in order to keep the method simple, given it is applied only for texts starting with the HTTP protocol, it's safe to keep it.

app/react/Metadata/helpers/wrapper.js
mfacar: Cleaned

app/api/entities/entitySavingManager.ts
mfacar: There is, cleaned

PR #8454, opened 2025-09-22
8453 titles indexes

txau:


txau:
To be included in https://github.com/huridocs/uwazi/pull/8452

PR #8552, opened 2025-10-27
Add file cache headers feature flag and implement caching logic in fi…

txau:
…le routes

fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/files/routes.ts
```diff
@@ -65,6 +67,54 @@ const filterByEntityPermissions = async (fileList: FileType[]): Promise<FileType
...
+    return false;
+  }
+
+  try {
+    const relatedEntities: EntitySchema[] = await entities.get(
+      { sharedId: file.entity },
+      'published',
+      { withoutDocuments: true }
+    );
+
```
txau: I am not sure about this try/catch block. Always weary of a catch block with silent behavior (although it returns false). What would be the logic and reason for not passing the try block?

app/api/files/routes.ts
```diff
@@ -221,15 +271,47 @@ export default (app: Application) => {
...
+          // Set Last-Modified header
+          if (file.creationDate) {
+            const lastModified = timestampToHTTPDate(file.creationDate);
+            res.setHeader('Last-Modified', lastModified);
+          }
+        }
+      } catch (error) {
+        // If any cache header logic fails, silently skip it and continue serving file
+        // This ensures backwards compatibility and prevents blocking file access
+      }
```
txau, line 326: Again a catch block failing silently

daneryl: Yes, lets not leave it like this please

> // This ensures backwards compatibility and prevents blocking file access

What this makes sure of is to hide potential important errors, let the error bubble up, in this case it is probably enough to remove the try/catch block

app/api/files/routes.ts
daneryl: We should not leave it like this, an error is an error, and transforming that error in a response like `false` is going to be a potential pain, let the error bubble up. 
In case there are specific known operational errors then conditionally catch them on the catch block.

daneryl (changes requested):

daneryl (approved):

Issue #8605, opened 2025-11-18
Search tab v2


konzz:
Implement the search tab in the entity viewer V2.

- Blank state
- Be able to search for snippets in the document.
- Represent the snippets. We don't have a design for this, but we can use cards like for the references without the template name and the buttons, just the text.
- Be able to navigate from the library to that tab, after a snippet search, and also navigate to the correct page.

[Desgins](https://www.figma.com/design/muX7sogg0CTEbjtdmR36ae/Full-V2?node-id=13-14853&p=f&t=O78ylbO4CaXADyF1-0)

<img width="397" height="598" alt="Image" src="https://github.com/user-attachments/assets/773e7ba3-b6c4-46ca-afe5-0f1d0e3dda37" />

Zasa-san:
Implementation visual reference:

<img width="1522" height="895" alt="Image" src="https://github.com/user-attachments/assets/6b351af9-7deb-421e-a1dd-5c5bdc048bbf" />

<img width="1522" height="895" alt="Image" src="https://github.com/user-attachments/assets/f9520418-b238-4e48-b6d6-e060b190eaab" />

<img width="1522" height="895" alt="Image" src="https://github.com/user-attachments/assets/cd063635-7398-484f-a467-ee833c39a5a7" />

konzz:
Wow, you nailed it!

Question, I think that in references, they have a slight border highlight when selected or clicked, have you consider something like that?

Zasa-san:
> Wow, you nailed it!
> 
> Question, I think that in references, they have a slight border highlight when selected or clicked, have you consider something like that?

Yes, I did not showcase it but they have a indigo highlight when selected, and they have hover effect too.

Zasa-san:
#8630 & #8613 implemented the search tab with scroll to page for the result.

I have avoided implementing the last point of directly navigating from the library to the search results since that's not currently possible given our route definitions, this cannot be tested on a production environment. 
I would like to discuss this with @huridocs/frontend to figure out the best way of doing that (atoms, router state, imputing the search on the navigation link). 

However the route knows to render the search tab by default if there's a search result returned from the loader on first render. We can take advantage of this with pretty much any strategy we want to use in the end.

PR #8613, opened 2025-11-20
8605 - V2 search tab

Zasa-san:
part of #8605 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


Zasa-san:
@konzz @mfacar I'm merging this PR even if the issue is not done.

Here we have some important things for the viewer that can be used in other developments:
- Event bus for the pdf to know when it's ready, what page is on, imperative scroll to page (tested).
- Preserve side tab when switching main tabs if possible (tested).
- I have refactored the different instances of query parameters set or get using constants to avoid typos.
- Display search results

What's missing:
- Pain results on PDF
- Scroll to result and not just to page
- Default to opening the search tab when there is a search in the URL.
- Testing the new search tab.

PR #8636, opened 2025-11-26
Feat/bulk-delete-entity

### LINKED ISSUE #8628: Entities Bulk Delete V2
Implement Happy path of bulk delete V2.

- input of entities id, should have a limit.
- delete from elastic
- dispatch jobs that will actually delete entites (batches of 50 ?)

Joao-vi:
fixes #8628

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/core/application/BulkDeleteEntity.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    await this.deps.search.bulkDeleteBySharedId(sharedIds);
+
+    const chunks = ArrayUtils.splitInChunks(sharedIds, 100);
+
+    await ArrayUtils.sequentialFor(chunks, async chunk =>
+      this.jobsDispatcher.dispatchMany(async dispatch =>
+        dispatch(BatchDeleteEntityJob, { sharedIds: chunk })
+      )
+    );
```
daneryl: In order to take advantage here of the dispatchMany it should be inverted:

```ts
    this.jobsDispatcher.dispatchMany(
      async dispatch =>
        await ArrayUtils.sequentialFor(chunks, async chunk =>
          dispatch(BatchDeleteEntityJob, { sharedIds: chunk })
        )
    );
```

otherwise its the same as using the normal ```jobsDispatcher.dispatch()```
also probably you do not need ArrayUtils here, just forEach or for loop

app/api/core/application/BulkCleanupEntity.ts
```diff
@@ -0,0 +1,50 @@
...
+class BulkCleanupEntityUseCase extends AbstractUseCase<Input, Output, Deps> {
+  static InputSchema = InputSchema;
+
+  protected async executeAsync(input: Input): Promise<Output> {
+    const { deleteEntities } = InputSchema.parse(input);
+    const sharedIds = deleteEntities.map(e => e.sharedId);
+
+    await this.transactionManager.run(async () => {
+      await this.deps.relationshipsDS.bulkDeleteBySharedId(sharedIds);
+      await this.deps.entitiesDS.deleteReferencesToSharedIds(deleteEntities);
```
Joao-vi: `await this.deps.entitiesDS.deleteReferencesToSharedIds(deleteEntities)`

This operation turns out to be a big problem: 

Amount of documents involved here are not under of our control, a tenant might have entity deleted that might be referenced in thousand of thousands of other entities. 

But here is even worse because is a bulk operation, image 100 deleted entities that might be connected to infinite of documents that needs to be cleaned up (remove the deleted sharedId from its metadata values).

I tested locally, deleting 930 entities, first part of the process went smoothly (`BulkDeleteEntityUseCase`), now while running this use case (`BulkCleanupEntityUseCase`) for the 10 jobs created, 9 were chunks of 100 entities, the only one that was able to finish was of the chunk of 30 entities (still this got 4 retries). 

All errors messages was regarding transactions timeout.

@daneryl

Joao-vi: Big win! 

I was able to improve this a lot! We went from a state where Jobs with 100 chunks were not even be able to be executed to being now executed in about 1.8~2s

The biggest bottle neck was quering all affected entities so we can dispatch reindex and to update on mongo. In order to do that we were scanning the whole entities collection. 

I am now, first, filtering entities that have metadata that might have a reference for the entity as a value,  this being `select`, `multiselect` and `relationship` 

@daneryl

PR #8696, opened 2025-12-14
Fix #8683: Refresh target template list when new template is created

### LINKED ISSUE #8683: Paragraph extraction
When adding an extractor, if the target template does not exist/meet the criteria, user needs to create this template. After creation, it does not show up in the list of available target templates until forcing refresh on the browser.

m4cd4r4:
Fixes #8683

The target template dropdown wasn't updating when you created a new template. Had to refresh the browser to see it.

Added a useEffect hook to sync the local state with the templates atom when it changes. Now it updates immediately.

Tested by creating a new template during extractor setup and it showed up right away.


konzz:
@m4cd4r4 This is amazing, thanks!

Could you point the PR to huridocs:production as we have changed the main branch.

m4cd4r4:
Thanks! I updated the target branch to production - should be good now.

PR #9083, opened 2026-05-12
9067 - Regressions in metadata extraction review UI 

### LINKED ISSUE #9067: [IX] - Regressions in metadata extraction review UI
**Describe the bug**
While working in the review UI in metadata extraction, some things that used to work as expected, now they don't. Below a list of observed issues:

- [ ] In the side panel, when a value is selected, the checkbox "Use for training" gets selected automatically. However, when selecting the "Accept" button the action deselects the checkbox, so when the panel is closed the relevant entry is not marked to be used for the training. You have to open the panel again and explicitly select the checkbox and accept again to mark the entry to be used for the training.

- [ ] In the side panel, when a value is selected and the panel is closed through the "Accepte" button, the selected value is not shown as the current value in the table.

Both problems can be observed in the following screen recording:

https://github.com/user-attachments/assets/7aea7ba6-776a-4a32-82ce-96debf11d5a6

---

- [ ] If I remember correctly, the popup that is displayed when you hover over a text in the context column, used to highlight the entry that was marked as the relevant context. This is not the case anymore.

Image

mfacar:
fixes #9067

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review

### DIFF
--- a/app/react/V2/Routes/Settings/IX/components/sidepanel/PDFSidepanel.tsx
+++ b/app/react/V2/Routes/Settings/IX/components/sidepanel/PDFSidepanel.tsx
@@ -124,7 +124,11 @@ const PDFSidepanel = ({
   const onSubmit = async (value: {
     field: PropertyValueSchema | PropertyValueSchema[] | undefined;
   }) => {
-    if (dirtyFields.field && entity?._id) {
+    const fieldDirty = dirtyFields.field;
+    const trainingSetDirty = dirtyFields.inTrainingSet;
+    const inTrainingSet = formContext.getValues().inTrainingSet || false;
+
+    if (fieldDirty && entity?._id) {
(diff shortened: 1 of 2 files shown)


mfacar:
IX sidepanel `onSubmit` handler,  in PDFSidepanel.tsx and PropertySidepanel.tsx, was reading dirtyFields and inTrainingSet after the async handleEntitySave call, at that time React Hook Form already reset those values, causing the `Use for training` checkbox to appear deselected and the accepted value not to reflect in the table. The fix captures dirtyFields and inTrainingSet before the async call and uses those captured values for the subsequent onEntitySave logic.

PR #9167, opened 2026-05-29
Fix HMR

### LINKED ISSUE #9166: HMR does not work
HMR (Hot Module Reload) does not work.
The command `yarn hot` was named like that because it should be reloading the client while you make changes for fast feedback loop.

This has been broken for ages now.

Possible fixes;
- Fix HRM so it works again as expected
- Rename the command to `yarn cold`

konzz:
fixes #9166

### Context

This PR fixes our HMR issue in the React Router setup.

Previously, updates were compiled (`[HMR] Updated modules`), but some route screens remained stale in the browser until a manual refresh, in most cases a **reboot of the server was needed**.

### What Changed

- Migrated route definitions to route-level lazy loading in `app/react/Routes.tsx`.
- Kept only base/index-level routes eagerly loaded.
- Preserved existing route behavior while changing how route modules are loaded and updated.

### Why This Fixes HMR

With eager route imports, route elements were effectively captured in a static router tree, making hot updates useless.

By moving routes to lazy loading, route modules are resolved at route boundaries, allowing HMR to correctly refresh those modules without requiring a full page reload.

### Results

- Route edits (e.g. `Entity`, `Settings/Users`) now apply without requiring a manual refresh (`F5`).
- HMR behavior is stable with React Router in our current webpack setup.

### Bundle Impact (Production Builds)

Measured on `uwazi` (old synchronous routes) vs `uwazi-wt-second` (new lazy routes):

| Asset | Before | After | Change |
|---------|---------|---------|---------|
| `main.js` (gzip) | 344.5 KB | 227.8 KB | **-34%** |
| Initial JS (`main + vendor + nprogress`, gzip) | 1.64 MB | 1.51 MB | **-8%** |
| Initial JS + CSS (gzip) | 1.74 MB | 1.61 MB | **-7.5%** |

### Route-Level Loading Behavior

Lazy routes also improve payload distribution:

- Navigating to `/entityv2/...` loads route chunks on demand (~30.2 KB gzip).
- Navigating to `/settings/users` loads its chunk on demand (~7.7 KB gzip).
- Previously, much of this route code was bundled into the initial `main.js`.

### Additional Notes

- Total JavaScript size across all chunks remains roughly the same; the primary improvement is **when** code is downloaded (initial load vs. on-demand route navigation).
- This approach aligns better with long-term route scalability and incremental bundle growth.


app/react/RoutePending.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { Loader } from './components/Elements/Loader.js';
+
+export const RoutePending = () => (
+  <div className="flex h-full min-h-[12rem] w-full items-center justify-center">
+    <Loader />
```
Zasa-san, line 6: We might want to use some v2 thing here instead of a an old loader? Is this always going to be in a tw-content context?

app/react/lazyRoute.tsx
Zasa-san, line 1: I think this is the best part (after some code cleanup is done). Since lazy loading routes should reduce load times.

app/react/entry-client.tsx
```diff
@@ -63,7 +63,42 @@ const App = () => {
...
+// lazy route modules are resolved again (dynamic import picks up HMR replacements).
+const renderClient = () => {
+  if (!clientRoot) {
+    return;
+  }
+  router = createBrowserRouter(getAppRoutes());
+  clientRoot.render(<App />);
+};
+
+if (typeof module !== 'undefined' && module.hot && clientRoot) {
```
Zasa-san, line 80: This checks to `module` is triggering an eslint warning:
```
Cannot use import declarations in modules that export using CommonJS (module.exports = 'foo' or exports.bar = 'hi')
```
I don't know if it's something to consider.

app/react/lazyRoute.tsx
Zasa-san, line 1: The hot reload, I'm not seeing it. 
The test I'm doing is:
- Go to settings
- Edit `app/react/V2/Routes/Settings/SettingsNavigation.tsx` to remove an item from the side navigation bar.
- Expect to see the item gone, but it remains there even if I navigate.

The item only goes when I reload, which is the regular behavior we have right now.

Zasa-san, line 1: One you do a production build, chunks are loaded with serialized names like "326.25eb439cf33c5fade893.bundle.js". 

We could import routes with `loadable` like we do for the PDF components, that allows you to name the chunks. Could be helpful if we need to debug in case there's an error loading a chunk, it's easier to know which one. 

Maybe we could research implementing this with `loadable/components` and `loadable/server` ?

PR #9254, opened 2026-06-18
9239 - V2 user create flow

### LINKED ISSUE #9239: [Users] Create user use case
### Description
As an **admin**, I want to create a new user account in the system so that the user can access the platform.

### Actor
- Admin (only users with role `admin` can create users)

### Input

```
{
  username: string        // required, must be unique, no spaces
  email: string           // required, must be unique
  role: 'admin' | 'editor' | 'collaborator'  // required
  password?: string       // optional — if omitted, system generates a strong random password
  groups?: Array<{        // optional — initial group memberships
    _id: string
    name: string
  }>
}
```

### Output

The created user object, with sensitive fields stripped (no `password`, `failedLogins`, `accountLocked`, `accountUnlockCode`).

### Side Effects

1. An email is sent to the user's address with a "Welcome to [site name]" message containing a password setup link (valid for 24 hours).
2. If `groups` are provided, the user is added as a member of those user groups.

### Database Storage

**`users` collection** — creates a document with:

| Field | Source | Notes |
|---|---|---|
| `username` | Input | Unique index |
| `email` | Input | Unique index |
| `role` | Input | |
| `password` | Input or auto-generated | Stored as bcrypt hash; excluded from default queries (`select: false`) |
| `using2fa` | — | Explicitly set to `false` (not settable via this use case) |
| `secret` | — | Explicitly set to `undefined` (not settable via this use case) |

**`usergroups` collection** — if `groups` provided, each referenced group document gets the new user's `_id` added to its `members` array.

**`passwordrecoveries` collection** — creates a recovery record with `key`, `user` (user's `_id`), and `expiresAt` (TTL: 24h) to support the password setup link.

### Business Rules / Invariants

1. **Username uniqueness** — no two users can share the same username.
2. **Email uniqueness** — no two users can share the same email address.
3. **No spaces in username** — username must not contain space characters.
4. **Cannot set 2FA fields** — `using2fa` and `secret` must not be settable through this use case; they are managed exclusively through a separate 2FA setup flow.
5. **Cannot create system users** — the public user (fixed `_id`) cannot be created here; it's a system-internal user.

### Error Conditions

| Condition | HTTP Status | Message |
|---|---|---|
| Username already exists | 409 | "Username already exists" |
| Email already exists | 409 | "Email already exists" |
| Username contains spaces | 400 | "Usernames can not contain spaces." |
| Missing required fields | 400 | (per schema validation) |

Zasa-san:
fixes #9239 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/core/application/CreateUser.ts
```diff
@@ -0,0 +1,54 @@
...
+
+class CreateUser extends AbstractUseCase<Input, Output, Dependencies> {
+  static InputSchema = UserCreateSchema;
+
+  async execute(input: Input): Promise<Output> {
+    const { password, ...userData } = input.user;
+
+    const user = new User({ _id: this.idGenerator.generate(), ...userData });
+
+    const usernameExists = await this.deps.usersDS.checkUniqueUsername(user);
```
Joao-vi: The `checkUniqueUsername` can return a ResultType, like this `checkUniqueUsername(): Promise<ResultType<boolean, UsernameExists>>` and then the caller can choose to get the data or throw, avoiding to manually throw at the client call. 

Example of real usage: 

`await this.deps.templatesDS.getDefaultTemplate()).getDataOrThrow()` it throws for you the correct error OR get the data if needed.


Same idea for `checkUniqueEmail`
@Zasa-san

app/api/core/application/CreateUser.ts
```diff
@@ -0,0 +1,54 @@
+import { randomBytes } from 'node:crypto';
+import { encryptPassword } from '#api/auth/encryptPassword.js';
+import { User } from '../domain/user/User.js';
+import { AbstractUseCase } from '../libs/UseCase.js';
+import { CreateUserDTO, UserCreateSchema } from './contracts/UserCreateSchema.js';
+import { UsersDataSource } from './contracts/UsersDataSource.js';
+import { EmailInUse, UsernameExists } from '../domain/user/errors.js';
+import { UsergroupsDataSource } from './contracts/UsergroupsDataSource.js';
+
+type Input = { user: CreateUserDTO; domain: string };
```
Joao-vi: I believe there's no need to create a separated file to hold the `CreateUserDTO` and its schema. This type along with its schema can be created right here on this file and be imported at Controller.

Because here we define the type and schema of the Input of this use case which is: 

```ts
const CreateUserInputSchema = z.object({
  username: z.string().trim(),
  role: z.enum(['admin', 'editor', 'collaborator']),
  email: z.string().email(),
  groups: z.array(z.object({ _id: z.string(), name: z.string() })).optional(),
  password: z.string().optional(),
  domain: z.string()
});
```

The above is the Input of Create User use case. 

```ts
type Input = z.inferType<typeof Create...>
```

The http contract will be different already, because it does not requires to send `domain` field. So you can declare contract properly at contracts shared folder, which will be the `CreateUserDTO` exactly what have now. 

@Zasa-san

app/api/core/application/CreateUser.ts
```diff
@@ -0,0 +1,54 @@
...
+      throw new UsernameExists(user.username);
+    }
+
+    if (emailInUse) {
+      throw new EmailInUse(user.email);
+    }
+
+    const rawPassword = password ?? randomBytes(32).toString('hex');
+    const encryptedPassword = await encryptPassword(rawPassword);
+    user.setPassword(encryptedPassword);
```
Joao-vi: `user.setPassword(value: string)` is very dangerous right now because it can accept a primitive string type and for this reason allowing to save a user's password in plan text, without encryption. 


What do you think of using a domain object called `EncryptedPassword` which is a class that represents a string password that is encrypted, the set password method would only start to accept that as parameter: `user.setPassword(password: EncryptedPassword)` no primitive types at all

`EncryptedPassword` can also have a factory method called `async static create(value?: string)` which creates a new password for us and makes sure to run the encryption behind the scenes, returning for us a secure password value. 

```ts
const encryptedPassword = await  EncryptedPassword.create(password) // password is a string or undefined, factory method generates a random value if undefined.

user.setPassword(encryptedPassword)
```

@Zasa-san

app/api/core/infrastructure/express/users/CreateUserController.ts
```diff
@@ -0,0 +1,23 @@
+import type { CreateUserResponse } from '#shared/contracts/Users.js';
+import { AbstractController } from '#api/common.v2/infrastructure/AbstractController.js';
+import { CreateUserUseCaseFactory } from '../../factories/CreateUserUseCaseFactory.js';
+import {
+  UserCreateSchema,
+  CreateUserDTO,
+} from '#api/core/application/contracts/UserCreateSchema.js';
+import { ExecutionContext } from '#api/core/libs/ExecutionContext.js';
+
+class CreateUserController extends AbstractController<CreateUserDTO> {
```
Joao-vi: This request payload type should come from shared contracts folder, something like `CreateUserRequest` same pattern used for `CreateUserResponse`, which is already in use! Nice. 

@Zasa-san

app/api/core/infrastructure/express/users/CreateUserController.ts
```diff
@@ -0,0 +1,23 @@
...
+  protected async handle(): Promise<void> {
+    const newUser = UserCreateSchema.parse(this.request.body);
+    const useCase = CreateUserUseCaseFactory.default();
+    const { password, ...user } = await useCase.execute({
+      user: newUser,
+      domain: `${this.request.protocol}://${ExecutionContext.tenant.domain}`,
+    });
+    const response: CreateUserResponse = { user };
```
Joao-vi: I would avoid to return the `user` object got from the use case output, in the future we might have internal fields on `User` domain object or private information that will be returned to the client. 

I would manually map the return type 

```ts
const response : CreateUserResponse = {
  user: {
    _id: user.id, 
    email: user.email
    ....
  }
}
```

@Zasa-san

app/api/core/application/contracts/Dispatcher.ts
```diff
@@ -48,6 +48,12 @@ type DeleteLanguageEntitiesParams = {
   language: LanguageISO6391;
 };
 
+type ConfigureRecoveryPasswordHandlerParms = {
```
Joao-vi: I believe there's a typo here: `...Parms` 

@Zasa-san

app/api/core/infrastructure/factories/UsergroupsDataSourceFactory.ts
```diff
@@ -0,0 +1,25 @@
...
+import { TransactionManager } from '#api/core/application/contracts/TransactionManager.js';
+import { ExecutionContext } from '#api/core/libs/ExecutionContext.js';
+import { UsergroupsDataSource } from '#api/core/application/contracts/UsergroupsDataSource.js';
+import { MongoUsergroupsDataSource } from '../mongodb/user/MongoUsergroupsDataSource.js';
+
+type Overrides = {
+  transactionManager?: TransactionManager;
+};
+
+class UsergroupsDataSourceFactory {
```
Joao-vi, line 6: I believe this would be better for reading as `UserGroups...`  and same for the file's name

@Zasa-san

Zasa-san: In this case I'd prefer the name match the collection "usergroups". It might no look as nice but I think it makes it visually more clear. I can change it if you think it's better.

app/api/core/infrastructure/jobs/ConfigureRecoveryPasswordHandler .ts
```diff
@@ -0,0 +1,21 @@
...
+} from '#api/core/libs/queue/application/contracts/UserAwareDispatchable.js';
+import users from '#api/users/users.js';
+
+type ConfigureRecoveryPasswordHandlerParams = UserAwareDispatchableParams & {
+  domain: string;
+  userId: string;
+  newUser?: boolean;
+};
+
+class ConfigureRecoveryPasswordHandler extends UserAwareDispatchable<ConfigureRecoveryPasswordHandlerParams> {
```
Joao-vi: As far as I understood, this job will only be used to sending an "welcome email" for the created user, therefore I would change the name to `SendWelcomeEmailHandler`, so `newUser` flag is not necessary anymore and the job will do just one thing. 

In the future, there will be a different job, called `SendRecoverPasswordEmailHandler`. 

@Zasa-san

app/api/core/infrastructure/mongodb/user/UserDBO.ts
```diff
@@ -0,0 +1,13 @@
+import { ObjectId } from 'mongodb';
+
+interface UserDBO {
+  _id: ObjectId;
+  username: string;
+  role: string;
```
Joao-vi: Here you can type role as 

```ts
role: UserRole // imported from domain
```

The persistent layer should follows what domain says. 

@Zasa-san

app/api/core/infrastructure/mongodb/user/MongoUsersMapper.ts
```diff
@@ -0,0 +1,31 @@
+import { ObjectId } from 'mongodb';
+import { User } from '#api/core/domain/user/User.js';
+import { EncryptedPassword } from '#api/core/domain/user/EncryptedPassword.js';
+import { UserDBO } from './UserDBO.js';
+
+export class MongoUsersMapper {
```
Joao-vi, line 6: I would create a dedicated test suite for the mapper.

Since the `password` field may or may not be loaded from the database, outside of the login and change-password use cases we'll usually work with a `User` domain object whose `password` is `undefined`.

This becomes risky when persisting changes because we could accidentally overwrite the database value with `null` or `undefined`.

To avoid that, I would make `toDBO()` include `password` only when its value is not `undefined`:

* `undefined` → password was never loaded, so don't modify the database field.
* `null` → explicitly clear the password field.
* `string` → the user changed their password and the new value should be persisted.

For `toDomain()`, I would map the value as follows:

```ts
if(db.password) {
 user.password = dbo.password;
}
```

This ensures that a missing password in the database is represented as `undefined` in the domain model, allowing us to distinguish between "not loaded" and "explicitly cleared".

@Zasa-san

app/api/users/routes.js
```diff
@@ -28,25 +28,6 @@ export default app => {
     }
   );
```
Joao-vi, line 29: It's a good idea to keep the v1 route as it is for now and deleting, here's why: 

We will deploy v2 create user behind a feature flag and start enabling it fow a few tenants in order to see if everything is working as expected, in case we need to disable v2, we can simply disable the feature flag and requests will start to be routed again to v1. 

Ideally it would be something like this 

```ts
app.post(
    '/api/users/new',
    needsAuthorization(),
    validatePasswordMiddleWare,
    validation.validateRequest({
      type: 'object',
      properties: {
        body: userSchema,
      },
      required: ['body'],
    }),
    CreateUserController.createHandler()
  );
```

On the controller we route the request to v1 or v2 based on the feature flag: 

```ts
class CreateUserController extends AbstractController<CreateUserDTO> {
  protected async handle(): Promise<void> {
    if(ExecutionContext.tenant.featureFlags.v2CreateUser) { 
      return users
        .newUser(req.body, getDomain(req))
        .then(response => res.json(response))
        .catch(next);
    }
    ....
  }
}
```

@Zasa-san

PR #9257, opened 2026-06-19
Data viz

### LINKED ISSUE #9097: Improve Dataset Endpoint for Aggregations and Visualization
## Summary
Create a dedicated backend endpoint for aggregations that properly supports charts, graphs, and other data-driven page features. The current endpoint is not optimized for visualization use cases and does not return data in formats easily consumable by charting libraries.

## Problem
The existing dataset endpoint requires significant frontend transformation to adapt raw data into formats suitable for visual components. This increases implementation complexity and creates duplicated logic across pages.

## Proposed Solution
- Design a dedicated aggregation-focused backend endpoint.
- Return data structures optimized for visualization consumption.
- Support common aggregation patterns required by charts and dashboards.
- Reduce frontend data transformation requirements.

## Benefits
- Improved support for graphs and visualizations.
- Reduced frontend complexity.
- Cleaner separation of responsibilities between frontend and backend.
- Better scalability for future data-driven page features.

## Stakeholders
- Backend
- Frontend

konzz:
fixes #9097


daneryl:
@konzz  Looking solid overall. Good separation between domain, application, and infrastructure. Found a few issues worth fixing before we merge:

### 1. Transaction boundaries owned by a service, not the use case

https://github.com/huridocs/uwazi/blob/1560f816625e76ea282696e9d7312974c7fc8b30/app/api/dataviz.v2/application/services/persistDatavizSnapshot.ts#L102-L108

The use case should own the transaction boundary. If it ever needs additional transactional work (events, audit logs), it can't — the service already committed.

---

### 2. Refresh job non retryable errors

Think about potential errors that make the job non retry able and throw the appropriate error, non retry able errors inform the job system to stop retrying.

example: https://github.com/huridocs/uwazi/blob/cd9e4714cb80f874971d3e77a46d474f44fde116/app/api/core/infrastructure/jobs/PDFPostProcessJobHandler.ts#L56-L58

---

### 3. Domain validation lives in use cases instead of the domain model

https://github.com/huridocs/uwazi/blob/1560f816625e76ea282696e9d7312974c7fc8b30/app/api/dataviz.v2/application/useCases/UpdateDataviz.ts#L37-L39

This is a domain invariant. It should be inside the Dataviz constructor so it holds regardless of entry point (HTTP, job, CLI, test).

---

### 4. Executor directly accesses entities collection, bypassing the contract

https://github.com/huridocs/uwazi/blob/1560f816625e76ea282696e9d7312974c7fc8b30/app/api/dataviz.v2/infrastructure/mongodb/executor/loadEntityTitleLabels.ts#L22-L28

The executor reaches under the EntitiesDataSource contract to touch db.collection('entities').

---

🟡 Medium

### 5. shouldPersistSnapshotOnSave is only used in one use case

app/api/dataviz.v2/application/services/shouldPersistSnapshotOnSave.ts 

A single-use service file adds indirection without reuse. Inline it or move the logic to the Dataviz domain model or use case

---

### 6. isManual should be a domain model property, not a shared utility import

https://github.com/huridocs/uwazi/blob/1560f816625e76ea282696e9d7312974c7fc8b30/app/api/dataviz.v2/application/useCases/CreateDataviz.ts#L30-L32

Use cases import a shared utility for a concept intrinsic to the domain. Dataviz.isManual would remove this leak.
validateLiveRefreshAllowed looks like also belongs into the domain.

---

### 7. No runtime input validation on complex use case inputs

https://github.com/huridocs/uwazi/blob/1560f816625e76ea282696e9d7312974c7fc8b30/app/api/dataviz.v2/application/useCases/CreateDataviz.ts#L14

No runtime validation means invalid dimensions, enums, or missing fields are only caught deep in the MongoDB executor.

example of input validation:
https://github.com/huridocs/uwazi/blob/cd9e4714cb80f874971d3e77a46d474f44fde116/app/api/customUploads/application/CustomFileUpload.ts#L15-L23

used in:

https://github.com/huridocs/uwazi/blob/cd9e4714cb80f874971d3e77a46d474f44fde116/app/api/customUploads/infrastructure/http/CustomFileUploadController.ts#L22-L24

🟢 Low
### 8. Job naming tweak

As per convention we say JobHandlers are the ones executin the job (usecase)

DatavizScheduledRefreshJob should be DatavizScheduledRefreshJobHandler
RefreshDatavizSnapshotUseCase should be RefreshDatavizSnapshotJob

---

### 9. Too many tiny executor helper files
app/api/dataviz.v2/infrastructure/mongodb/executor/ 
Eight pure pipeline-building helpers (buildFilterMatch, dimensionBucketExpression, etc.) are only imported by the executor and could be private methods.

---

### 10. existsByName leaks an update-scenario optimization into the repository contract

https://github.com/huridocs/uwazi/blob/1560f816625e76ea282696e9d7312974c7fc8b30/app/api/dataviz.v2/application/contracts/DatavizDataSource.ts#L10

The excludeId parameter exists only because UpdateDatavizUseCase needs to exclude the current document from the uniqueness check. A purer contract would be existsByName(name: string) and the use case would handle the exclusion.

konzz:
Review services and factories too

PR #9290, opened 2026-06-25
9250 - Persist user in jobs

### LINKED ISSUE #9250: What happens to Jobs when the user has been deleted?
**Describe the bug**
With the new architecture of Jobs, there are scenarios where a user can initiate a long running process (job), or the queue handler has other jobs that are being processed before, and, BEFORE you get to the job, the user is DELETED.

This is now returning errors, because the user is not found.

We have discussed for the longest time that, most likely, delete users should be a soft-delete, because deleting a user will have a lot of repercussions in different areas: entity creators and permissions, job owners, login sessions, etc.  

Still, we need to find a way to deal with this on jobs: do we simply pick another user? Is the job actually void "no go" because the user is deleted? (probably not!)

Zasa-san:
fixes #9250 

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


Zasa-san:
will be solved via https://github.com/huridocs/uwazi/pull/9298

PR #9319, opened 2026-07-04
Handle malformed select filter values in search

### LINKED ISSUE #9297: TypeError: values.reduce is not a function` in `processParentThesauri` when search filter is a bare array
## Description

When a search request is sent to `/api/search` with a filter value as a bare array (e.g., `filters={"cycle":["ANY_ID"]}`) instead of the expected object format (`{"cycle":{"values":["ANY_ID"]}}`), the backend crashes with `TypeError: values.reduce is not a function` at `app/api/search/search.js:45`.

This happens in the `processFilters` function at line 92, which directly accesses `value.values` and passes it to `processParentThesauri`:

```js
value.values = processParentThesauri(property, value.values, dictionaries, properties);
```

If `value` is a bare array `["ANY_ID"]` instead of an object `{values: ["ANY_ID"]}`, then `value.values` is `undefined` (caught by the falsy guard). However, if `value` is an object where `value.values` is a truthy non-array (e.g., a string `"ANY_ID"`), it reaches `values.reduce(...)` and crashes.

### Backend: `app/api/search/search.js:56-93`

The `processFilters` function assumes all select/multiselect/relationship filter values follow the `{ values: [...] }` shape. It does not normalize or validate the shape before passing to `processParentThesauri`.

### Frontend: standard flow

The standard frontend form flow (via `SelectFilter.jsx` → model path `.filters.cycle.values`) correctly produces `{ values: [...] }`. This is preserved through `processFilters()` in `libraryActions.js` and rison encoding/decoding.

### Where the bare array comes from

The bare array format can reach the API through:

1. **Custom markdown datasets/hooks** (`app/react/Markdown/markdownDatasets.js:61-64`) that construct search queries with bare arrays.
2. **Direct API calls or external integrations** that build `filters` as `{ property: [id] }` instead of `{ property: { values: [id] } }`.
3. **Shared/bookmarked rison URLs** manually constructed with bare array rison syntax `(filters:(cycle:!(id)))` instead of the correct `(filters:(cycle:(values:!(id))))`.

"cycle" is a user-defined property, so the exact client-side source cannot be identified from the codebase.

## Steps to reproduce

1. Send a GET request to `/api/search` with filters as a bare array:
   ```
   /api/search?filters=%7B%22cycle%22%3A%5B%22ANY_ID%22%5D%7D&types=%5B%22templateId%22%5D
   ```
2. Observe the `TypeError: values.reduce is not a function` crash.

Or with a value string instead of array:
   ```
   /api/search?filters=%7B%22cycle%22%3A%7B%22values%22%3A%22ANY_ID%22%7D%7D
   ```

## Suggested fix

Two-part fix:

### 1. Backend defensive fix (recommended)

In `app/api/search/search.js`, normalize filter values in `processFilters` before passing to `processParentThesauri`. Handle bare arrays and non-array values:

```js
if (['select', 'multiselect', 'relationship'].includes(type)) {
  type = 'multiselect';
  // Normalize: if value is a bare array, wrap it
  if (Array.isArray(value)) {
    value = { values: value };
  }
  // Guard: ensure values is always an array
  const vals = Array.isArray(value.values) ? value.values : [];
  value.values = processParentThesauri(property, vals, dictionaries, properties);
}
```

### 2. Client-side investigation

If the bare array format originates from a client-side source, fix the source to use the `{ values: [...] }` format. This is the expected contract.

## Affected files

- `app/api/search/search.js` — `processFilters()` (line 56) and `processParentThesauri()` (line 30)

## Error log

```
TypeError: values.reduce is not a function
    at processParentThesauri (app/api/search/search.js:45:17)
    at app/api/search/search.js:92:22
    at Array.reduce (<anonymous>)
    at processFilters (app/api/search/search.js:57:37)
    at buildQuery (app/api/search/search.js:788:19)
```

## Request that triggered the error

```
GET /api/search?_types=["ANY_ID"]&filters={"cycle":["ANY_ID"]}&limit=200&order=desc&sort=creationDate&from=0&include=["permissions"]
```

jmtdev0:
## Summary

- Normalize legacy search filter values for select-like metadata filters before they reach `processParentThesauri`.
- Support bare arrays, bare scalar values, and `{ values: scalar }` in addition to the existing `{ values: [...] }` shape.
- Apply the same normalization when adding the synthetic `Any` aggregation bucket, since that path reads the original request filters.
- Add regression coverage for both the internal search path and the public `GET /api/search` route.

## Context

Fixes #9297.

The legacy search path assumed select, multiselect, and relationship filters always used the `{ values: [...] }` shape. When a request supplied a bare array or a scalar `values` field, the backend could crash while expanding thesaurus parent values or while post-processing aggregations.

## Testing

- `DEBUG=true node --no-experimental-fetch ./node_modules/.bin/jest app/api/search/specs/search.spec.js app/api/search/specs/routes.spec.ts -w=1`
- `./node_modules/.bin/eslint app/api/search/search.js app/api/search/specs/search.spec.js app/api/search/specs/routes.spec.ts --quiet`

*This PR was developed by Codex under jmtdev0's supervision.*


jmtdev0:
I checked the failing `api_unit_tests` check. It does not appear to be a test failure from this PR.

The failing job is the aggregate `api_unit_tests` job. In this run, `precheck` set `should_run=true`, but `build_ci_image` was skipped because this PR comes from a fork. Since the API test shards depend on `build_ci_image`, the shard job was skipped as well. The aggregate job then saw `needs.api_unit_tests_shards.result` as `skipped` instead of `success` and failed with `One or more shards failed`.

So the check seems to be failing before any API unit tests are executed. The PR-specific local validation I ran remains:

- `DEBUG=true node --no-experimental-fetch ./node_modules/.bin/jest app/api/search/specs/search.spec.js app/api/search/specs/routes.spec.ts -w=1`
- `./node_modules/.bin/eslint app/api/search/search.js app/api/search/specs/search.spec.js app/api/search/specs/routes.spec.ts --quiet`

RafaPolit:
Thank you and thank your for your contribution.  We have identified the problem with our CI for external fork PRs.  We are fixing this right now.  That will allow this PR to finish the tests successfully and, after that, I'll merge your PR.

RafaPolit:
Thanks once more for your contribution.  This fix has now been merged and deployed to our servers.  We appreciate it.  Thank you.

jmtdev0:
Thank you for your work and for reviewing this PR. 

Espero que estos cambios den buen resultado 😉.

PR #9373, opened 2026-07-15
9372 - fix regression in entity saving wrapping entity metadata

### LINKED ISSUE #9372: Geolocation properties: Can not save values
**Describe the bug**
It is not possible to save the value of a geolocation property in an entity, whether is a new entity or an edited one.

**To Reproduce**
1. Go to Library
2. Click on 'Edit' in an entity with a geolocation field.
3. Point a coordinate in the map; or type manually a valid latitude and longitude.
4. Click on 'Save'
5. View the entity and see error: geolocation new value is not saved.

**Expected behavior**
Back to normality,,,

 - [x] Desktop
 - [ ] Mobile

**Additional context**
It is tested in different Uwazi instances, therefore different templates, browsers, computers, human beings...

mfacar:
fixes #9372


salvalacruz:
Thank you @mfacar 
Watching the files changed, just a suggestion here
https://github.com/huridocs/uwazi/pull/9373/changes#diff-30eb0417ef8c244c3ef83985fa3516ec6bf689f49965c6bcc5a1ee5e0dc52791

It seems the code validates a geolocation just by checking that latitude and longitude are numbers; but a valid geolocation should  be in this range:

-90 <= latitude <= 90
180 <= longitude <= 180

Fyi, just in case you think it would be good to include this validation.

mfacar:
Thanks @salvalacruz, absolutely, I'll include that validation in the work I'm doing fro Entity Edition V2 (where I introduced the regression :grimacing:  )

PR #9465, opened 2026-07-29
9448 relationship types postgres

### LINKED ISSUE #9448: [Relationtypes] move to PG

RafaPolit:
fixes #9448

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/core/infrastructure/postgresql/relationshipType/PostgresRelationshipTypesDataSource.ts
```diff
@@ -0,0 +1,93 @@
...
+    const rows = await this.table.all();
+    return rows.map(PostgresRelationshipTypeMapper.toDomain);
+  }
+
+  async getById(id: string): Promise<RelationshipType | null> {
+    const row = await this.table.where({ _id: id }).first();
+    return row ? PostgresRelationshipTypeMapper.toDomain(row) : null;
+  }
+
+  async create(input: { name: string }): Promise<RelationshipType> {
```
daneryl: By convention we pass domain models to create/update operations, id assignment should happen on the use  case with the ID abstraction so we do not require mongo ObjectId in a postgresDS

app/api/core/infrastructure/postgresql/common/ArrayResultSet.ts
```diff
@@ -0,0 +1,111 @@
+/* eslint-disable no-await-in-loop */
```
daneryl: I understand this abstraction, but we are removing as much as we can the usage of the result set, it can be useful in some situations but we mostly load everything into memory, and for sure relationtypes do not need this, so please remove from both mongo and pg versions of datasources the usage of resultSet, and remove also this abstraction which really is for compatibility because it expects everything loaded into memory anyways.

app/api/utils/testingEnvironment.ts
```diff
@@ -180,14 +185,24 @@ const testingEnvironment = {
...
+            .map(([table, fixture]) => {
+              const pgTable = PG_TABLE_BY_MONGO_COLLECTION[table] ?? table;
+
+              return [
+                pgTable,
+                fixture.map((f: any) => {
+                  if (table === 'relationtypes') {
+                    const id = f._id?.toHexString?.() ?? String(f._id);
+                    return { _id: id, name: f.name };
+                  }
```
daneryl: The below sanitized version should cover this, no need for a special relationtypes case.

scripts/scripts.v2/auditRelationshipTypeNameCollisions.ts
```diff
@@ -0,0 +1,146 @@
+/**
```
daneryl, line 1: explain this please 🔍

daneryl (commented):
👍  with some comments.

app/api/utils/testingEnvironment.ts
RafaPolit: The reason for this was that relationtypes had support for properties, which we removed in the PG version of the table.  However, the v1 relationship types were kept as originally designed in order to don't spend time with that.

I'm going to remove support for .properties in V1 as well to remove this special case.

scripts/scripts.v2/auditRelationshipTypeNameCollisions.ts
RafaPolit: ### What this script is for

Before (or while) relying on **case-insensitive + trim** uniqueness for relationship type names (same rule as `existsByName` in the DS), we need to know whether any tenant already has Mongo `relationtypes` docs that collide under that normalization.

Example collision: `"Related To"`, `"related to"`, and `"  RELATED TO  "` would all normalize to `"related to"` and fight each other once PG/app uniqueness is enforced.

This is a **read-only audit** — it does not migrate data or change anything.

### How to run it

```bash
# all tenants
node scripts/runner.js scripts/scripts.v2/auditRelationshipTypeNameCollisions.ts

# one tenant
node scripts/runner.js scripts/scripts.v2/auditRelationshipTypeNameCollisions.ts --tenant <name>
```

### What it does

1. Connects to Mongo and loads tenants.
2. For each tenant (or the one passed via `--tenant`):
   - Skips tenants with no `relationtypes` collection.
   - Aggregates docs grouped by `lower(trim(name))`.
   - Keeps only groups with `count > 1`.
3. Prints a JSON summary:
   - `scannedTenants`
   - `tenantsWithRelationtypes`
   - `tenantsWithCollisions`
   - `collisions`: per tenant, each group’s normalized name, count, original `names`, and `ids`

### Why it exists for the Postgres cutover

Uniqueness stays **application-level** (no unique index on `name` in PG v1). Colliding labels in existing data would make creates/updates fail unpredictably after cutover. Run this against real/staging tenants, remediate (rename one of the labels, etc.), then proceed with data copy + flag flip.

app/api/utils/testingEnvironment.ts
RafaPolit: Done

PR #9575, opened 2026-08-18
9574 - Entity V2 relationship phased endpoints

### LINKED ISSUE #9574: Entity V2 - GET relationships contracts (summary, anchors, resolved)
Back-end services, split from #9497

mfacar:
fixes #9574


daneryl:
@mfacar 

**Overall: looks good.** Clean service, good tests, follows repo conventions.

These new endpoints and their response shapes become a contract the frontend will build against, but the target V2 relationships model isn't designed yet,  we don't know its details, so this
 contract will likely change when it lands. Building the frontend against it now (and committing to these shapes in the API) risks rework on both sides, that being said here is the only changes i would like:

### 1. Remove the `user` parameter from `RelationshipsQueryServiceFactory`

The factory threads `user` through just to hand it to the DAO, but `EntitiesDAOFactory.default()` already resolves the actor from `ExecutionContext` (`overrides?.user ?? ExecutionContext.actor ?? User.createFrom(null)`), which is the documented way — factories consume the actor from context, controllers don't pass it in. Change:

```ts
static default(user: User) {
  const entitiesDAO = EntitiesDAOFactory.default({ user, transactionManager });
  ...
}
```

to:

```ts
static default() {
  const entitiesDAO = EntitiesDAOFactory.default({ transactionManager });
  ...
}
```

### 2. Move the implementation out of `core` into `app/api/relationships/`

This is a bridge for a legacy feature that will be **deprecated** once the V2 relationships lands, it shouldn't live in `core`, which is reserved for the durable domain. Move the new code to `app/api/relationships/`.

mfacar:
I'll merge the client part of the implementation from #9591

Thanks @daneryl, I moved the files into app/api/relationships/query/ keeping data access additions in app/api/core/infrastructure/mongodb/MongoRelationshipsV1DataSource.ts

Joao-vi:
Nice slice @mfacar  — the phased split (summary / anchors / resolved), the ACL coverage in
`routes.spec.ts`, and the `keepHubsWithSource` + `dropSingletonHubs` filtering are all solid.
The comments below are mostly about **where things live and who is allowed to know about whom**,
not about the logic, which I think is right.

Most of them follow from one root observation, so I've put that first.

---

## 1. `RelationshipsQueryService` is an infrastructure service — let's place it as one

**Comment:** This service's job is to fulfil the HTTP API's needs: it takes what the client sends,
reads from the database, and returns exactly what the client needs to paint the UI. That is an
infrastructure service, not an application service. It currently lives under `application/`, which
is what's driving it to hide its dependencies behind hand-written ports and to invent its own
internal DTO layer. Let's move it and let it be what it is.

**Reasoning:** There's no use case here, no domain invariant being protected, no transaction being
orchestrated. `application/` implies a layer that domain and use cases can sit on top of and that
must stay ignorant of delivery and persistence details. This component is ignorant of neither, by
design, and shouldn't have to pretend otherwise.

We already have this pattern in the codebase — `MongoPXEntityStatusesQueryService` and
`MongoPXExtractorsQueryService` live in `app/api/paragraphExtraction/infrastructure/`, and
`MongoUsersQueryService` lives under `core/infrastructure/mongodb/user/`. Read-side query services
belong in infrastructure.

Once it's placed correctly, comments 2 and 3 stop being trade-offs and become obvious.

```
app/api/relationships/query/
├── infrastructure/
│   └── RelationshipsQueryService.ts   ← moved from application/
├── express/
└── factory/
```

---

## 2. Drop the hand-written ports, depend on the DAOs directly

**Comment:** `FilesPort` and `RelationshipsPort` don't need to exist. Import `MongoFilesDAO` and
`MongoRelationshipsV1DataSource` and depend on them directly.

**Reasoning:** An interface is only worth its maintenance cost when we could genuinely swap the
implementation for something else — when it's truly agnostic. That's hard to achieve and it has to
pay off. Neither of these does: there is one files DAO and one relationships v1 data source, and
there's no scenario on the table where we'd substitute either. What we get instead is a second shape
to keep in sync by hand, forever.

Coupling isn't automatically bad. It's bad when it's *uncontrolled*. Here it's controlled and
deliberate: the DAO exists precisely to handle the database for us, and this service and the DAOs
are components of the same layer — they're allowed to know each other.

There's a concrete cost to the ports too. `getHubConnectionsForEntity` has an inferred return type,
and `RelationshipsPort` declares a structurally-compatible shape by hand. Nothing checks that the two
agree — they match today by luck, and they'll drift silently the first time the query's projection
changes. Importing the class makes the compiler enforce it.

Same for the `{ toString(): string }` in `FilesPort` — that's a duck-typed dodge around `ObjectId`
that only exists because the port has to describe a shape it doesn't own.

**One exception: leave `entitiesDAO` on its contract.** `EntitiesDAO` has two live implementations
(`MongoEntitiesDAO` and `PostgresEntitiesDAO`) selected by the `postgresEntities` feature flag. The
swap is real and already paid for. So the mix below is deliberate, not inconsistent — it's the same
rule applied to three dependencies whose circumstances differ.

```ts
// app/api/relationships/query/infrastructure/RelationshipsQueryService.ts
import { EntitiesDAO } from '#api/core/application/contracts/EntitiesDAO.js';
import { MongoFilesDAO } from '#api/core/infrastructure/mongodb/files/MongoFilesDAO.js';
import {
  MongoRelationshipsV1DataSource,
  HubConnection,
} from '#api/core/infrastructure/mongodb/MongoRelationshipsV1DataSource.js';

// FilesPort / RelationshipsPort / the local HubConnection: all deleted

type Deps = {
  entitiesDAO: EntitiesDAO;            // real contract — two implementations behind a flag
  filesDAO: MongoFilesDAO;             // one implementation, no swap in sight
  relationshipsDataSource: MongoRelationshipsV1DataSource;
};
```

And so the mapping is compiler-checked rather than trusted:

```ts
// app/api/core/infrastructure/mongodb/MongoRelationshipsV1DataSource.ts
type HubConnection = {
  _id: string;
  hub: string;
  entity: string;
  template: string | null;
  file?: string;
  reference?: { text: string; selectionRectangles?: SelectionRect[] };
};

async getHubConnectionsForEntity(sharedId: string): Promise<HubConnection[]> { /* ... */ }

export type { HubConnection };
```

---

## 3. `RelationshipQueryDTOs` is the client's contract — move it to `shared/contracts`

**Comment:** These types aren't internal to the backend. Nothing on our side consumes them: no use
case, no domain service, no application service. They exist for one reason — so the client can
populate its UI. That makes them the HTTP API contract, and the contract belongs in
`app/shared/contracts/` so backend and frontend import the same declaration.

**Reasoning:** Worth being precise about the distinction, because it decides whether we need one type
or two.

A **read model** is a server-side projection shaped for internal consumers, kept separate from the
wire format so the two can version independently. That's the pattern in `GetUsersController`, where
`UserProfile` (the read model) is mapped field-for-field to `GetUsersResponse` (the wire contract).
The second type earns its place there because internal callers depend on `UserProfile`.

Nothing depends on `RelationshipSummaryDTO` except the JSON we serialise. Introducing a read model
alongside it would be inventing a consumer that doesn't exist, and paying the mapping cost for a
degree of freedom nobody needs. If an internal consumer shows up later — a use case, another
application service — that's when we create a read model for it, and the mapping will have an actual
reason to exist.

So: one type, in `shared/contracts`, and this service takes and returns it directly. Which is
consistent with comment 1 — an infrastructure service serving the HTTP API should speak the HTTP
API's types.

```ts
// app/shared/contracts/Relationships.ts

type SelectionRect = {
  top: number;
  left: number;
  width: number;
  height: number;
  page: string;
};

type RelationshipSummary = {
  _id: string;
  hub: string;
  entity: string;
  template: string | null;
  file?: string;
  entityData: { title: string; template: string };
};

type RelationshipAnchor = {
  _id: string;
  reference: { selectionRectangles: readonly [SelectionRect] };
};

type RelationshipResolved = {
  _id: string;
  reference: { text: string; selectionRectangles: SelectionRect[] };
};
```

---

## 4. The endpoints have no declared request/response contract

**Comment:** All three controllers are missing their API contract definitions — the `Request` and
`Response` shapes we normally declare in `shared/contracts`. Right now the `{ rows }` envelope exists
only as a literal inside three controller bodies, and the frontend has nothing to import.

**Reasoning:** The envelope is a real part of the contract and it's currently undocumented anywhere.
Declaring the pair per endpoint also gives us a place to pin the zod schema against, so a change to
the schema that diverges from the published request shape fails to compile instead of silently
shipping.

```ts
// app/shared/contracts/Relationships.ts (continued)

type GetRelationshipsSummaryRequest = { sharedId: string };
type GetRelationshipsSummaryResponse = { rows: RelationshipSummary[] };

type GetRelationshipsAnchorsRequest = { sharedId: string; file: string };
type GetRelationshipsAnchorsResponse = { rows: RelationshipAnchor[] };

type GetRelationshipsResolvedRequest = { sharedId: string };
type GetRelationshipsResolvedResponse = { rows: RelationshipResolved[] };
```

```ts
// app/api/relationships/query/express/GetRelationshipsSummaryController.ts
import type { GetRelationshipsSummaryResponse } from '#shared/contracts/Relationships.js';

class GetRelationshipsSummaryController extends AbstractController {
  protected async handle(): Promise<void> {
    try {
      const query = SharedIdQuerySchema.parse(this.request.query);

      const rows = await RelationshipsQueryServiceFactory.default().getSummary({
        ...query,
        language: this.language,
      });

      const response: GetRelationshipsSummaryResponse = { rows };
      this.response.json(response);
    } catch (error: unknown) {
      sendNotFoundRows(this.response, error);
    }
  }
}
```

Pinning the schema to the published request shape:

```ts
// app/api/relationships/query/express/RelationshipQuerySchemas.ts
const SharedIdQuerySchema = z.object({
  sharedId: z.string().min(1),
}) satisfies z.ZodType<GetRelationshipsSummaryRequest>;

const AnchorsQuerySchema = SharedIdQuerySchema.extend({
  file: z.string().min(1),
}) satisfies z.ZodType<GetRelationshipsAnchorsRequest>;
```

Note `language` is deliberately absent from the request contract — it comes from `Accept-Language`
via `AbstractController`, not from the query string.

---

## 5. The module-level helpers should be private methods

**Comment:** `dropRedundantEntityRows`, `dropSingletonHubs`, `keepHubsWithSource` and `toReadableRow`
should be private methods on the class. If eslint complains about not using `this`, disable the rule
for the line or for the whole file.

**Reasoning:** Cohesion. These four functions exist only to serve this class, are used only by this
class, and describe rules that only make sense in this class's context. Having them as loose
module-level functions splits the reader's attention between two scopes for no benefit — you have to
scroll past all four before reaching the thing that actually explains what they're for.

`class-methods-use-this` is a style rule, and this is exactly the case where it gets in the way of
the better structure. Disabling it is the cheaper trade.

```ts
// app/api/relationships/query/infrastructure/RelationshipsQueryService.ts

/* eslint-disable class-methods-use-this */

class RelationshipsQueryService {
  constructor(private deps: Deps) {}

  // ... public methods first

  private dropRedundantEntityRows(rows: ReadableHubRow[]): ReadableHubRow[] {
    const withFile = new Set(rows.filter(row => row.file).map(row => `${row.hub}:${row.entity}`));
    return rows.filter(row => row.file || !withFile.has(`${row.hub}:${row.entity}`));
  }

  private dropSingletonHubs(rows: ReadableHubRow[]): ReadableHubRow[] { /* ... */ }

  private keepHubsWithSource(params: KeepHubsWithSourceParams): HubConnection[] { /* ... */ }

  private toReadableRow(params: ToReadableRowParams): ReadableHubRow { /* ... */ }
}
```

---

## 6. Name the input types

**Comment:** Inline object types in signatures hurt readability. Let's declare a named type alias for
every input introduced in this PR, not just the public ones.

**Reasoning:** A named type gives the parameter a concept instead of a shape, it shows up in
hover/IDE tooling as a single word, and it gives us one place to change when an input grows. It also
makes the family relationship between the three endpoints visible — all three are the same query plus
or minus a refinement, which the inline versions hide.

```ts
// app/api/relationships/query/infrastructure/RelationshipsQueryService.ts

type RelationshipsQuery = {
  sharedId: string;
  language: LanguageISO6391;
};

type GetSummaryInput = RelationshipsQuery;
type GetResolvedInput = RelationshipsQuery;
type GetAnchorsInput = RelationshipsQuery & { file: string };

async getSummary(input: GetSummaryInput): Promise<RelationshipSummary[]> { /* ... */ }
async getAnchors(input: GetAnchorsInput): Promise<RelationshipAnchor[]> { /* ... */ }
async getResolved(input: GetResolvedInput): Promise<RelationshipResolved[]> { /* ... */ }
```

---

## 7. Single parameter object, applied to everything in this PR

**Comment:** Prefer a single parameter object over positional parameter lists. This applies to all
the functions and methods this PR introduces, not only the ones that are long today.

**Reasoning:** It avoids the long-parameter-list anti-pattern before we get there, it makes call
sites self-documenting, and it makes future changes cheap — adding an input becomes a field on a
type rather than an edit at every call site with a positional argument to get right. Doing it
uniformly also means we're not re-litigating "is three parameters too many?" on every review; the
convention decides it.

The private methods need this as much as the public ones — `readableRowsFromConnections(connections,
sharedId, language)` is already three positional arguments, two of them strings.

```ts
// app/api/relationships/query/infrastructure/RelationshipsQueryService.ts

type ToReadableRowParams = {
  connection: HubConnection;
  entityData: EntityLabel;
};

type KeepHubsWithSourceParams = {
  connections: HubConnection[];
  sharedId: string;
};

type ReadableRowsFromConnectionsParams = {
  connections: HubConnection[];
  sharedId: string;
  language: LanguageISO6391;
};

type ReadableFileIdsParams = {
  connections: HubConnection[];
  language: LanguageISO6391;
};

private toReadableRow({ connection, entityData }: ToReadableRowParams): ReadableHubRow {
  // ...
}
```

---

## 8. The factory shouldn't build or prop-drill the transaction manager

**Comment:** `TransactionManagerFactory.default()` shouldn't be instantiated here — it should be
taken from the execution context. The transaction manager is a cross-cutting concern and shouldn't
be prop-drilled through this factory.

**Reasoning:** This one isn't only about style — I think it's a live defect.

`EntitiesDAOFactory.default()` already resolves the transaction manager itself, and it prefers the
ambient one:

```ts
// app/api/core/infrastructure/factories/EntitiesDAOFactory.ts:24-28
const transactionManager =
  overrides?.transactionManager ??
  (ExecutionContext.getStore()
    ? ExecutionContext.transactionManager
    : TransactionManagerFactory.default());
```

Passing `{ transactionManager }` explicitly takes the **override** branch, which defeats that lookup
and forces a fresh, detached `MongoTransactionManager` even when we're running inside an
`ExecutionContext`. Meanwhile the line right below it, `filesDAO: FilesDAOFactory.default()`, passes
nothing — so it *does* resolve from context.

The result is that a single `RelationshipsQueryServiceFactory.default()` call hands its two DAOs
**different transaction managers**: one ambient, one detached. For read-only queries the blast radius
is small today, but it's wrong, and it will bite the moment this service is called from inside a
transaction — the entity reads won't join the ambient session while the file reads will.

The fix is to stop passing it at all and let each factory resolve it. `MongoRelationshipsV1DataSource`
does need one in its constructor, so it gets a factory of its own, consistent with every other
dependency here — which also gets `getConnection()` out of this file.

```ts
// app/api/relationships/query/factory/RelationshipsQueryServiceFactory.ts
class RelationshipsQueryServiceFactory {
  static default() {
    return new RelationshipsQueryService({
      entitiesDAO: EntitiesDAOFactory.default(),
      filesDAO: FilesDAOFactory.default(),
      relationshipsDataSource: MongoRelationshipsV1DataSourceFactory.default(),
    });
  }
}
```

```ts
// app/api/core/infrastructure/factories/MongoRelationshipsV1DataSourceFactory.ts
class MongoRelationshipsV1DataSourceFactory {
  static default() {
    const transactionManager = ExecutionContext.getStore()
      ? ExecutionContext.transactionManager
      : TransactionManagerFactory.default();

    return new MongoRelationshipsV1DataSource(
      getConnection(),
      transactionManager,
      EntitiesDAOFactory.default()
    );
  }
}
```
---

## 10. `sendNotFoundRows` — 404 carrying a success-shaped body

**Comment:** Returning `404` with `{ rows: [] }` gives an error status a body that looks like a
successful empty result. Let's decide the error contract deliberately and write it down, and reshape
the helper so its control flow is visible at the call site.

**Reasoning:** To be clear, the part I think is *right*: collapsing "entity doesn't exist" and
"entity exists but you can't read it" into the same 404 is a deliberate and good choice — it avoids
leaking existence to a collaborator or anonymous user, and the spec covers it.

The problem is the body. Two different outcomes currently serialise to byte-identical JSON:

| Case | Status | Body |
|---|---|---|
| entity missing or unreadable | `404` | `{ "rows": [] }` |
| entity readable, graph empty | `200` | `{ "rows": [] }` |

The status is the *only* signal distinguishing them, and any client that reaches for `body.rows`
without checking `response.status` first will render a normal empty state on an error. Given the
frontend hasn't been written against these endpoints yet, that's a trap we're setting for ourselves.
Either the body should differ, or — if empty-rows really is the intended client-facing semantic for
"not found" — this should arguably just be a `200`, and the 404 is telling us something we don't
actually mean.

Whichever way it goes, it needs to be in the response contract from comment 4, since right now it's
knowledge that lives only in `routes.spec.ts`.

Secondary: the helper reads like a responder but rethrows on the non-matching branch, so control flow
escapes from inside something named `send…`. Either the name should signal it (`handleNotFound…`), or
invert it into a guard the controller branches on explicitly:

```ts
// app/api/relationships/query/express/sendNotFoundRows.ts — current
function sendNotFoundRows(response: Response, error: unknown): void {
  if (error instanceof EntityNotFoundError) {
    response.status(404).json({ rows: [] });
    return;
  }
  throw error; // ← not visible at the call site
}
```

mfacar:
@Joao-vi thank you for the review, I did the requested changes :+1:

Issue #9632, opened 2026-08-27
Tech debt: remove --no-experimental-fetch flag


daneryl:
## Summary
Remove the `--no-experimental-fetch` Node CLI flag from `package.json` scripts, `scripts/runner.js`, and `SELF_HOSTED_INSTRUCTIONS.md`.

## Why it exists
- Added in Aug 2023 (commit `955a8f38f`) when running on Node 18, where `fetch` was still experimental.
- The app calls the bare global `fetch()` in `app/shared/JSONRequest.js` and `app/shared/apiClient/transport/http.ts`, relying on `isomorphic-fetch` to polyfill it. isomorphic-fetch's node entry only installs `node-fetch` as `global.fetch` when `fetch` is undefined, so the flag effectively switches the runtime HTTP implementation between undici (native) and node-fetch 2.6.7 (pinned in package.json).
- Production currently runs with the flag and depends on node-fetch behavior.

## What removing entails
- Strip the flag from all `package.json` scripts, `scripts/runner.js` (`nodeArgs`), and the `SELF_HOSTED_INSTRUCTIONS.md` example.
- Optionally drop the now-dead `isomorphic-fetch` side-effect imports and the pinned `node-fetch@2.6.7` dependency.
- Verify HTTP paths that depend on fetch: login/session cookies, sync, external services (e.g. topic classification), file uploads.

## Verification already done
- Native fetch (undici) is stable since Node 21 and behaves identically to node-fetch for the patterns `JSONRequest` uses (JSON POST/PUT, manual `Cookie` header, `headers.get('set-cookie')`, case-insensitive headers, `URLSearchParams` bodies) — tested locally on Node 22.23.2.
- No code checks `typeof fetch === 'undefined'`, so nothing silently depends on the flag.

## Risk
- `--no-experimental-fetch` is still supported in Node 22 but nodejs/node has discussed eventually making it throw (see nodejs/node#50554 discussion), so this should be done before moving past Node 22.

PR #9633, opened 2026-08-27
Upgrade to Node 22.23.2

daneryl:
## What
Upgrade the runtime to Node 22.23.2 so CI runs green on version 22.

## Changes
- **`.nvmrc`** → `22.23.2` — all CI workflows read `node-version-file: '.nvmrc'`, so they pick up Node 22 automatically.
- **`.github/ci/Dockerfile`** → `node:22.23.2-bookworm` — the unit-tests job runs inside this image; without the bump it would still test on Node 20.
- **`scripts/runner.js`** — added a comment explaining why `--no-experimental-fetch` is kept (production relies on node-fetch via isomorphic-fetch; native fetch is stable since Node 21). Removal is tracked as tech debt in #9632.

## Notes
- `--no-experimental-fetch` is still fully supported on Node 22 (no warning, still disables the global fetch), so no behavior change.
- `package.json` `engines` (`>=20.19.6`) is satisfied by 22; `@types/node` left as-is (types only, separate concern).

## CI
This PR exists to confirm all CI jobs are green on Node 22.


mfacar:
@daneryl I=no changes are required from the FE perspective. Just one minor thing, the version also needs to be updated in the package engines, README, and Cypress Docker image.

daneryl:
> @daneryl I=no changes are required from the FE perspective. Just one minor thing, the version also needs to be updated in the package engines, README, and Cypress Docker image.

done :+1:

PR #9663, opened 2026-09-03
Entityv2 chores rerender

mfacar:
fixes multiple re render in Entity V2


mfacar:
## Reviewer notes (rerender / URL)

**What this is trying to stop**
- **Map:** `markerSyncKey` so Leaflet is not torn down and rebuilt when marker *identity* is unchanged.
- **Tabs:** main pane and side pane stay on separate tab groups so a hash/side-tab update does not rerender the main document surface.
- **URL / page:** document page lives in `entityPageAtom`. Page-only ticks use `history.replaceState` so React Router does not see every page change. `EntityUrlSync` seeds the atom from the router hash on first render (deep links like `#page=N`) and unbinds the module session on unmount. Library preview is wrapped in `EntityUrlSync` so it does not reuse a stale entity navigate/store.

**Create-relationship `sourcePage`**
`store.get(entityPageAtom)` is intentional: after a page-only replace, the router hash can still show the old page. The atom is initialized when `EntityUrlSync` mounts (entity viewer and library preview). Opening the modal outside that tree would still read `'1'`, which matches an absent `page` hash.

PR #9706, opened 2026-09-10
refactor persistence layer of information extraction

### LINKED ISSUE #9705: [IX] Refactor persistence layer of the application
In order to migrate this module to Postgre database in a more confident way, we need to at least isolate mongodb usage across this module, build proper methods, increase cohesion, and decrease coupling between application layer and mongodb internal methods.

Joao-vi:
fixes #9705

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


Joao-vi:
## What this PR fixes, in plain terms

The main goal here is a refactor — information extraction no longer talks to MongoDB directly, which
is what makes the move to Postgres possible. That part is invisible to users.

Along the way, testing against a real running stack (Uwazi + the ML service, both Mongo-backed)
turned up a number of genuine bugs. These are the user-visible ones now fixed:

**Things that lost or corrupted work**

- **Editing an extractor could silently break it.** Opening the Edit modal and changing anything —
  even just the name — could convert a property-based extractor into a PDF one, because the modal
  never received the extractor's real source and made you re-pick it. The extractor then quietly
  stopped producing anything useful.
- **Editing a document's source text left a stale suggestion behind.** The suggestion still showed a
  value taken from text that no longer existed, and because it wasn't flagged as out of date, no
  future run would ever revisit it. Those suggestions are now marked obsolete.

**Things that got stuck**

- **"Cancel" didn't cancel.** It flipped the extractor back to ready but left the work in flight:
  results from the cancelled run were still written, and rows could stay stuck "processing" forever.
  Cancel now releases those rows, clears the run, and discards results that arrive late from a run
  you already cancelled.
- **Training could hang forever, silently.** An extractor configured with a source that can't
  actually be trained on would sit at "processing" indefinitely with no error. It now fails loudly
  and releases the extractor.
- **An extractor could get stuck reporting "finding suggestions"** with nothing left to process.

**Things that reported the wrong thing, or nothing at all**

- **A failed task showed only "Failed"** with no explanation, even though the server had already
  sent the reason. The reason is now shown.
- **Accepting a suggestion that failed showed "A server side error has occurred"** instead of the
  actual reason ("Suggestion(s) not found", "Some suggestions have an error", and so on).
- **Starting a second run while one was already running** returned a generic server error with a
  stack trace. It now returns a clear "model is not ready" message.

**Configuration and reliability**

- **Invalid extractors could be created** — no template selected, a source property that doesn't
  exist, or a source that is neither a PDF nor a property. These are now rejected on both create and
  edit, with a message saying what's wrong. (One of these was the cause of the silent training hang
  above.)
- **A missing database index** meant every model lookup scanned the whole collection. Added, along
  with a uniqueness guarantee that was being relied on but never enforced.
- **Retries on the ML service never actually retried** — the retry logic was discarding the very
  information it used to decide whether to retry.
- **One unusable entry in a batch of suggestions discarded the whole batch** rather than just that
  entry.

### Testing

Beyond the automated suite, the feature was driven end to end against a real ML service across three
manual passes: the full extractor lifecycle, training, finding, cancelling, accepting, auto-accept,
and the entity/template lifecycle — including a two-language collection and all six supported
property types (text, number, date, select, multi-select, markdown).

PR #9716, opened 2026-09-11
Add translation service feature flags and update API routes

txau:
- Updated .env.example to include TRANSLATION_SERVICE_URL and FEATURE_FLAG_TRANSLATION_SERVICE.
- Modified api.js to import translation service routes.
- Enhanced config.ts to handle translation service feature flags and URLs.
- Updated tenantContext.ts and tenantsModel.ts to include translation service properties.
- Added optional timeout parameter to PostJsonInput in HttpClient.ts.

fixes #

PR checklist:
- [ ] Update READ.me ?

QA checklist:
- [ ] Test for side effects
- [ ] UI responsiveness
- [ ] Cross browser testing
- [ ] Code review


app/api/translationService/application/contracts/TranslationServiceContracts.ts
```diff
@@ -0,0 +1,11 @@
+type TranslateInput = {
```
Joao-vi, line 1: I think we could move all these types declarations to `app/api/translationService/application/contracts/TranslationService.ts`, less files, less back and  fourth while navigating on the code base and it will increase cohesion of the TranslationService contract definition (input, outputs + methods definitions)

app/api/translationService/infrastructure/express/RequestTranslationController.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const RequestSchema = z.object({
+  text: z.string().trim().min(1),
+  language_from: z.string().trim().min(1),
+  language_to: z.string().trim().min(1),
+});
+
+type RequestBody = z.infer<typeof RequestSchema>;
+
+class RequestTranslationController extends AbstractController<RequestBody> {
```
Joao-vi: We should create type definition for Request and Response for the api at `app/shared/contracts/[name of the module].ts` so both frontend and backend can import from and depend on it. 

The schema definition can also be placed there (I don't know if frontend can use them to validate the input at the form on  UI, it would be great if so, same validation back and front) if you are not sure it's fine just placing type definition for request and response there and import it on the controller. 

reference where the pattern is applied `CreateUserController.ts`

app/api/translationService/infrastructure/ExternalTranslationService.ts
```diff
@@ -0,0 +1,46 @@
...
+      this.dependencies.url
+    );
+    translateUrl.searchParams.set('text', input.text);
+    translateUrl.searchParams.set('language_from', input.language_from);
+    translateUrl.searchParams.set('language_to', input.language_to);
+
+    const body = await this.dependencies.httpClient.postJson<TranslateResponseDTO>({
+      url: translateUrl.toString(),
+      body: {},
+      timeoutMs: TRANSLATION_TIMEOUT_MS,
```
Joao-vi: I like the fact timeouts are now  being used on external integrations! 

I hope in the future we could also start using Circuit Breaker so we can protect uwazi api and also the external service.

app/api/translationService/infrastructure/ExternalTranslationService.ts
```diff
@@ -0,0 +1,46 @@
...
+    });
+
+    if (typeof body?.translated_text !== 'string') {
+      throw new Error('Translation service did not return translated_text');
```
Joao-vi: Native javascript errors are now being handled/mapped as an unexpected error (something really bad or unexpected happened), I think here we could create a dedicated error class to throw this validation error. 

You can see a reference for this pattern right here: `app/api/pages.v2/domain/errors.ts`

app/api/translationService/infrastructure/TranslationServiceFactory.ts
```diff
@@ -0,0 +1,21 @@
...
+import { HttpClientFactory } from '#api/common.v2/infrastructure/HttpClientFactory.js';
+import type { TranslationService } from '../application/contracts/TranslationService.js';
+import { ExternalTranslationService } from './ExternalTranslationService.js';
+
+class TranslationServiceFactory {
+  static createDefault(): TranslationService {
+    const url = tenants.current().featureFlags?.translationServiceUrl;
+
+    if (!url) {
+      throw new Error('Translation service URL is not configured for this tenant');
```
Joao-vi, line 11: Perfect usage of native error object,  a `url` not being here is a configuration problem on infra side and therefore application can not work properly.

app/api/translationService/application/contracts/TranslationServiceContracts.ts
txau: Ok, makes sense but I am going to keep it a it is for consistency with AI Assistant

app/api/translationService/infrastructure/express/RequestTranslationController.ts
txau: Done

PR #9730, opened 2026-09-15
react19 upgrade

mfacar:


mfacar:
closed in favor of #9783

